PR #2, opened 2025-04-25
Sync with sigs.k8s.io/yaml/goyaml.v3

dims:
Pick up changes from `sigs.k8s.io/yaml`:
https://github.com/kubernetes-sigs/yaml/commits/master/goyaml.v3

Specifically these:
https://github.com/kubernetes-sigs/yaml/commit/330e1e54d757cdcf6d0785ed5551b315935f877e
https://github.com/kubernetes-sigs/yaml/commit/b6d8b4506c6b4d97e8acb8ecc315607bf11e30bc
https://github.com/kubernetes-sigs/yaml/commit/63d028eee8d132544869a74ddb58523a36cd4662

Co-Authored-By: Jordan Liggitt <liggitt@google.com>
Co-Authored-By: natasha41575 <natashasarkar@google.com>

cc @liggitt @natasha41575


patch.go
liggitt, line 1: If upstreaming, this should probably go on the main file rather than a patch/patch_test file

dims: @liggitt +1 will add another commit to move things. i wanted us to be able to compare the files here.

ingydotnet:
@dims I amended your 2nd commit msg to start with a capital letter (to match the other commits)

We can write up some contribution guidelines soon...

patch.go
dims: @liggitt Done!

dims:
@ingydotnet thanks for fixing commit message. I've consolidated the patch/patch_test into the rest of the files (we used separate files earlier to better track/maintain things).

@ingydotnet we gotta add some linters! recommend starting with the default options for `golangci-lint` and then start tweaking it as we see fit. (i instinctively run `gofmt -w` as well, so we should see if we want to verify the formatting as well).

ingydotnet (approved):

ingydotnet:
@dims, :+1: for linters but let's resolve #3 first!

Issue #4, opened 2025-05-02
Add linters to the project


ingydotnet:
Once we decide on our rules for #3 let's add linters for the various kinds of files used in the repo.

banjoh:
https://github.com/golangci/golangci-lint is widely used. Once we get to deciding what direction to take, we can consider using it. Its a runner for many linters implemented as plugins, and can be configured to a small set.

zostay:
o/ Ingy! I have a lot of projects setup to use golangci-lint and can contribute boilerplate. I just ran my usual configuration against the source and got 149 issues to start. For example, it found a lot of unused internal parameters, some formatting and spelling issues, some unchecked integer conversions, a bunch of switch statements that aren't exhaustive (but probably don't need to be), a couple unchecked errors, and a few cases of errors being compared directly rather than using `errors.As` or `errors.Is`.

I'm happy to provide a PR for this and include a github action that will automatically lint pushes or PRs. I like going as far as having gofumpt enforced on my own projects, but I'm happy to contribute my full suite with fixes or something that is all but empty to provide a starting point.

ingydotnet:
@zostay Can you group categories of issues into their own PRs to make them easier to approve?

Feel free to start submitting PRs for this.

ingydotnet:
Closing this as I think we are solid on linters.

PR #5, opened 2025-05-10
Fix `go vet` errors

stefanprodan:
This PR fixes the syntactical errors reported by `go vet` and formats the code with `go fmt`.

In addition `go vet` and `go fmt` are now being run in CI.

Part of: #4

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -17,5 +17,7 @@ jobs:
       uses: actions/setup-go@v2
       with:
         go-version: ${{ matrix.go-versions }}
+    - name: Run go vet
+      run: go vet ./...
     - name: Run go test
       run: GO111MODULE=on go test -v -race ./...
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -6,15 +6,32 @@ on:
   pull_request:
     branches: [ v3 ]
 jobs:
+  fmt:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      with:
+        go-version: stable
+    - name: Run go fmt
+      run: go fmt ./...
+    - name: Check if working tree is dirty
+      run: |
+        if [[ $(git diff --stat) ]]; then
+          git --no-pager diff
+          echo 'Run go fmt ./... and commit changes.'
+          exit 1
+        fi
   build:
     strategy:
       matrix:
         go-versions: [1.21.x, 1.22.x, 1.23.x, 1.24.x]
     runs-on: ubuntu-latest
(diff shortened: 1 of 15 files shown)


dims (approved):

.github/workflows/go.yaml
```diff
@@ -6,15 +6,32 @@ on:
...
     branches: [ v3 ]
 jobs:
+  fmt:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      with:
+        go-version: 'stable'
```
ingydotnet: Quotes not needed here.

.github/workflows/go.yaml
```diff
@@ -6,15 +6,32 @@ on:
...
+    - name: Run go fmt
+      run: go fmt ./...
+    - name: Check if working tree is dirty
+      run: |
+        if [[ $(git diff --stat) != '' ]]; then
```
ingydotnet: `if [[ $(git diff --stat) ]]; then` is preferred.

decode_test.go
```diff
@@ -27,6 +27,7 @@ import (
 	"time"
 
 	. "gopkg.in/check.v1"
+
```
ingydotnet, line 30: Can you amend this commit with the literal command that you ran to "Fix go vet errors"?

This would be helpful for me and the others on the team who aren't as well learned in Go lang.

.github/workflows/go.yaml
```diff
@@ -6,15 +6,32 @@ on:
...
+          git --no-pager diff
+          echo 'run go fmt ./... and commit changes'
```
ingydotnet: I'd prefer 
```
echo 'Run go fmt ./... and commit changes.'
```

.github/workflows/go.yaml
```diff
@@ -6,15 +6,32 @@ on:
...
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      with:
+        go-version: 'stable'
+    - name: Run go fmt
+      run: go fmt ./...
+    - name: Check if working tree is dirty
+      run: |
```
ingydotnet, line 20: I'd like to see shell scripts be in their own files so we can lint them more easily.
Leave this one alone for now...
I have some ideas that I'll present later this week or so.

decode_test.go
stefanprodan: I haven’t used a command, the fixes were all manual, going over the vet errors one by one.

ingydotnet: Ah. Thanks for letting me know.

ingydotnet (approved):
Thanks for the updates. LGTM.
Will merge.

PR #19, opened 2025-05-18
Allow unwrapping of non-terminal TypeErrors

andig:
This PR combines https://github.com/go-yaml/yaml/pull/901 and https://github.com/andig/yaml to allow:

- retrieving location of parser errors through new `ParserError`
- retrieving location though the Go errors mechanism by adding `UnmarshalError`and making `TypeError` unwrappable

It was never merged upstream but got positive feedback.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -22,7 +22,6 @@ import (
 	"io"
 	"math"
 	"reflect"
-	"strconv"
 	"time"
 )
 
@@ -110,7 +109,6 @@ func (p *parser) peek() yaml_event_type_t {
 }
 
 func (p *parser) fail() {
-	var where string
 	var line int
 	if p.parser.context_mark.line != 0 {
 		line = p.parser.context_mark.line
@@ -125,16 +123,13 @@ func (p *parser) fail() {
 			line++
 		}
 	}
-	if line != 0 {
-		where = "line " + strconv.Itoa(line) + ": "
-	}
 	var msg string
 	if len(p.parser.problem) > 0 {
 		msg = p.parser.problem
 	} else {
 		msg = "unknown problem parsing YAML content"
 	}
-	failf("%s%s", where, msg)
+	fail(&ParserError{msg, line})
 }
 
 func (p *parser) anchor(n *Node, anchor []byte) {
@@ -313,7 +308,7 @@ func (p *parser) mapping() *Node {
 type decoder struct {
 	doc     *Node
 	aliases map[*Node]bool
-	terrors []string
+	terrors []*UnmarshalError
 
 	stringMapType  reflect.Type
 	generalMapType reflect.Type
@@ -357,19 +352,29 @@ func (d *decoder) terror(n *Node, tag string, out reflect.Value) {
 			value = " `" + value + "`"
 		}
 	}
-	d.terrors = append(d.terrors, fmt.Sprintf("line %d: cannot unmarshal %s%s into %s", n.Line, shortTag(tag), value, out.Type()))
+	d.terrors = append(d.terrors, &UnmarshalError{
+		Err:    fmt.Errorf("cannot unmarshal %s%s into %s", shortTag(tag), value, out.Type()),
+		Line:   n.Line,
+		Column: n.Column,
+	})
 }
 
 func (d *decoder) callUnmarshaler(n *Node, u Unmarshaler) (good bool) {
 	err := u.UnmarshalYAML(n)
-	if e, ok := err.(*TypeError); ok {
+	switch e := err.(type) {
+	case nil:
+		return true
+	case *TypeError:
 		d.terrors = append(d.terrors, e.Errors...)
 		return false
+	default:
+		d.terrors = append(d.terrors, &UnmarshalError{
+			Err:    err,
+			Line:   n.Line,
+			Column: n.Column,
+		})
+		return false
 	}
-	if err != nil {
-		fail(err)
-	}
-	return true
 }
 
 func (d *decoder) callObsoleteUnmarshaler(n *Node, u obsoleteUnmarshaler) (good bool) {
@@ -384,14 +389,20 @@ func (d *decoder) callObsoleteUnmarshaler(n *Node, u obsoleteUnmarshaler) (good 
 		}
 		return nil
 	})
-	if e, ok := err.(*TypeError); ok {
+	switch e := err.(type) {
+	case nil:
+		return true
+	case *TypeError:
 		d.terrors = append(d.terrors, e.Errors...)
 		return false
+	default:
+		d.terrors = append(d.terrors, &UnmarshalError{
+			Err:    err,
+			Line:   n.Line,
+			Column: n.Column,
+		})
+		return false
 	}
-	if err != nil {
-		fail(err)
-	}
-	return true
 }
 
 // d.prepare initializes and dereferences pointers and calls UnmarshalYAML
@@ -763,7 +774,11 @@ func (d *decoder) mapping(n *Node, out reflect.Value) (good bool) {
 			for j := i + 2; j < l; j += 2 {
 				nj := n.Content[j]
 				if ni.Kind == nj.Kind && ni.Value == nj.Value {
-					d.terrors = append(d.terrors, fmt.Sprintf("line %d: mapping key %#v already defined at line %d", nj.Line, nj.Value, ni.Line))
+					d.terrors = append(d.terrors, &UnmarshalError{
+						Err:    fmt.Errorf("mapping key %#v already defined at line %d", nj.Value, ni.Line),
+						Line:   nj.Line,
+						Column: nj.Column,
+					})
 				}
 			}
 		}
@@ -911,7 +926,11 @@ func (d *decoder) mappingStruct(n *Node, out reflect.Value) (good bool) {
 		if info, ok := sinfo.FieldsMap[sname]; ok {
 			if d.uniqueKeys {
 				if doneFields[info.Id] {
-					d.terrors = append(d.terrors, fmt.Sprintf("line %d: field %s already set in type %s", ni.Line, name.String(), out.Type()))
+					d.terrors = append(d.terrors, &UnmarshalError{
+						Err:    fmt.Errorf("field %s already set in type %s", name.String(), out.Type()),
+						Line:   ni.Line,
+						Column: ni.Column,
+					})
 					continue
 				}
 				doneFields[info.Id] = true
@@ -931,7 +950,11 @@ func (d *decoder) mappingStruct(n *Node, out reflect.Value) (good bool) {
 			d.unmarshal(n.Content[i+1], value)
 			inlineMap.SetMapIndex(name, value)
 		} else if d.knownFields {
-			d.terrors = append(d.terrors, fmt.Sprintf("line %d: field %s not found in type %s", ni.Line, name.String(), out.Type()))
+			d.terrors = append(d.terrors, &UnmarshalError{
+				Err:    fmt.Errorf("field %s not found in type %s", name.String(), out.Type()),
+				Line:   ni.Line,
+				Column: ni.Column,
+			})
 		}
 	}
 
--- a/decode_test.go
+++ b/decode_test.go
@@ -1082,6 +1082,25 @@ func TestDecoderErrors(t *testing.T) {
 	}
 }
 
+func TestParserError(t *testing.T) {
+	var v struct {
+		A, B int
+	}
+	data := "a: 1\n=\nb: 2"
+	err := yaml.Unmarshal([]byte(data), &v)
+	asErr := new(yaml.ParserError)
+	if !errors.As(err, &asErr) {
+		t.Fatalf("error returned by Unmarshal doesn't unwrap into yaml.ParserError")
+	}
+	expectedErr := &yaml.ParserError{
+		Message: "could not find expected ':'",
+		Line:    2,
+	}
+	if !reflect.DeepEqual(asErr, expectedErr) {
+		t.Errorf("wrong err, got %#v expected %#v", asErr, expectedErr)
+	}
+}
+
 var unmarshalerTests = []struct {
 	data, tag string
 	value     interface{}
@@ -1247,8 +1266,8 @@ func TestUnmarshalerWholeDocument(t *testing.T) {
 }
 
 func TestUnmarshalerTypeError(t *testing.T) {
-	unmarshalerResult[2] = &yaml.TypeError{[]string{"foo"}}
-	unmarshalerResult[4] = &yaml.TypeError{[]string{"bar"}}
+	unmarshalerResult[2] = &yaml.TypeError{[]*yaml.UnmarshalError{{Err: errors.New("foo"), Line: 1, Column: 1}}}
+	unmarshalerResult[4] = &yaml.TypeError{[]*yaml.UnmarshalError{{Err: errors.New("bar"), Line: 1, Column: 1}}}
 	defer func() {
 		delete(unmarshalerResult, 2)
 		delete(unmarshalerResult, 4)
@@ -1265,8 +1284,8 @@ func TestUnmarshalerTypeError(t *testing.T) {
 	expectedError := "" +
 		"yaml: unmarshal errors:\n" +
 		"  line 1: cannot unmarshal !!str `A` into int\n" +
-		"  foo\n" +
-		"  bar\n" +
+		"  line 1: foo\n" +
+		"  line 1: bar\n" +
 		"  line 1: cannot unmarshal !!str `B` into int"
 	if err == nil {
 		t.Fatalf("Unmarshal() expected error, got none.")
@@ -1300,8 +1319,8 @@ func TestUnmarshalerTypeError(t *testing.T) {
 }
 
 func TestObsoleteUnmarshalerTypeError(t *testing.T) {
-	unmarshalerResult[2] = &yaml.TypeError{[]string{"foo"}}
-	unmarshalerResult[4] = &yaml.TypeError{[]string{"bar"}}
+	unmarshalerResult[2] = &yaml.TypeError{[]*yaml.UnmarshalError{{Err: errors.New("foo"), Line: 1, Column: 1}}}
+	unmarshalerResult[4] = &yaml.TypeError{[]*yaml.UnmarshalError{{Err: errors.New("bar"), Line: 1, Column: 1}}}
 	defer func() {
 		delete(unmarshalerResult, 2)
 		delete(unmarshalerResult, 4)
@@ -1318,8 +1337,8 @@ func TestObsoleteUnmarshalerTypeError(t *testing.T) {
 	expectedError := "" +
 		"yaml: unmarshal errors:\n" +
 		"  line 1: cannot unmarshal !!str `A` into int\n" +
-		"  foo\n" +
-		"  bar\n" +
+		"  line 1: foo\n" +
+		"  line 1: bar\n" +
 		"  line 1: cannot unmarshal !!str `B` into int"
 	matched, mre := regexp.MatchString(expectedError, err.Error())
 	if mre != nil {
@@ -1449,9 +1468,30 @@ func (ft *failingUnmarshaler) UnmarshalYAML(node *yaml.Node) error {
 }
 
 func TestUnmarshalerError(t *testing.T) {
-	err := yaml.Unmarshal([]byte("a: b"), &failingUnmarshaler{})
-	if err != failingErr {
-		t.Fatalf("Unmarshal() returned %v, want %v", err, failingErr)
+	data := `{foo: 123, bar: {}, spam: "test"}`
+	dst := struct {
+		Foo  int
+		Bar  *obsoleteFailingUnmarshaler
+		Spam string
+	}{}
+	err := yaml.Unmarshal([]byte(data), &dst)
+	expectedErr := &yaml.TypeError{
+		Errors: []*yaml.UnmarshalError{
+			{Line: 1, Column: 17, Err: failingErr},
+		},
+	}
+	if !reflect.DeepEqual(expectedErr, err) {
+		t.Errorf("expected error %#v, got #%v", expectedErr, err)
+	}
+	// whatever could be unmarshaled must be unmarshaled
+	if got, want := dst.Foo, 123; got != want {
+		t.Errorf("foo unmarshaled incorrectly, got %#v, want  %#v", got, want)
+	}
+	if got, want := dst.Bar, (&obsoleteFailingUnmarshaler{}); !reflect.DeepEqual(got, want) {
+		t.Errorf("bar unmarshaled incorrectly, got %#v, want  %#v", got, want)
+	}
+	if got, want := dst.Spam, "test"; got != want {
+		t.Errorf("spam unmarshaled incorrectly, got %#v, want  %#v", got, want)
 	}
 }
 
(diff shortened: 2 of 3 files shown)


ingydotnet:
@andig Thanks for this. 

I'd like to wait until after the launch, until we start merging PRs. Hopefully soon.

andig:
Absolutely, thanks for considering it at all.

WGH-:
It's also possible to handle errors returned by `UnmarshalYAML` the same way. I can also write tests for this, and submit them as a PR into your PR, if you don't mind.

```diff
diff --git i/decode.go w/decode.go
index bfb8de5..f02ce37 100644
--- i/decode.go
+++ w/decode.go
@@ -363,14 +363,20 @@ func (d *decoder) terror(n *Node, tag string, out reflect.Value) {
 
 func (d *decoder) callUnmarshaler(n *Node, u Unmarshaler) (good bool) {
 	err := u.UnmarshalYAML(n)
-	if e, ok := err.(*TypeError); ok {
+	switch e := err.(type) {
+	case nil:
+		return true
+	case *TypeError:
 		d.terrors = append(d.terrors, e.Errors...)
 		return false
+	default:
+		d.terrors = append(d.terrors, UnmarshalError{
+			Message: err.Error(),
+			Line:    n.Line,
+			Column:  n.Column,
+		})
+		return false
 	}
-	if err != nil {
-		fail(err)
-	}
-	return true
 }
 
 func (d *decoder) callObsoleteUnmarshaler(n *Node, u obsoleteUnmarshaler) (good bool) {
@@ -385,14 +391,20 @@ func (d *decoder) callObsoleteUnmarshaler(n *Node, u obsoleteUnmarshaler) (good
 		}
 		return nil
 	})
-	if e, ok := err.(*TypeError); ok {
+	switch e := err.(type) {
+	case nil:
+		return true
+	case *TypeError:
 		d.terrors = append(d.terrors, e.Errors...)
 		return false
+	default:
+		d.terrors = append(d.terrors, UnmarshalError{
+			Message: err.Error(),
+			Line:    n.Line,
+			Column:  n.Column,
+		})
+		return false
 	}
-	if err != nil {
-		fail(err)
-	}
-	return true
 }
 
 // d.prepare initializes and dereferences pointers and calls UnmarshalYAML
```

andig:
Good idea 👍🏻

decode.go
```diff
@@ -313,7 +308,7 @@ func (p *parser) mapping() *Node {
 type decoder struct {
 	doc     *Node
 	aliases map[*Node]bool
-	terrors []string
+	terrors []UnmarshalError
```
andig: I didn‘t check this, maybe we‘d also make this a simple error and use errors.Join instead of append? Can be a separate PR though.

andig:
@WGH- do your want to open PR targeting this branch?

WGH-:
> @WGH- do your want to open PR targeting this branch?

I've just done it. See https://github.com/andig/go-yaml/pull/1

decode.go
WGH-: I think it's better this way. We can be certain that errors here include location, and it's not just some random `fmt.Errorf` without context.

That said, I'd go further and change `Message string` to `Err error` so exact error (e.g. returned by `UnmarshalYAML` method) can be inspected.

andig:
@ingydotnet any chance we could start with merging soon? This is the only reason I must keep my fork open ;)

ingydotnet:
I'll merge by EOD regardless

andig:
Much appreciated! Please don't forget to squash on merge to keep things clean ;)

ingydotnet:
> Much appreciated! Please don't forget to squash on merge to keep things clean ;)

@andig Sorry I was in a rush to the airport on my phone and thought that this was a different PR.

I'll try to get this handled ASAP, but maybe not today.

ingydotnet:
@andig, can you do me a favor and do another push? Maybe squash and force?
I want to see if the CodeQL workflow gets triggered...

andig:
Updated the branch

ingydotnet:
> Updated the branch

Thank you. That did trigger the workflows. I had to approve them manually afterwards, but they ran and passed.

ingydotnet:
@andig, I'm pretty sure this will need to go into v4. The team hasn't discussed v4 releases yet. I'll call a meeting this week.

andig:
No problem, I can reopen against v4. It would be good if v4 then also becomes installable via go get, even if not released.

ingydotnet:
We need reviews on this from people other than me.

WGH-:
> We need reviews on this from people other than me.

I'll take a look in a few days

ingydotnet:
> > We need reviews on this from people other than me.
> 
> I'll take a look in a few days

I appreciate it, but what I meant is that we need 2 team members to review this in order to merge and I'm recusing myself.

yaml.go
```diff
@@ -307,24 +308,69 @@ type yamlError struct {
...
+	panic(&yamlError{err})
 }
 
 func failf(format string, args ...interface{}) {
-	panic(yamlError{fmt.Errorf("yaml: "+format, args...)})
+	panic(&yamlError{fmt.Errorf("yaml: "+format, args...)})
+}
+
+// ParserError is each error with a source line position found by the parser.
+// Unlike UnmarshalErrors, it is only used for terminal failures.
```
carloslima: Slightly different wording:

```suggestion
// ParserError represents a fatal error encountered during the parsing phase.
// These errors typically indicate a syntax issue in the YAML document that
// prevents further processing.
```

yaml.go
```diff
@@ -307,24 +308,69 @@ type yamlError struct {
...
+	var b strings.Builder
+	b.WriteString("yaml: ")
+	if e.Line != 0 {
+		b.WriteString("line " + strconv.Itoa(e.Line) + ": ")
+	}
+	b.WriteString(e.Message)
+	return b.String()
+}
+
+// UnmarshalError is each error with a source position found by Unmarshal.
```
carloslima: Slightly different wording:

```suggestion
// UnmarshalError represents a single, non-fatal error that occurred during
// the unmarshaling of a YAML document into a Go value.
```

decode_test.go
```diff
@@ -982,6 +982,18 @@ func (s *S) TestDecoderErrors(c *C) {
...
+		A, B int
+	}
+	data := "a: 1\n=\nb: 2"
+	err := yaml.Unmarshal([]byte(data), &v)
+	c.Assert(err, DeepEquals, &yaml.ParserError{
+		Message: "could not find expected ':'",
+		Line:    2,
+	})
+}
+
```
carloslima, line 1103: would it be useful to add a test using `errors.As()` for inspecting the wrapped errors as a demonstration of the Unwrap []errors implementation?

carloslima (approved):
Left a few smallish suggestions. This looks good 👍

yaml.go
```diff
@@ -307,24 +308,69 @@ type yamlError struct {
...
+type UnmarshalError struct {
+	Message string
```
WGH-: What do you guys think of replacing `Message string` with `Err error`? So any error values returned by `UnmarshalYAML` will be preserved as is.

carloslima: I think it makes sense. 👍 
Was there a reason you didn't go that way at first? :)

andig: This is done. I've noticed that `gofmt` wants to do quite a bit more- it would make sende to `gofmt` everything in a separate PR before making any more changes.

decode_test.go
andig: This is also done

andig:
Another oddity I've stumbled upon is that the tests import and refer to `yaml` since this is the current module that shouldn't be necessary and creates confusion. However, this would make for a bigger diff so leaving as-is for now.

WGH-:
> Another oddity I've stumbled upon is that the tests import and refer to yaml since this is the current module that shouldn't be necessary and creates confusion. However, this would make for a bigger diff so leaving as-is for now.

That's because tests are not in `yaml` package, but rather in `yaml_test`. This forces tests to access only exported types/functions to facilitate proper black box testing. Although somewhat seldom used in Go, this has always been supported by go test, and [mentioned in `testing` package docs](https://pkg.go.dev/testing).

WGH- (approved):

andig:
> That's because tests are not in yaml package, but rather in yaml_test.

Same folder, different module. Didn't notice/expect that. Another thing learned, thank you!

WGH-:
> Same folder, different module. Didn't notice/expect that. Another thing learned, thank you!

That's packages. Modules and packages are different things in Go, too ;)

andig:
I'm stuck rebasing this on main:


```
--- FAIL: TestUnmarshalerTypeError (0.00s)
    decode_test.go:1389: Unmarshal() returned error "yaml: unmarshal errors:\n  line 1: cannot unmarshal !!str `A` into int\n  line 0: foo\n  line 0: bar\n  line 1: cannot unmarshal !!str `B` into int", want match for "yaml: unmarshal errors:\n  line 1: cannot unmarshal !!str `A` into int\n  foo\n  bar\n  line 1: cannot unmarshal !!str `B` into int"
--- FAIL: TestObsoleteUnmarshalerTypeError (0.00s)
    decode_test.go:1439: Unmarshal() returned error "yaml: unmarshal errors:\n  line 1: cannot unmarshal !!str `A` into int\n  line 0: foo\n  line 0: bar\n  line 1: cannot unmarshal !!str `B` into int", want match for "yaml: unmarshal errors:\n  line 1: cannot unmarshal !!str `A` into int\n  foo\n  bar\n  line 1: cannot unmarshal !!str `B` into int"
--- FAIL: TestUnmarshalerError (0.00s)
    decode_test.go:1564: Unmarshal() returned yaml: unmarshal errors:
          line 1: failingErr, want failingErr
--- FAIL: TestObsoleteUnmarshalerError (0.00s)
    decode_test.go:1577: Unmarshal() returned yaml: unmarshal errors:
          line 1: failingErr, want failingErr
FAIL
FAIL	go.yaml.in/yaml/v3	2.726s
F
```

any help would be greatly appreciated.

WGH-:
@andig Most likely you tried to compare `T` and `*T` with `reflect.DeepEqual`.

In any case, here is the rebased result which passes the tests: https://github.com/WGH-/go-yaml/tree/patch-1-rebased

I did not clean up commit history though, and some commits are broken. This PR has to be squashed into 1-2 commits before merge anyway.

andig:
Rebased thanks to help of @WGH-. Unfortunately gofmt has now crept in. Please let me know if this is mergeable or I need to revert.

andig:
I've fixed the stray `gofmt`, sorry for the noise.

One last thing I've noted that I'm not happy with is, that we're mixing receiver types between 

    func (e *ParserError) Error() string

and

    func (e UnmarshalError) Error() string

Would everybody agree that we should consistently make this a pointer receiver?

ingydotnet:
@andig A commit message consisting only of "Rebase" for this entire PR isn't really helpful.

Can you amend that and force push?

If you do that I might able to merge since `Waiting on 1 reapproval from someone other than the last pusher.` will no longer apply.

Also regarding "Would everybody agree that we should consistently make this a pointer receiver?",
should this be holding up the merge?

andig:
> A commit message consisting only of "Rebase" for this entire PR isn't really helpful.

I thought we had already agreed on simply setting the desired message on merge: https://github.com/yaml/go-yaml/pull/19#issuecomment-3018556191

> Also regarding "Would everybody agree that we should consistently make this a pointer receiver?",
should this be holding up the merge?

I currently need to to check for both error types like this:

```go
	var (
		ype *yaml.ParserError
		yue yaml.UnmarshalError
	)
	switch {
	case errors.As(err, &ype):
		res.Line = ype.Line
	case errors.As(err, &yue):
		res.Line = yue.Line
	}
```

This is highly unintuitive and error-prone. Since both types are new and create new public api we should add them consistently (my mistake not doing to in the first place). Sorry for the late addition.

WGH-:
I'd also appreciate it if you mentioned me in "Co-authored-by:" trailer (https://github.com/andig/go-yaml/pull/1)

WGH-:
> I currently need to to check for both error types like this: <...>

Are you talking about the tests, or some yet unpublished code? Or is this how you expect end-users to use the error location?

andig:
Thank you for the question, I should have given more context. I'm talking about our consumer-side code. We're highlighting yaml errors to the user and need to extract the line numbers for that purpose. Since both error types can occur and contain line information we're checking for both using `errors.As()`. This `error` is (was, last commit changed that) using different receiver types doing so required an inconsistent mixture of pointer and non-pointer receiver. If we could do this any better I'd be happy to change it.

andig:
Done @WGH- please check commit message- you've asked for "trailer"?

decode_test.go
```diff
@@ -1241,8 +1260,8 @@ func TestUnmarshalerWholeDocument(t *testing.T) {
 }
 
 func TestUnmarshalerTypeError(t *testing.T) {
-	unmarshalerResult[2] = &yaml.TypeError{[]string{"foo"}}
-	unmarshalerResult[4] = &yaml.TypeError{[]string{"bar"}}
+	unmarshalerResult[2] = &yaml.TypeError{[]*yaml.UnmarshalError{{errors.New("foo"), 1, 1}}}
+	unmarshalerResult[4] = &yaml.TypeError{[]*yaml.UnmarshalError{{errors.New("bar"), 1, 1}}}
```
ccoVeille: Please use field names, it could break things if there is a refactoring or simply if fields are swapped.


```suggestion
	unmarshalerResult[2] = &yaml.TypeError{[]*yaml.UnmarshalError{{Err: errors.New("foo"), Line: 1, Column: 1}}}
	unmarshalerResult[4] = &yaml.TypeError{[]*yaml.UnmarshalError{{Err: errors.New("bar"), Line: 1, Column: 1}}}
```

decode_test.go
```diff
@@ -1076,6 +1076,25 @@ func TestDecoderErrors(t *testing.T) {
...
+	if !errors.As(err, &asErr) {
+		t.Fatalf("error returned by Unmarshal doesn't unwrap into yaml.ParserError")
+	}
+	expectedErr := &yaml.ParserError{
+		Message: "could not find expected ':'",
+		Line:    2,
+	}
+	if !reflect.DeepEqual(asErr, expectedErr) {
+		t.Errorf("wrong err, got %#v expected %#v", asErr, expectedErr)
+	}
```
ccoVeille, line 1101: I'm unsure why there are pointers everywhere now.


```go
        asErr := new(yaml.ParserError) 
         if !errors.As(err, &asErr) {
```

So here we are using a pointer of pointer, it looks overkill. Is there any existing pattern for that? Is this a memory/performance optimization?

Why comparing the pointers with deep equal? I would have compared the values

They cannot be equal as it's a newly created object, and so a new pointer.

Also why having a []*yaml.ParserError ?

I'm new to the code, but all this looks strange to me

decode_test.go
```diff
@@ -1294,8 +1313,8 @@ func TestUnmarshalerTypeError(t *testing.T) {
 }
 
 func TestObsoleteUnmarshalerTypeError(t *testing.T) {
-	unmarshalerResult[2] = &yaml.TypeError{[]string{"foo"}}
-	unmarshalerResult[4] = &yaml.TypeError{[]string{"bar"}}
+	unmarshalerResult[2] = &yaml.TypeError{[]*yaml.UnmarshalError{{errors.New("foo"), 1, 1}}}
+	unmarshalerResult[4] = &yaml.TypeError{[]*yaml.UnmarshalError{{errors.New("bar"), 1, 1}}}
```
ccoVeille: Same as

https://github.com/yaml/go-yaml/pull/19#discussion_r2216642010

decode_test.go
```diff
@@ -1456,9 +1496,30 @@ func (ft *obsoleteFailingUnmarshaler) UnmarshalYAML(unmarshal func(interface{})
...
+	}
+	// whatever could be unmarshaled must be unmarshaled
+	if got, want := dst.Foo, 123; got != want {
+		t.Errorf("foo unmarshaled incorrectly, got %#v, want  %#v", got, want)
+	}
+	if got, want := dst.Bar, (&obsoleteFailingUnmarshaler{}); !reflect.DeepEqual(got, want) {
+		t.Errorf("bar unmarshaled incorrectly, got %#v, want  %#v", got, want)
+	}
+	if got, want := dst.Spam, "test"; got != want {
+		t.Errorf("spam unmarshaled incorrectly, got %#v, want  %#v", got, want)
```
ccoVeille: Remove the extra spaces


```suggestion
		t.Errorf("foo unmarshaled incorrectly, got %#v, want %#v", got, want)
	}
	if got, want := dst.Bar, (&obsoleteFailingUnmarshaler{}); !reflect.DeepEqual(got, want) {
		t.Errorf("bar unmarshaled incorrectly, got %#v, want %#v", got, want)
	}
	if got, want := dst.Spam, "test"; got != want {
		t.Errorf("spam unmarshaled incorrectly, got %#v, want %#v", got, want)
```

yaml.go
```diff
@@ -307,24 +308,75 @@ type yamlError struct {
...
+	return e.Err
 }
 
 // A TypeError is returned by Unmarshal when one or more fields in
 // the YAML document cannot be properly decoded into the requested
 // types. When this error is returned, the value is still
 // unmarshaled partially.
 type TypeError struct {
-	Errors []string
+	Errors []*UnmarshalError
```
ccoVeille, line 358: Same remark as

https://github.com/yaml/go-yaml/pull/19#discussion_r2216651608

yaml.go
```diff
@@ -307,24 +308,75 @@ type yamlError struct {
...
+	return b.String()
+}
+
+func (e *TypeError) Unwrap() []error {
+	errs := make([]error, 0, len(e.Errors))
+	for _, err := range e.Errors {
+		errs = append(errs, err)
+	}
+	return errs
 }
```
ccoVeille, line 376: I'm unsure I get it, isn't this enough?

```suggestion
func (e *TypeError) Unwrap() []error {
	return e.Errors
 }
```

yaml.go
```diff
@@ -307,24 +308,75 @@ type yamlError struct {
 	err error
 }
 
+func (e *yamlError) Unwrap() error {
+	return e.err
+}
```
ccoVeille: Can't you just unwrap on the object and not its pointer?


```suggestion
func (e yamlError) Unwrap() error {
	return e.err
}
```

I feel like this is why you are using pointer in the panic a few like below

yaml.go
andig: ..if you can make this work? But e.Errors are not errors...

decode_test.go
andig: > So here we are using a pointer of pointer

Because of https://github.com/yaml/go-yaml/pull/19#issuecomment-3082732569. Either value or pointer but it needs to be consistent.

The reason why I chose pointer is https://groups.google.com/g/golang-nuts/c/3M3QDkochcs, or more specifically https://go.dev/play/p/AQ5F-ycJCbE: As long as you return errors as plain structs, a consumer using errors.As() can never be sure if he should match against plain or pointer type without looking into your code. If you return pointer to struct, the compiler will error if you try to match on plain type which removes the ambiguity.

Effectively, a library returning pointer to type but implementing `Error` on the plain type makes `errors.As` unusable afaiu. Returning pointer to type and implementing `Error` on the pointer receiver makes sure consumers need to pick the right type for their comparison.

Bottom line is: never return pointers to errors where `Error` is implemented on the plain receiver. We (of course do that a well).

decode_test.go
andig: di you want to suggest the required change?

ccoVeille:
```suggestion
	unmarshalerResult[2] = &yaml.TypeError{[]*yaml.UnmarshalError{{Err: errors.New("foo"), Line: 1, Column: 1}}}
	unmarshalerResult[4] = &yaml.TypeError{[]*yaml.UnmarshalError{{Err: errors.New("bar"), Line: 1, Column: 1}}}
```

yaml.go
WGH-: Go lacks variance for slices, so unfortunately no

decode_test.go
WGH-: > So here we are using a pointer of pointer, it looks overkill. Is there any existing pattern for that? Is this a memory/performance optimization?

You can't do that the other way with `errors.As`. You need to pass it a pointer to a value assignable to an error you're looking for. Since you're looking for a `*yaml.ParserError`, you have to pass `**yaml.ParserError`.

> Why comparing the pointers with deep equal? I would have compared the values

In this particular case, given that all fields of `ParserError` are primitive, `*asErr != *expectedErr` would be fine. But that would break if `ParserErrors` gets new "non-comparable" fields like slices. Better compare everything in the same universal way.

> Also why having a []*yaml.ParserError ?

Did you mean `[]*UnmarshalError`?

I don't think one choice is strictly better than another.

1. `[]UnmarshalError` is more compact is memory.
2. `[]*UnmarshalError` is easier to convert to `[]error`, as it won't involve new copies when "boxing" into an `error`.

I'd prefer to leave `[]*UnmarshalError` as `error` is implemented on a pointer.

yaml.go
WGH-: Performance-wise, there's no difference, since `yamlError` gets packed into an `interface{}` when panicking. It escapes to the heap anyway.

WGH-: Now that @ccoVeille mentioned this, this method appears to be unused.

decode_test.go
WGH-: If you're asking why `error` interface is implemented for a pointer receiver for all `TypeError`, `UnmarshalError`, and `ParserError`, then the answer is mostly consistency (https://github.com/yaml/go-yaml/pull/19#issuecomment-3077172986).

Another reason is that it reduces copies when error is type-asserted to a concrete type.

And argumentum ad populum: most Go code, including stdlib, does that.

yaml.go
andig: It‘s required for unwrapping the inner error.

WGH-:
> Thank you for the question, I should have given more context. I'm talking about our consumer-side code. We're highlighting yaml errors to the user and need to extract the line numbers for that purpose. Since both error types can occur and contain line information we're checking for both using errors.As(). This error is (was, last commit changed that) using different receiver types doing so required an inconsistent mixture of pointer and non-pointer receiver. If we could do this any better I'd be happy to change it.

I suppose you could add a function `ErrorLocation(err error) *Location` that would extract location from an error, if any. I'd personally postpone it for a different PR as there are many ways to design that (ranging fom simple `ErrorLocation(error) int` to somewhat excessive `ErrorLocations(error) iter.Seq[Location]`).

> Done @WGH- please check commit message- you've asked for "trailer"?

I meant `Co-authored-by: WGH <wgh@torlan.ru>`.

yaml.go
WGH-: But `yamlError` is handled by `recover()` in `handleErr` (which accesses the field directly), and then thrown away.

decode_test.go
ccoVeille: Thanks for your feedbacks

yaml.go
```diff
@@ -307,24 +308,75 @@ type yamlError struct {
...
+
```
andig:
```suggestion
```

andig:
Comments addressed, squashed and commit message updated. If everybody is happy with the pointer receivers that's it. If we prepare plain receivers I'll update one more time.

yaml.go
andig: We could make the member `[]error` for this purpose but I've preferred to the keep the error type fixed.

andig:
friendly ping @ingydotnet lest it becomes bit-rotten again

ingydotnet:
@andig, I'll get this done today or tomorrow. Thanks for the reminder!

ingydotnet:
@andig Merged! 

BTW, you can also ping us by joining the conversation here: https://cloud-native.slack.com/archives/C08PPAT8PS7

If you need an invite to the CNCF slack, email me: ingy@ingy.net

andig:
@ingydotnet I'm lost how to `go get` the latest commit:

```
go get go.yaml.in/yaml/v3@main
server/http_site_handler.go:101:13: undefined: yaml.UnmarshalError

go get go.yaml.in/yaml@main
go: go.yaml.in/yaml@main: unrecognized import path "go.yaml.in/yaml": parse https://go.yaml.in/yaml?go-get=1: no go-import meta tags ()

go get go.yaml.in/yaml/v4@main
go: go.yaml.in/yaml/v4@main: unrecognized import path "go.yaml.in/yaml/v4": reading https://go.yaml.in/yaml/v4?go-get=1: 404 Not Found

go get go.yaml.in/yaml/v3@3e63b6a
go: go.yaml.in/yaml/v3@3e63b6a: go.yaml.in/yaml/v3@v3.0.3-0.20250730164212-3e63b6ac2aad: invalid version: go.mod has non-.../v3 module path "go.yaml.in/yaml/v4" at revision 3e63b6ac2aad

go get go.yaml.in/yaml@3e63b6a
go: go.yaml.in/yaml@3e63b6a: unrecognized import path "go.yaml.in/yaml": parse https://go.yaml.in/yaml?go-get=1: no go-import meta tags ()
```

Does this no longer work due to https://github.com/yaml/go-yaml/commit/7f72b374515434e1a8f114f8732b91e7266105da and a missing v4 forward?

stefanprodan:
@ingydotnet we could do a release candidate so people can test out what we have in main e.g. `v4.0.0-rc.1`

andig:
Just getting access to the branch via forward should be enough. I don't think this requires a release.

ingydotnet:
@stefanprodan @andig @carloslima can we meet on the CNCF slack in 11-12 hours from now to discuss?

ingydotnet:
@stefanprodan @andig @carloslima Also I merged https://github.com/yaml-go/go-yaml-in/pull/2

Does that help?

carloslima:
> @stefanprodan @andig @carloslima Also I merged https://github.com/yaml-go/go-yaml-in/pull/2
> 
> Does that help?
> 

Yes, it works :-)

https://go.dev/play/p/iUVOUhWIt32

PR #22, opened 2025-05-28
[v3] Formalize new package name

dims:
As discussed with @ingydotnet @carloslima @stefanprodan and @scottrigby during the call today (May 28th)


stefanprodan (approved):
LGTM

PR #26, opened 2025-06-04
Basic linter configuration, and a lint action for CI

zostay:
This adds minimal configuration for [golangci-lint](https://golangci-lint.run/). As of this PR, it runs none of the standard linters other than govet, which I chose to do as the first implementation because govet doesn't find any issues with it. I have also ensured that the formatter configuration is disabled.

This also configures a new workflow action named `lint` which runs with the others. It makes use of the latest [golangci/golangci-lint-action@v8] action, which uses the Github output format thing that inserts any problems found into the code tab for review.

This technically resolves issue #4 in the most lukewarm way possible. Followup PRs can expand the linters run and fix the problems identified.

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -24,6 +24,17 @@ jobs:
           echo 'Run go fmt ./... and commit changes.'
           exit 1
         fi
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v5
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      with:
+        go-version: stable
+    - uses: golangci/golangci-lint-action@v8
+      with:
+        version: v2.4
   build:
     strategy:
       matrix:
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,22 @@
+version: '2'
+
+linters:
+  enable:
+  - govet
+  disable:
+  - errcheck
+  - ineffassign
+  - staticcheck
+  - unused
+  exclusions:
+    generated: lax
+    presets:
+    - comments
+    - common-false-positives
+    - legacy
+    - std-error-handling
+
+formatters:
+  enable: []
+  exclusions:
+    generated: lax


.github/workflows/go.yaml
```diff
@@ -23,6 +23,17 @@ jobs:
           echo 'Run go fmt ./... and commit changes.'
           exit 1
         fi
+  lint:
```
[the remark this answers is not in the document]
zostay: I think I'd rather expand the CI with more descriptive names and such as a separate PR since the existing CI is pretty barren as far as commentary goes.

.golangci.yaml
```diff
@@ -0,0 +1,22 @@
+version: "2"
+run:
+  concurrency: 4
```
ccoVeille: Why forcing the concurrency?

Yes, GHA workflows are launched with 4 core, but here you are implying anyone launching golangci-lint locally to use only 4 core. Most developers have more cores.

ccoVeille (approved):

.golangci.yaml
zostay: No strong opinion. Happy to do whatever. It's just what I use on my projects because I don't trust the performance of golangci-lint very much. So far v2 seems better, but v1 could be pretty naughty sometimes. Is this a nit or you want it changed?

ccoVeille: If you have a local problem with golangci-lint, maybe you should use `-j 4` locally.

I can leave with the value of 4 here. It surprised me, and I found no reason for this to be pushed on a public repository

ldez: I have been invoked by @ccoVeille in a DM.

The high CPU usage, in rare cases, of the v1 has been fixed in v2, and the default value has been changed to support Linux container CPU quota, you don't need to limit the concurrency by hand.

yaml.go
```diff
@@ -17,7 +17,7 @@
 //
 // Source code and other details for the project are available at GitHub:
 //
-//	https://github.com/go-yaml/yaml
+//	https://github.com/yaml/go-yaml
```
ingydotnet: Can we do this in a separate PR?

ingydotnet:
@zostay Thanks for all of this. 

I want to get the main branch updated for copyright adjustments etc.

I'll call a team meeting next week.

.golangci.yaml
ccoVeille: As discussed here

- https://github.com/yaml/go-yaml/pull/26#discussion_r2131894215

 please remove this line.


```suggestion
```

ccoVeille:
I would love to see this merged

Could you rebase @zostay ?

yaml.go
ccoVeille: in fact, I found it too and opened a PR yesterday

- https://github.com/yaml/go-yaml/pull/72

and today I found your comment here

.golangci.yaml
zostay: Made this change.

zostay:
I think I've incorporated all the feedback now.

.github/workflows/go.yaml
```diff
@@ -24,6 +24,17 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      with:
+        go-version: stable
+    - uses: golangci/golangci-lint-action@v8
+      with:
+        version: v2.1
```
ccoVeille: Please use v2.4


```suggestion
        version: v2.4
```

.github/workflows/go.yaml
```diff
@@ -24,6 +24,17 @@ jobs:
           echo 'Run go fmt ./... and commit changes.'
           exit 1
         fi
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
```
ccoVeille: v5 came out recently

```suggestion
    - uses: actions/checkout@v5
```

ccoVeille (approved):

ingydotnet (approved):

stefanprodan (approved):
LGTM

dolmen (approved):

Issue #28, opened 2025-06-04
Add dependabot to project


ccoVeille:
Dependabot should be configured to update GHA workflows, but also to keep the go.mod up to date.

I thought it would be added to #26, but I feel like this PR should stay simple.

esacteksab:
Looks like this was solved in #30, is there anything else outstanding to do? Looking at the [file](https://github.com/yaml/go-yaml/blob/main/.github/dependabot.yaml), I don't see anything.

ccoVeille:
Thanks for taking the time to look at opened issues.

This one is indeed done.

PR #30, opened 2025-06-04
Add dependabot configuration

zostay:
I just saw #28. This seems like a pretty obvious config to add to me, so I went ahead and added the standard weekly config for go.mod.


zostay:
Sorry, Copilot is incorrect because I accidentally pushed from the wrong point in the tree. I force pushed to fix it. This only adds `dependabot.yml`

.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
```
ccoVeille: This comment is not necessary, considering the file will evolve

yaml.go
```diff
@@ -17,7 +17,7 @@
 //
 // Source code and other details for the project are available at GitHub:
 //
-//	https://github.com/go-yaml/yaml
+//	https://github.com/yaml/go-yaml
```
ccoVeille: Please drop the commit that introduces this.

.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
...
+version: 2
+updates:
+  - package-ecosystem: "gomod" # See documentation for possible values
```
ccoVeille: Do you plan to add `github-actions` ecosystem in this PR?

Or via a different one, still related to #28

zostay: Done.

yaml.go
zostay: Done.

.github/dependabot.yml
zostay: Done.

ccoVeille (approved):

stefanprodan:
There is nothing to update, the direct dependency `gopkg.in/check.v1` hasn’t got any updates in years.

ccoVeille:
It won't hurt then 😅

The PR also introduces the upgrade of GitHub actions by automatic PR creation.

.github/dependabot.yml
```diff
@@ -0,0 +1,10 @@
...
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
+  - package-ecosystem: "github-actions" # See documentation for possible values
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
```
stefanprodan:
```suggestion
  - package-ecosystem: "github-actions" # See documentation for possible values
    directory: "/" # Location of package manifests
    groups:
      actions:
        patterns:
          - "*"
    schedule:
      interval: "weekly"
```

Let's group them to have a single PR for all actions. Docs here https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/optimizing-pr-creation-version-updates

ingydotnet:
Updated to yaml file standards. @zostay can you double-check please?

zostay:
I just saw this. \o/

PR #37, opened 2025-06-08
Create dependabot.yml

andig:
Check github actions versions. Complements https://github.com/yaml/go-yaml/pull/26.

Afair this will also require adding dependabot to the repo.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,8 +1,8 @@
 version: 2
 updates:
-  - package-ecosystem: "github-actions"
-    directory: "/"
+  - package-ecosystem: github-actions
+    directory: /
     schedule:
-      interval: "monthly"
+      interval: monthly
     labels:
-      - "infrastructure"
+      - infrastructure


stefanprodan (commented):
To avoid spam I think we should group the updates based on type. Also we really need to pin the actions by their digests, otherwise we risk running malware if any of the actions we use are compromised. 

Example for grouping: https://github.com/fluxcd/helm-controller/blob/main/.github/dependabot.yaml
Example for pinned actions: https://github.com/fluxcd/helm-controller/blob/main/.github/workflows/nightly.yaml

.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
```
ingydotnet: None of the double quotes in this file are needed.

seq-in-map lines (starting with `-`) should not be extra indented.

ingydotnet (changes requested):
As prominent YAML project I would like to use the best YAML practices here in all the YAML files in this repo.

Let's rename the file to `dependabot.yaml` and mind the extra quotes and seq-in-map indentation.

.github/dependabot.yml
andig: My linter insists on indents :(

andig:
Feel free to grab...

PR #38, opened 2025-06-08
Add general purpose GNUmakefile

ingydotnet:
This `GNUmakefile` auto-installs (unless go is installed already) go locally in `.cache/local/go-<version>/bin/go` and uses that for all make commands.
We use the name `GNUmakefile` instead of `Makefile` because that will prevent people from using a non-GNU `make` command to run `make` commands.

This doesn't prevent anyone from doing things using their normal style but it adds the following benefits for everyone:

* `make test` just works in almost any environment
* You can clone the repo and run `make test` without needing anything installed (except make, curl and bash).
* Test with any go version using: `make test GO-VERSION=1.23.4`
* To get a shell with a particular go version installed and ready: `make shell GO-VERSION=1.23.4`


GNUmakefile
```diff
@@ -0,0 +1,14 @@
+M := .git/.makes
```
ccoVeille: I wouldn't add things to .git folder.

I would use .cache or build

Then add the folder to .gitignore

ingydotnet: I'll update it to `.cache`.

I'm interested in hearing why you wouldn't put things under the .git directory.

I've been doing that a lot lately in various automations and it's been working out nicely so far.

ingydotnet: Updated in 2cdfdff

ingydotnet:
This should work out for anyone on Linux / Mac if they have `git`, `bash`, `curl` and `make` installed. (minimal)
It self-installs the Makefile parts and then self-installs Go and sets the environment to run it, without affecting the user's system in any other way.

Example output of the self-installing `make` commands:

```
$ make distclean
rm -f -r .cache
$ 
$ make test
Installing 'go' locally
curl -Ls https://go.dev/dl/go1.24.3.linux-amd64.tar.gz > /home/ingy/src/go-yaml/.cache/.local/cache/go1.24.3.linux-amd64.tar.gz
tar -C /home/ingy/src/go-yaml/.cache/.local -xzf /home/ingy/src/go-yaml/.cache/.local/cache/go1.24.3.linux-amd64.tar.gz
mv /home/ingy/src/go-yaml/.cache/.local/go /home/ingy/src/go-yaml/.cache/.local/go-1.24.3
touch /home/ingy/src/go-yaml/.cache/.local/go-1.24.3/bin/go

go test 
OK: 50 passed
PASS
ok  	go.yaml.in/yaml/v3	2.236s
$ 
$ make test
go test 
OK: 50 passed
PASS
ok  	go.yaml.in/yaml/v3	2.245s
$ 
```

ingydotnet:
I need to finish the work for macOS, then I'll take it out of draft.

GNUmakefile
ccoVeille: > I'll update it to `.cache`.
> 
> I'm interested in hearing why you wouldn't put things under the .git directory.
> 
> I've been doing that a lot lately in various automations and it's been working out nicely so far.

Here is the explanation I worked by using Gemini on to explain you

<details>
Could you store temporary or cache files created by my program inside the Git repository's .git folder?

No, you should never store temporary or cache files created by your program directly inside the .git folder of your repository.

Here's why this is a bad practice:
Risk of Repository Corruption

The .git folder is Git's internal database. It holds all the critical information Git needs to manage your project's version history, including commits, branches, and configurations. Introducing arbitrary files into this folder can confuse Git's internal mechanisms, potentially leading to corruption of your repository. This could result in lost history, unexpected errors, or an unstable project.
Unnecessary Bloat and Performance Issues

While files within the .git folder aren't "tracked" in the same way your source code is, placing large temporary or cache files there will still consume significant disk space within your repository. This bloat makes operations like copying, backing up, or even just exploring your repository locally much slower and more resource-intensive than necessary.
No Benefit, Only Downsides

There's simply no advantage to storing your program's temporary data in the .git folder. Your program can easily create its own dedicated directories, such as tmp/ or cache/, within your project's main working directory.
The Correct Approach: Use .gitignore

The proper way to handle temporary and cache files is to store them in a dedicated directory outside the .git folder, typically at the root level of your project (e.g., tmp/, cache/, build/). Once you've created these directories, add their names to your .gitignore file. This tells Git to explicitly ignore these files and directories, preventing them from ever being accidentally staged or committed to your version control history.

Example .gitignore entry:

# Ignore temporary and cache files
tmp/
cache/
*.log
*.swp
*.bak

By keeping your temporary files separate and using .gitignore, you maintain a clean, stable, and efficient Git repository.
</details>

ingydotnet: Nice. Thanks.
I decided to continue the AI conversation: https://gist.github.com/ingydotnet/574e1bf1be75f94961ef04e1de1f943b

These were useful points. I'm surprised it didn't catch the case where `.git` is not a dir (worktrees). My solution still works because I can find the `.git/` dir, but it didn't know that.

To be clear, I won't cache under the `.git/` dir for this repo.
I may or may not continue on my minor projects.

It's fun to try new things. I guess that's how YAML happened in the first place!

ingydotnet:
This is working well on Linux and Mac now.

Ready for review.

ingydotnet:
Had to remove a file. Ready for real now.

carloslima (approved):

GNUmakefile
```diff
@@ -0,0 +1,17 @@
...
+$(shell [ -d $M ] || ( \
+  git clone -q https://github.com/makeplus/makes $M && \
+  git -C $M reset -q --hard $(M-COMMIT)))
+ifneq ($(shell git -C $M rev-parse $(M-COMMIT)),\
+       $(shell git -C $M rev-parse HEAD))
+$(error $M is not at the correct commit: $(M-COMMIT))
+endif
+
+include $M/init.mk
+include $M/go.mk
```
stefanprodan: Downloading the runtime with Make is very uncommon for Go projects. The Go toolchain knows how to download the version specified in go.mod https://go.dev/doc/toolchain

ingydotnet: This is an optional way to do testing without having Go installed, which is how I prefer to work in as many languages as possible.
I can run these on both Mac and Linux:
```
make test
make test GO-VERSION=...
make shell
make shell GO-VERSION=...
```
You don't need to use this, and we don't need to integrate it into GHA, but I'd like it to be available for those who do.

ingydotnet: @stefanprodan ,

Note that we want a Makefile / GNUmakefile for automating other stuff (like YTS testing).

If you want I can easily change the behavior so that it only installs a local go when `GO-VERSION` is provided.

stefanprodan: > If you want I can easily change the behavior so that it only installs a local go when GO-VERSION is provided.

Yes I think that would be best. Another option would be testing `go version` and download Go only when it's not installed.

ingydotnet: > Yes I think that would be best. Another option would be testing `go version` and download Go only when it's not installed.

Good idea. I'll try that out.

ccoVeille: > > If you want I can easily change the behavior so that it only installs a local go when GO-VERSION is provided.
> 
> Yes I think that would be best. Another option would be testing `go version` and download Go only when it's not installed.

I second this.

ccoVeille: Please note this should be considered instead of `go version`

`go env GOVERSION`

The result is easier to parse

ingydotnet: Fixed in 2acf43e

ingydotnet:
Updated and ready for review. Added support for `make test v=1`. Only installs local go is no go is installed or GO-VERSION=... is used.

stefanprodan (approved):
LGTM

GNUmakefile
```diff
@@ -0,0 +1,46 @@
...
+  git clone -q https://github.com/makeplus/makes $(MAKES) && \
+  git -C $(MAKES) reset -q --hard $(MAKES-COMMIT)))
+ifneq ($(shell git -C $(MAKES) rev-parse HEAD), \
+       $(shell git -C $(MAKES) rev-parse $(MAKES-COMMIT)))
+$(error $(MAKES) is not at the correct commit: $(MAKES-COMMIT))
+endif
+include $(MAKES)/init.mk
+
+# Only auto-install go if no go exists or GO-VERSION specified:
+ifeq (,$(shell which go))
```
ccoVeille, line 15: I tend to use `command -v go` now, but `which go` works well.

GNUmakefile
```diff
@@ -0,0 +1,46 @@
...
+GO-VERSION ?= 1.24.0
```
ccoVeille, line 16: this is good, but maybe you could consider using `go env GOVERSION` that will return the current Go version with a "go" prefix

Here is what I get locally

```console
$ go env GOVERSION
go1.24.4
```

README.md
```diff
@@ -165,6 +165,28 @@ b:
...
+* `bash`
+* `curl`
+* `make`
+
+Some `make` commands are:
+
+* `make test`
+* `make test GO-VERSION=1.2.34`
+* `make shell` Start a shell with the local `go` environment
+* `make shell GO-VERSION=1.2.34`
```
ccoVeille, line 186: to be consistent with Go tooling and ecosystem (GOPATH, GOBIN, GOROOT ...), I would use GOVERSION as a name

I point the documentation so code and documentation would be updated if you agree.

ingydotnet: This is very much on purpose. `GO-VERSION` as used here is a Make variable, not a shell variable.
I don't want to conflict with actual go env vars.

GNUmakefile
ingydotnet: But line 16 only happens when `go` is **not** installed.

GNUmakefile
ingydotnet: That's what I used at first but it didn't work for some reason.
FWIW, `command` is not a command! :smile: 
It's a shell builtin.
```
$ which command
$ which which
/usr/bin/which
$ type command
command is a shell builtin
$ type which
which is hashed (/usr/bin/which)
$ 
```
I might take another look to see why `command -v` didn't work...

GNUmakefile
ingydotnet: That said, I am open to suggestions on what the default version should be here.

GNUmakefile
ingydotnet: So `command -v` _does_ work. I'll make that change next time I need to update the GNUmakefile.

ccoVeille (approved):

PR #39, opened 2025-06-15
Downgrade the go version number in go.mod

wwqgtxx:
The go.mod in the current version is set to 1.22, which will make it impossible for projects that need to be compiled with an older version of golang to reference the package.

It is recommended to downgrade the go line back to 1.12 in the original version so that more projects can switch to the new fork.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,5 @@
 module go.yaml.in/yaml/v3
 
-go 1.22
+go 1.16
 
 require gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405


andig:
Seems this may have been an unintentional change to v3 in https://github.com/yaml/go-yaml/commit/cbdacaa8b43b201f76df30a930fd2e06af9cc9d5, for v4 I wouldn't mind.

wwqgtxx:
From the go.mod files that have been released, the go versions of v2 and v3 are different.
v2.4.2:
https://github.com/yaml/go-yaml/blob/246a95c22c57f15ef6d3305a1f1b8a0b05e4d560/go.mod#L1-L5
v3.0.3:
https://github.com/yaml/go-yaml/blob/148efbbf2f90cec938ae080558d875f29bfaa18d/go.mod#L1-L5
Should they be unified to 1.12 of `kubernetes-sigs/yaml`, or simply delete the line?

This can also make it unnecessary to modify the version number of the downstream project PR (https://github.com/kubernetes-sigs/yaml/pull/133)

liggitt:
Consumers can opt back into older go runtime behaviors with godebug / godefault settings, so I actually think it's reasonable for latest tagged versions to use (at minimum) the oldest still-supported go version, which is currently go 1.23.

wwqgtxx:
> Consumers can opt back into older go runtime behaviors with godebug / godefault settings, so I actually think it's reasonable for latest tagged versions to use (at minimum) the oldest still-supported go version, which is currently go 1.23.

Due to some special requirements, we need to support operating systems that are no longer supported by the new version of golang, such as Win7 and old versions of MacOS. This operation cannot be restored through godebug / godefault. The current go-yaml does not rely on third-party libraries or use newer syntax features. I think it is unnecessary to significantly increase the minimum golang version. This may force us to use the original go-yaml or fork the project again to maintain compatibility.

ingydotnet:
Since we don't seem to have consensus here, let's discuss this in our next meeting. 

@wwqgtxx (anyone else) if you'd like to join the meeting, contact us in the `#go-yaml` channel of the CNCF slack.

stefanprodan:
I don't think we should be supporting EOL Go versions, this will block us from using modern testing packages like Ginkgo https://github.com/onsi/ginkgo/blob/7470c9e6f1c150a3e4162bc5a54b1dc5e8eb3961/go.mod#L3

wwqgtxx:
> I don't think we should be supporting EOL Go versions, this will block us from using modern testing packages like Ginkgo https://github.com/onsi/ginkgo/blob/7470c9e6f1c150a3e4162bc5a54b1dc5e8eb3961/go.mod#L3

In fact, you can split the test into a separate directory `/test` and base it on its own independent go.mod, so that the main project can keep zero external dependencies.

ingydotnet:
@dims, you added the `go 1.22` line here: 
https://github.com/yaml/go-yaml/commit/cbdacaa8b43b201f76df30a930fd2e06af9cc9d5#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R3

What would happen if we got rid of this line for v3?

@carloslima and I think that there are 2 different discussions to be had:
One for v3 and one for main/v4.

We want all projects using go-yaml/yaml v2 and v3 to be able to switch to yaml/go-yaml.

https://go.dev/ref/mod#go-mod-file-go says:
"If the go directive is missing, go 1.16 is assumed."

Also:

> At go 1.21 or higher:
> * The go line declares a required minimum version of Go to use with this module.
> * The go line must be greater than or equal to the go line of all dependencies.
> * The go command no longer attempts to maintain compatibility with the previous older version of Go.
> * The go command is more careful about keeping checksums of go.mod files in the go.sum file.
> 

1.21 was released August 2023, which is long after Gustavo was actively maintaining this.
So not having a `go 1.xx` line would probably work for him.

It seems to us that for yaml/go-yaml to be literally compatible with go-yaml/yaml, we'd want to change v3 to:
```
go 1.16
```

The v2 branch uses `1.15`: 
https://github.com/yaml/go-yaml/blob/v2/go.mod#L3
so nothing to change there.

@wwqgtxx would this work for your situations?

wwqgtxx:
> It seems to us that for yaml/go-yaml to be literally compatible with go-yaml/yaml, we'd want to change v3 to:
> 
> ```
> go 1.16
> ```
> 
> The v2 branch uses `1.15`: https://github.com/yaml/go-yaml/blob/v2/go.mod#L3 so nothing to change there.
> 
> @wwqgtxx would this work for your situations?


This modification is OK for me. I chose to set it to `1.12` in this PR just to be consistent with `kubernetes-sigs/yaml`.

SuperSandro2000:
Setting such an old go version has at least two major drawbacks for everyone:
- You cannot use new language features that are many years old and standard for programmers especially if they only learned go in the last year or two. This also makes it hard to contribute new code.
- If not at least 1.16 is used then I think we require the test dependency for downstream users for no good reason. https://go.dev/ref/mod#go-mod-file-go:~:text=At%20go%201.16%20or,packages%2C%20and%20so%20on.

Also in addition to that: Windows 7 is also EOL and insecure. It is not as bad as Windows XP that you are hacked within hours when browsing the internet but it is on a good way there.

I would recommend that you fork the project as you probably already have done with many other dependencies.

ingydotnet:
> > @wwqgtxx would this work for your situations?
> 
> This modification is OK for me. I chose to set it to `1.12` in this PR just to be consistent with `kubernetes-sigs/yaml`.

@wwqgtxx would you please update the PR to use `go 1.16`.

ingydotnet:
> Should they be unified to 1.12 of `kubernetes-sigs/yaml`, or simply delete the line?

BTW, Just thought I'd mention... When I tried deleting the line `go tidy` added it back.

ingydotnet:
> This modification is OK for me. I chose to set it to `1.12` in this PR just to be consistent with `kubernetes-sigs/yaml`.

@wwqgtxx I'm confused. Isn't kubernetes-sigs/yaml set to 1.22?

https://github.com/kubernetes-sigs/yaml/blob/v1.5.0/go.mod#L3

wwqgtxx:
> > This modification is OK for me. I chose to set it to `1.12` in this PR just to be consistent with `kubernetes-sigs/yaml`.
> 
> @wwqgtxx I'm confused. Isn't kubernetes-sigs/yaml set to 1.22?
> 
> https://github.com/kubernetes-sigs/yaml/blob/v1.5.0/go.mod#L3

That's because the PR submitted by @dims has been merged to increase the version in `kubernetes-sigs/yaml`, which was previously 1.12.

ingydotnet (approved):

ingydotnet:
CodeQL isn't triggering for some reason. Merging this by bypassing the restrictions.

ingydotnet:
v3.0.4 released.

ingydotnet:
> Consumers can opt back into older go runtime behaviors with godebug / godefault settings, so I actually think it's reasonable for latest tagged versions to use (at minimum) the oldest still-supported go version, which is currently go 1.23.

@liggitt We can certainly consider this logic for v4 and beyond, but since this is a new repository that we are encouraging everyone (not just in big things like Kubernetes, but anyone using go-yaml in Go code, anywhere) I think we need zero friction to making the transition.

I'm glad @wwqgtxx brought this up because in general I think most people would be silent if yaml/go-yaml v3 didn't work exactly like go-yaml/yaml out of the gate. They would just not switch.

@SuperSandro2000 and @stefanprodan to your points about using modern Go tooling, with my YAML maintainer hat on, it's also important to move this project forward in parallel with pyyaml and libyaml, which are both also very heavily used and from which go-yaml was derived (and thus shares many of the same issues to fix).

We can and will discuss everything as a team, but for things like choosing a test system, I want the tests to be shared as widely across the YAML landscape as possible. If we can do that while still using Go best practices, then I'm keen.

PR #45, opened 2025-06-26
Handle decoding of `-0` from YAML

carloslima:
This change is necessary because Go constants cannot represent `-0.0`: https://staticcheck.dev/docs/checks/#SA4026

As a result, when parsing `-0` or `-0.0`, the value ends up as `0` or `0.0` in Go.

Before:
```go
> go run a.go
Decoded into Go: 0

Re-encoded YAML:
0
```

After:
```go
> go run x/a.go
Decoded into Go: -0

Re-encoded YAML:
-0
```

Source:
```go
package main

import (
	"fmt"
	"log"

	"go.yaml.in/yaml/v3"
)

func main() {
	input := []byte("-0\n")

	var f float64
		log.Fatal(err)
	}

	out, err := yaml.Marshal(f)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Print(string(out))
}
```

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -31,6 +31,11 @@ import (
 	"go.yaml.in/yaml/v3"
 )
 
+// negativeZero represents -0.0 for YAML test cases
+// this is needed because Go constants cannot express -0.0
+// https://staticcheck.dev/docs/checks/#SA4026
+var negativeZero = math.Copysign(0.0, -1.0)
+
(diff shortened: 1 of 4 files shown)


stefanprodan (approved):
LGTM

Thanks @carloslima

ingydotnet (approved):

PR #46, opened 2025-06-26
Handle roundtrip of "\t\n"

carloslima:
This fixes a go-yaml issue where strings like `\t\n` could be encoded but not decoded properly.

Before
```
~/c/go-yaml-2 (main)> go run x/b.go
Initial string: "\t\n"

Encoded into YAML:
|
        
2025/06/27 00:54:57 yaml: line 2: found a tab character where an indentation space is expected
exit status 1

~/c/go-yaml-2 (main) [1]>
```

After
```
~/c/go-yaml-2 (main) [1]> git checkout handle-tab-string
Switched to branch 'handle-tab-string'

carlos@hk-carlos ~/c/go-yaml-2 (handle-tab-string)> go run x/b.go
Initial string: "\t\n"

Encoded into YAML:
|
        
Decoded into Go: "\t\n"

```

Source
```
~/c/go-yaml-2 (handle-tab-string)> cat x/b.go 
package main

import (
	"fmt"
	"log"

	"go.yaml.in/yaml/v3"
)

func main() {
	initial := "\t\n"
	fmt.Printf("Initial string: %q\n", initial)

	out, err := yaml.Marshal(initial)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println("\nEncoded into YAML:")
	fmt.Print(string(out))

	var f string
	if err := yaml.Unmarshal([]byte(out), &f); err != nil {
		log.Fatal(err)
	}
}
```

### DIFF
--- a/scannerc.go
+++ b/scannerc.go
@@ -2401,7 +2401,7 @@ func yaml_parser_scan_block_scalar_breaks(parser *yaml_parser_t, indent *int, br
 		}
 
 		// Check for a tab character messing the indentation.
-		if (*indent == 0 || parser.mark.column < *indent) && is_tab(parser.buffer, parser.buffer_pos) {
+		if (parser.mark.column < *indent) && is_tab(parser.buffer, parser.buffer_pos) {
 			return yaml_parser_set_scanner_error(parser, "while scanning a block scalar",
 				start_mark, "found a tab character where an indentation space is expected")
 		}
--- a/decode_test.go
+++ b/decode_test.go
@@ -84,12 +84,18 @@ var unmarshalTests = []struct {
 	}, {
 		"v: -.1",
 		map[string]interface{}{"v": -0.1},
+	}, {
+		"a: |\n    \t\n    \t\n",
+		map[string]string{"a": "\t\n\t\n"},
 	},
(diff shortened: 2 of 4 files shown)


ingydotnet:
I think the right fix here is to no allow that input to be marshaled as a literal scalar.
A YAML dumper (marshaler) has to decide which scalar style to use.
This uses a set of heuristics.
We should tighten what is allowed to become a literal scalar.
Only double quoted style scalars can encode any possible value.

scannerc.go
```diff
@@ -2400,7 +2400,7 @@ func yaml_parser_scan_block_scalar_breaks(parser *yaml_parser_t, indent *int, br
 		}
 
 		// Check for a tab character messing the indentation.
-		if (*indent == 0 || parser.mark.column < *indent) && is_tab(parser.buffer, parser.buffer_pos) {
+		if (parser.mark.column < *indent) && is_tab(parser.buffer, parser.buffer_pos) {
```
ingydotnet, line 2403: I'm not sure we want to relax this check.

ingydotnet:
@carloslima Can you close this now?

PR #47, opened 2025-06-27
bug: Fix error unmarshalling alias/anchor with non alphanumeric name

### LINKED ISSUE #44: [Bug] Unmarshal does not support the full range of valid characters as specified in the spec for anchors and aliases
As discussed with @ingydotnet; opening here after discussion on mikefarah/yq#2406 as yq depends on this library and the bug has been confirmed to be related to this library.

## Describe the bug

In the yq application, it is possible render/resolve anchors/aliases through the `explode` operator.

Unmarshalling fails due to parsing errors in this library.

Other used library has been confirmed not being the cause.

## Reproduction

Reproduced in https://github.com/TyIsI/fafo-go-yaml/blob/main/yaml-go-yaml/main.go

### Input Yaml

Message.yml:
```yaml
references:
    - &network.core.v1.Ping ./network/core/v1/Ping.yml
    - &network.core.v1.Pong ./network/core/v1/Pong.yml
    - &network.enums.v1.MessageType ./network/enums/v1/MessageType.yml
---
Message:
    - - type: *network.enums.v1.MessageType
        payload: *network.core.v1.Ping
      - type: *network.enums.v1.MessageType
        payload: *network.core.v1.Pong
```

### Command

``` bash
yq -P eval-all "explode(.) | select(document_index == 1)" Message.yml
```

### Actual behavior

```
Error: bad file 'Message.yml': yaml: line 2: did not find expected alphabetic or numeric character
```

### Expected behavior

```yaml
Message:
  - - type: ./network/enums/v1/MessageType.yml
      payload: ./network/core/v1/Ping.yml
    - type: ./network/enums/v1/MessageType.yml
      payload: ./network/core/v1/Pong.yml
```

### Additional context
If I remove the dots and use dashes (-) or slashes (/) instead, it also doesn't work. Single string works fine.

Easy link to the spec:
https://yaml.org/spec/1.2.2/#rule-ns-anchor-name

carloslima:
closes #44

This addresses the parser being overly restrictive on acceptable anchor/alias names.

Before
```
~/c/go-yaml-2 (main)> go run x/a.go
2025/06/28 06:57:37 yaml: line 3: did not find expected alphabetic or numeric character
exit status 1
```

After
```
~/c/go-yaml-2 (main) [1]> git checkout fix-anchor-names 
Switched to branch 'fix-anchor-names'
~/c/go-yaml-2 (fix-anchor-names)> go run x/a.go
Decoded into Go: map[string]interface {}{"Message":[]interface {}{[]interface {}{map[string]interface {}{"payload":"./network/core/v1/Ping.yml", "type":"./network/enums/v1/MessageType.yml"}, map[string]interface {}{"payload":"./network/core/v1/Pong.yml", "type":"./network/enums/v1/MessageType.yml"}}}, "references":[]interface {}{"./network/core/v1/Ping.yml", "./network/core/v1/Pong.yml", "./network/enums/v1/MessageType.yml"}}

Re-encoded YAML:
Message:
    - - payload: ./network/core/v1/Ping.yml
        type: ./network/enums/v1/MessageType.yml
      - payload: ./network/core/v1/Pong.yml
        type: ./network/enums/v1/MessageType.yml
references:
    - ./network/core/v1/Ping.yml
    - ./network/core/v1/Pong.yml
    - ./network/enums/v1/MessageType.yml
```
Source
```
~/c/go-yaml-2 (fix-anchor-names)> cat x/a.go
package main

import (
	"fmt"
	"log"

	"go.yaml.in/yaml/v3"
)

func main() {
	input := []byte(`---
references:
    - &network.core.v1.Ping ./network/core/v1/Ping.yml
    - &network.core.v1.Pong ./network/core/v1/Pong.yml
    - &network.enums.v1.MessageType ./network/enums/v1/MessageType.yml
Message:
    - - type: *network.enums.v1.MessageType
        payload: *network.core.v1.Ping
      - type: *network.enums.v1.MessageType
        payload: *network.core.v1.Pong
`)
	var f map[string]interface{}
		log.Fatal(err)
	}

	out, err := yaml.Marshal(f)
	// out, err := yaml.Marshal("\t\n")
	if err != nil {
		log.Fatal(err)
	}
	fmt.Print(string(out))
}
```

### DIFF
--- a/node_test.go
+++ b/node_test.go
@@ -1110,6 +1110,59 @@ var nodeTests = []struct {
 			}},
 		},
 	}, {
+		"a: &anchor(.!@#$%^&*+=?:;)name [1, 2]\nb: *anchor(.!@#$%^&*+=?:;)name\n",
+		yaml.Node{
+			Kind:   yaml.DocumentNode,
+			Line:   1,
+			Column: 1,
+			Content: []*yaml.Node{{
+				Kind:   yaml.MappingNode,
+				Line:   1,
+				Column: 1,
+				Tag:    "!!map",
+				Content: []*yaml.Node{{
+					Kind:   yaml.ScalarNode,
+					Value:  "a",
+					Tag:    "!!str",
+					Line:   1,
+					Column: 1,
+				},
+					saveNode("anchor(.!@#$%^&*+=?:;)name", &yaml.Node{
+						Kind:   yaml.SequenceNode,
+						Style:  yaml.FlowStyle,
+						Tag:    "!!seq",
+						Anchor: "anchor(.!@#$%^&*+=?:;)name",
+						Line:   1,
+						Column: 4,
+						Content: []*yaml.Node{{
+							Kind:   yaml.ScalarNode,
+							Value:  "1",
+							Tag:    "!!int",
+							Line:   1,
+							Column: 33,
+						}, {
+							Kind:   yaml.ScalarNode,
+							Value:  "2",
+							Tag:    "!!int",
+							Line:   1,
+							Column: 36,
+						}},
+					}), {
+						Kind:   yaml.ScalarNode,
+						Value:  "b",
+						Tag:    "!!str",
+						Line:   2,
+						Column: 1,
+					}, {
+						Kind:   yaml.AliasNode,
+						Value:  "anchor(.!@#$%^&*+=?:;)name",
+						Alias:  dropNode("anchor(.!@#$%^&*+=?:;)name"),
+						Line:   2,
+						Column: 4,
+					}},
+			}},
+		},
+	}, {
 
 		"# One\n# Two\ntrue # Three\n# Four\n# Five\n",
 		yaml.Node{
--- a/yamlprivateh.go
+++ b/yamlprivateh.go
@@ -46,7 +46,33 @@ const (
 // Check if the character at the specified position is an alphabetical
 // character, a digit, '_', or '-'.
 func is_alpha(b []byte, i int) bool {
-	return b[i] >= '0' && b[i] <= '9' || b[i] >= 'A' && b[i] <= 'Z' || b[i] >= 'a' && b[i] <= 'z' || b[i] == '_' || b[i] == '-'
+	return b[i] >= '0' && b[i] <= '9' || b[i] >= 'A' && b[i] <= 'Z' ||
+		b[i] >= 'a' && b[i] <= 'z' || b[i] == '_' || b[i] == '-'
+}
+
+// Check if the character at the specified position is a flow indicator as
+// defined by spec production [23] c-flow-indicator ::=
+// c-collect-entry | c-sequence-start | c-sequence-end |
+// c-mapping-start | c-mapping-end
+func is_flow_indicator(b []byte, i int) bool {
+	return b[i] == '[' || b[i] == ']' ||
+		b[i] == '{' || b[i] == '}' || b[i] == ','
+}
+
+// Check if the character at the specified position is valid for anchor names
+// as defined by spec production [102] ns-anchor-char ::= ns-char -
+// c-flow-indicator.
+// This includes all printable characters except: CR, LF, BOM, space, tab, '[',
+// ']', '{', '}', ','.
+// We further limit it to ascii chars only, which is a subset of the spec
+// production but is usually what most people expect.
+func is_anchor_char(b []byte, i int) bool {
+	return is_printable(b, i) &&
+		!is_break(b, i) &&
+		!is_blank(b, i) &&
+		!is_bom(b, i) &&
+		!is_flow_indicator(b, i) &&
+		is_ascii(b, i)
 }
 
 // Check if the character at the specified position is a digit.
@@ -61,7 +87,8 @@ func as_digit(b []byte, i int) int {
 
 // Check if the character at the specified position is a hex-digit.
 func is_hex(b []byte, i int) bool {
-	return b[i] >= '0' && b[i] <= '9' || b[i] >= 'A' && b[i] <= 'F' || b[i] >= 'a' && b[i] <= 'f'
+	return b[i] >= '0' && b[i] <= '9' || b[i] >= 'A' && b[i] <= 'F' ||
+		b[i] >= 'a' && b[i] <= 'f'
 }
 
 // Get the value of a hex-digit.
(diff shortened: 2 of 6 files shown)


yamlprivateh.go
```diff
@@ -49,6 +49,18 @@ func is_alpha(b []byte, i int) bool {
...
 
+// Check if the character at the specified position is valid for anchor names
+// as defined by spec production [102] ns-anchor-char ::= ns-char - c-flow-indicator.
+// This includes all printable characters except: CR, LF, BOM, space, tab, '[', ']', '{', '}', ','.
+func is_anchor_char(b []byte, i int) bool {
+	if !is_printable(b, i) || is_break(b, i) || is_blank(b, i) || is_bom(b, i) ||
+		b[i] == '[' || b[i] == ']' || b[i] == '{' || b[i] == '}' || b[i] == ',' {
+		return false
+	}
+
```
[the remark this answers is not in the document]
stefanprodan: This suggestion makes sense, code is more readable.

node_test.go
```diff
@@ -985,6 +985,59 @@ var nodeTests = []struct {
 			}},
 		},
 	}, {
+		"a: &anchor(.!@#$%^&*+=?:;)name [1, 2]\nb: *anchor(.!@#$%^&*+=?:;)name\n",
```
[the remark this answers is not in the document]
carloslima: intentional and matches all the other tests

yamlprivateh.go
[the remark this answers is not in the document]
carloslima: I went with a slightly different code which I think achieves the same result while being aligned with all the other checks. Can you take a look? :)

stefanprodan (approved):
LGTM

Thanks @carloslima

yamlprivateh.go
```diff
@@ -49,6 +49,24 @@ func is_alpha(b []byte, i int) bool {
 	return b[i] >= '0' && b[i] <= '9' || b[i] >= 'A' && b[i] <= 'Z' || b[i] >= 'a' && b[i] <= 'z' || b[i] == '_' || b[i] == '-'
 }
 
+// Check if the character at the specified position is a flow indicator as defined
+// by spec production [23] c-flow-indicator ::=
+// c-collect-entry | c-sequence-start | c-sequence-end | c-mapping-start | c-mapping-end
+func is_flow_indicator(b []byte, i int) bool {
+	return b[i] == '[' || b[i] == ']' || b[i] == '{' || b[i] == '}' || b[i] == ','
+}
```
ccoVeille: This would be safer

```suggestion
func is_flow_indicator(b []byte, i int) bool {
	c, ok := b[i]
	if !ok {
		return false
	}
	return c == '[' || c == ']' || c == '{' || c == '}' || c == ','
}
```

But apparently, all code is like that. Maybe it could be addressed later for other method, but here it won't hurt

yamlprivateh.go
```diff
@@ -49,6 +49,24 @@ func is_alpha(b []byte, i int) bool {
 	return b[i] >= '0' && b[i] <= '9' || b[i] >= 'A' && b[i] <= 'Z' || b[i] >= 'a' && b[i] <= 'z' || b[i] == '_' || b[i] == '-'
 }
 
+// Check if the character at the specified position is a flow indicator as defined
+// by spec production [23] c-flow-indicator ::=
+// c-collect-entry | c-sequence-start | c-sequence-end | c-mapping-start | c-mapping-end
+func is_flow_indicator(b []byte, i int) bool {
+	return b[i] == '[' || b[i] == ']' || b[i] == '{' || b[i] == '}' || b[i] == ','
```
ccoVeille: Don't we have constants for these ?

ccoVeille: I would use a switch also

```suggestion
	switch b[i] {
		case '[', ']', '{', '}', ',':
		return true
	}
	return false
}
```

https://pkg.go.dev/bytes#ContainsAny (Go 1.7) could be considered maybe.

yamlprivateh.go
```diff
@@ -49,6 +49,24 @@ func is_alpha(b []byte, i int) bool {
 	return b[i] >= '0' && b[i] <= '9' || b[i] >= 'A' && b[i] <= 'Z' || b[i] >= 'a' && b[i] <= 'z' || b[i] == '_' || b[i] == '-'
 }
 
+// Check if the character at the specified position is a flow indicator as defined
+// by spec production [23] c-flow-indicator ::=
+// c-collect-entry | c-sequence-start | c-sequence-end | c-mapping-start | c-mapping-end
+func is_flow_indicator(b []byte, i int) bool {
```
ccoVeille, line 57: I'm surprised code works with byte and not rune

Maybe it's OK because of the spec that limit to ASCII only.

But I'm curious how lib handle Unicode character in a string node or comment

ccoVeille (approved):
Some minor remarks, mostly based on the fact I'm discovering code.

otherwise LGTM 👍

yamlprivateh.go
carloslima: That's because the validation for the buffer and the position in the buffer are handled elsewhere. By the time this (or the other helper functions are called) the buffer and the position are guaranteed to be valid.

yamlprivateh.go
carloslima: I followed the style of the other functions surrounding it and went with the `if` approach instead of a `switch`. 

Regarding ContainsAny, we'd need to look closer. Those helpers are in the hot path of the parser.

carloslima: I don't think so

yamlprivateh.go
carloslima: At this point it's looking for markers, which are single byte. At other states the parser reads a character, at that point it calls `width(b byte){...}` in `yamlprivateh.go` to decide how many bytes to read. Follow those calls and you should get the answer you are looking for.

yamlprivateh.go
ccoVeille: > I followed the style of the other functions surrounding it and went with the `if` approach instead of a `switch`. 

We are inheriting old code, so your code is fine in the context of existing code.

I feel like Go linters would suggest to replace things. But we can do it later.

> Regarding ContainsAny, we'd need to look closer. Those helpers are in the hot path of the parser.

You might be right and it would be slower.

ccoVeille (approved):

stefanprodan (approved):
LGTM

Thanks @carloslima 🏅

ccoVeille (approved):

PR #48, opened 2025-06-27
Handle roundtrip for `{"<<": []}`

carloslima:
This fixes a go-yaml issue where the string `<<` was encoded unquoted and treated as a merge key when decoded.

Before
```
~/c/go-yaml-2 (main)> go run x/b.go
Initial value: map[string]interface {}{"<<":[]interface {}{}}

Encoded into YAML:
<<: []
Decoded into Go: map[string]interface {}{}
```

After
```
carlos@hk-carlos ~/c/go-yaml-2 (main)> git checkout fix-merge-key
Switched to branch 'fix-merge-key'
carlos@hk-carlos ~/c/go-yaml-2 (fix-merge-key)> go run x/b.go
Initial value: map[string]interface {}{"<<":[]interface {}{}}

Encoded into YAML:
"<<": []
Decoded into Go: map[string]interface {}{"<<":[]interface {}{}}
```

Source
```
carlos@hk-carlos ~/c/go-yaml-2 (fix-merge-key)> cat x/b.go
package main

import (
	"fmt"
	"log"

	"go.yaml.in/yaml/v3"
)

func main() {
	initial := map[string]interface{}{"<<": []interface{}{}}
	fmt.Printf("Initial value: %#v\n", initial)

	out, err := yaml.Marshal(initial)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println("\nEncoded into YAML:")
	fmt.Print(string(out))

	var f map[string]interface{}
	if err := yaml.Unmarshal([]byte(out), &f); err != nil {
		log.Fatal(err)
	}
}
```

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -321,6 +321,10 @@ func isOldBool(s string) (result bool) {
 	}
 }
 
+func looksLikeMerge(s string) (result bool) {
+	return s == "<<"
+}
+
 func (e *encoder) stringv(tag string, in reflect.Value) {
 	var style yaml_scalar_style_t
 	s := in.String()
@@ -342,7 +346,7 @@ func (e *encoder) stringv(tag string, in reflect.Value) {
 		// tag when encoded unquoted. If it doesn't,
 		// there's no need to quote it.
 		rtag, _ := resolve("", s)
-		canUsePlain = rtag == strTag && !(isBase60Float(s) || isOldBool(s))
+		canUsePlain = rtag == strTag && !(isBase60Float(s) || isOldBool(s) || looksLikeMerge(s))
 	}
 	// Note: it's possible for user code to emit invalid YAML
 	// if they explicitly specify a tag and a string containing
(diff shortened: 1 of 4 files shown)


encode.go
```diff
@@ -321,6 +321,10 @@ func isOldBool(s string) (result bool) {
 	}
 }
 
+func looksLikeMerge(s string) (result bool) {
```
stefanprodan, line 328: Maybe add a comment here to explain why is this needed.

carloslima: Added. Does it read ok?

stefanprodan (approved):
LGTM

Thanks @carloslima

ccoVeille:
Can you refer the go-yaml issue you are referring to?

carloslima:
> Can you refer the go-yaml issue you are referring to?

There is no Github issue, if that's what you're asking for. Sorry for the overloaded word, I meant "problem".

The definition of the problem is what's in the description: that a `map[string]any{"<<": true}` doesn't roundtrip correctly.
See: https://go.dev/play/p/VzPffHRhktN

PR #49, opened 2025-06-28
bug: fix handling of '?' in unquoted values

### LINKED ISSUE #36: Error parsing valid yaml with "?"
Hello!

Fantastic to see this library being maintained by the official YAML org!

I'm the author of [yq](https://github.com/mikefarah/yq); which makes heavy use of this library 😄  - there are a number of little issues people have raised over the years, not sure what the plan is with this library in terms of maintenance and getting things fixed and released?

Anyway here's a sample issue in unmarshalling based off an [issue](https://github.com/mikefarah/yq/issues/2294) raised in yq:

```
foo: {bar: a?bc}
```

https://go.dev/play/p/5C349_xJVce

Copying from the yq issue - it's worth noting that:

Yaml spec didn't mentioned anything about ? being special so it should be fine to be unquoted. https://yaml.org/spec/1.2.2/#822-block-mappings
I've tried a bunch of online yaml parser (like https://onlineyamltools.com/validate-yaml, https://jsonformatter.org/yaml-validator, https://yamlchecker.com/ and https://www.yamllint.com/) and they are fine with it.
Libraries like js-yaml (npm) and serde_yaml (rust) has no problem to parse this as well

carloslima:
closes #36

Fix the conditions for considering `?` as key indicator.

Before
```
~/c/go-yaml-2 (main)> go run x/a.go
Initial value: "foo: {bar: a?bc}"
2025/06/28 08:20:26 yaml: did not find expected ',' or '}'
exit status 1
```

After
```
~/c/go-yaml-2 (main) [1]> git checkout fix-question-mark-value 
Switched to branch 'fix-question-mark-value'
~/c/go-yaml-2 (fix-question-mark-value)> go run x/a.go
Initial value: "foo: {bar: a?bc}"
Decoded into Go: map[string]interface {}{"foo":map[string]interface {}{"bar":"a?bc"}}

Re-encoded YAML:
foo:
    bar: a?bc
```

Source
```
~/c/go-yaml-2 (fix-question-mark-value)> cat x/a.go
package main

import (
	"fmt"
	"log"

	"go.yaml.in/yaml/v3"
)

func main() {
	input := []byte("foo: {bar: a?bc}")
	fmt.Printf("Initial value: %q\n", input)
	var f map[string]interface{}
		log.Fatal(err)
	}

	out, err := yaml.Marshal(f)
	if err != nil {
		log.Fatal(err)
	}
	fmt.Print(string(out))
}
```

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -815,6 +815,39 @@ var unmarshalTests = []struct {
 			"c": []interface{}{"d", "e"},
 		},
 	},
+	// bug: question mark in value
+	{
+		"foo: {ba?r: a?bc}",
+		map[string]interface{}{
+			"foo": map[string]interface{}{"ba?r": "a?bc"},
+		},
+	}, {
+		"foo: {?bar: ?abc}",
+		map[string]interface{}{
+			"foo": map[string]interface{}{"?bar": "?abc"},
+		},
+	}, {
+		"foo: {bar?: abc?}",
+		map[string]interface{}{
+			"foo": map[string]interface{}{"bar?": "abc?"},
+		},
+	}, {
+		"foo: {? key: value}",
+		map[string]interface{}{
+			"foo": map[string]interface{}{"key": "value"},
+		},
+	}, {
+		`---
+foo:
+  ? complex key
+  : complex value
+ba?r: a?bc
+`,
+		map[string]interface{}{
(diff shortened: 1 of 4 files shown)


stefanprodan (approved):
LGTM

Thanks @carloslima 🥇

ingydotnet (approved):

PR #51, opened 2025-07-03
Validate conformance to the yaml spec using yaml-test-suite

carloslima:
This PR integrates the official [YAML test suite](https://github.com/yaml/yaml-test-suite), improving test coverage and ensuring compliance with the YAML specification. It also maintains a list of known failing tests to track outstanding issues and prevent regressions.

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -37,7 +37,7 @@ jobs:
         go-version: ${{ matrix.go-versions }}
     - name: Run go vet
       run: go vet ./...
-    - name: Run go test
-      run: GO111MODULE=on go test -v -race ./...
     - name: Run yaml-test-suit tests
       run: make test-yts
+    - name: Run go test
+      run: GO111MODULE=on go test -v -race .


.github/workflows/go.yaml
```diff
@@ -39,3 +39,5 @@ jobs:
       run: go vet ./...
     - name: Run go test
       run: GO111MODULE=on go test -v -race ./...
+    - name: Run yaml-test-suit tests
+      run: GO111MODULE=on make yts
```
ccoVeille: Comsidering go.mod use Go 1.22, I don't think there is a need for using the old env variable


```suggestion
      run: make yts
```

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "yaml-test-suite/testdata/data-2022-01-17"]
+	path = yaml-test-suite/testdata/data-2022-01-17
```
ccoVeille: What is the reason behind usage of a path about 2022-01-17 ?

I mean if there is a need to state it's old, using 2022 would be enough, no?

To me here it looks odd.

Or maybe there is a need for another folder, now or later that will contain a new test suite.

But for now, I wouldn't use a date in the path.

yaml.go
```diff
@@ -701,3 +701,112 @@ func isZero(v reflect.Value) bool {
...
+		case yaml_SINGLE_QUOTED_SCALAR_STYLE:
+			b.WriteString(" '")
+		case yaml_DOUBLE_QUOTED_SCALAR_STYLE:
+			b.WriteString(" \"")
+		}
+		// Escape special characters for consistent event output.
+		val := strings.NewReplacer(
+			"\\", "\\\\",
+			"\n", "\\n",
+			"\t", "\\t",
```
ccoVeille: Go strings literals can be helpful here


```suggestion
			b.WriteString(` "`)
		}
		// Escape special characters for consistent event output.
		val := strings.NewReplacer(
			`\`, `\\`,
			"\n", `\n`,
			"\t", `\t`,
```

.gitmodules
carloslima: It's the release version of yaml-test-suite we are testing against: https://github.com/yaml/yaml-test-suite/tree/data-2022-01-17

I wanted that to be stated explicitly somewhere and that's what I came up with. I didn't like it much but I was out of ideas :sweat_smile:

ccoVeille: Naming is a pain 😅

Let's see if someone among the reviewers has an idea

yaml-test-suite/test_suite_test.go
```diff
@@ -0,0 +1,230 @@
+package yaml_test
```
stefanprodan: The package name collides with the one in the main package, I suggest naming it `yts`. If the intention is to move all the tests from main to a dedicated dir, then let's rename `yaml-test-suite` to just `test` and `test_suite_test.go` to `yaml_suite_test.go`.

carloslima: I initially set it up as a separate test in its own directory, but now that you mention it, it probably makes more sense to move it to the root directory and include it in the regular test run. I don't see any reason why the conformance tests that are already passing shouldn't be part of the main test suite, especially to help catch regressions. What do you think?

carloslima: I applied your name change. If we want to move the files we can move in another PR.

.gitmodules
carloslima: I worked with @ingydotnet on the Makefile and we made it a bit more clear there while keeping the same path.

GNUmakefile
```diff
@@ -0,0 +1,44 @@
...
+$(shell [ -d $M ] || ( \
+  git clone -q https://github.com/makeplus/makes $M && \
+  git -C $M reset -q --hard $(M-COMMIT)))
+ifneq ($(shell git -C $M rev-parse $(M-COMMIT)),\
+       $(shell git -C $M rev-parse HEAD))
+$(error $M is not at the correct commit: $(M-COMMIT))
+endif
+
+include $M/init.mk
+include $M/go.mk
```
stefanprodan: This installs Go locally, but we shouldn't be doing that in tests as it conflicts with the GitHub Actions where each test run uses a different Go version to ensure compatibility.

It's very uncommon for Go projects to download the runtime to run tests. Why are we doing this?

ingydotnet: The GNUmakefile #38 is an optional way to do developer testing, which I prefer.

I wasn't intending it for GHA.

We can either wire up the YTS testing directly in GHA, or if we want to use the GNUmakefile to run the YTS tests in GHA, I can make it not install a local go in that context.

ingydotnet: Addressed by #38 which is merged with this on https://github.com/yaml/go-yaml/blob/PR/51/GNUmakefile

stefanprodan (approved):
LGTM

PS. I would squash the commits before merge.

carloslima:
> PS. I would squash the commits before merge.

woops, I didn't want to squash all of them but there were a couple that should have been. Sorry, I was a bit trigger happy and clicked the button before seeing the comment :sweat_smile:

andig:
This breaks simply running `go test`:

```
got ./...
ok  	go.yaml.in/yaml/v3	(cached)
--- FAIL: TestYAMLSuite (0.00s)
    test_suite_test.go:58: Failed to read directory ./testdata/data-2022-01-17: open ./testdata/data-2022-01-17: no such file or directory
FAIL
FAIL	go.yaml.in/yaml/v3/yts	0.225s
FAIL
```

It might be nice to skip this test if test suite is not cloned?

carloslima:
Skipping would be misleading because you would expect a test pass to mean that all tests passed. 

I opened a PR (https://github.com/yaml/go-yaml/pull/58) to make the error message clear about the problem and the step that is needed to fix it.

PR #52, opened 2025-07-04
Replace gopkg.in/check.v1 with the standard testing library

### LINKED ISSUE #21: Refactor tests to use vanilla testing package plus github.com/stretchr/testify
Some problems with https://gopkg.in/check.v1:

* It is unmaintained. Last commit 5 years ago.
* It is not particularly popular. Most projects use vanilla `testing`, sometimes aided by third-party helpers such as https://github.com/stretchr/testify
* It doesn't support subtests. For example, should any of the table-driven  `unmarshalTests` test cases fail, the entire `TestUnmarshal` test is immediately aborted, and you can't estimate how broken your patch is :).
* It doesn't distinguish `Error` (mark failed)  vs. `Fatal` (mark failed and abort). (although in most cases you'd want to structure different checks as subtests, but _occasionally_ it doesn't worth it)
* It doesn't support `go test -run` to run tests/subtests selectively.
* `go test -v` doesn't show captured output.

I've had good experience with `testing` +`github.com/stretchr/testify` combo in several closed source projects.

Some examples how existing tests might change:

```go
func TestUnmarshal(t *testing.T) {
	for _, item := range unmarshalTests {
		t.Run("", func(t *testing.T) {
			t.Logf("%q", item.data)
			value := reflect.New(reflect.ValueOf(item.value).Type())
			err := yaml.Unmarshal([]byte(item.data), value.Interface())
			if _, ok := err.(*yaml.TypeError); !ok {
				require.NoError(t, err) // fail and abort
			}
			assert.Equalf(t, item.value, value.Elem().Interface(), "error: %v", err)
		})
	}
}

/*
func (s *S) TestUnmarshal(c *C) {
	for i, item := range unmarshalTests {
		c.Logf("test %d: %q", i, item.data)
		t := reflect.ValueOf(item.value).Type()
		value := reflect.New(t)
		err := yaml.Unmarshal([]byte(item.data), value.Interface())
		if _, ok := err.(*yaml.TypeError); !ok {
			c.Assert(err, IsNil)
		}
		c.Assert(value.Elem().Interface(), DeepEquals, item.value, Commentf("error: %v", err))
	}
}
*/
```

Failure output example, showcasing multiple failing subtests: (yeah, diff output is somewhat excessive, but it can be useful in more complicated tests cases)
```
--- FAIL: TestUnmarshal (0.01s)
    --- FAIL: TestUnmarshal/#149 (0.00s)
        decode_test.go:830: "a: 2015-01-01\n"
        decode_test.go:836:
            	Error Trace:	/home/wgh/projects/_forks/go-yaml/decode_test.go:836
            	Error:      	Not equal:
            	            	expected: map[string]time.Time{"a":time.Date(2016, time.January, 1, 0, 0, 0, 0, time.UTC)}
            	            	actual  : map[string]time.Time{"a":time.Date(2015, time.January, 1, 0, 0, 0, 0, time.UTC)}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -3,3 +3,3 @@
            	            	   wall: (uint64) 0,
            	            	-  ext: (int64) 63587203200,
            	            	+  ext: (int64) 63555667200,
            	            	   loc: (*time.Location)(<nil>)
            	Test:       	TestUnmarshal/#149
            	Messages:   	error: <nil>
    --- FAIL: TestUnmarshal/#171 (0.00s)
        decode_test.go:830: "a: b\r\nc:\r\n- d\r\n- e\r\n"
        decode_test.go:836:
            	Error Trace:	/home/wgh/projects/_forks/go-yaml/decode_test.go:836
            	Error:      	Not equal:
            	            	expected: map[string]interface {}{"a":"b", "c":[]interface {}{"d", "e", "f"}}
            	            	actual  : map[string]interface {}{"a":"b", "c":[]interface {}{"d", "e"}}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -2,6 +2,5 @@
            	            	  (string) (len=1) "a": (string) (len=1) "b",
            	            	- (string) (len=1) "c": ([]interface {}) (len=3) {
            	            	+ (string) (len=1) "c": ([]interface {}) (len=2) {
            	            	   (string) (len=1) "d",
            	            	-  (string) (len=1) "e",
            	            	-  (string) (len=1) "f"
            	            	+  (string) (len=1) "e"
            	            	  }
            	Test:       	TestUnmarshal/#171
            	Messages:   	error: <nil>
```

I'm willing to undertake this refactoring, if maintainers agree with me.

carloslima:
closes #21


encode_test.go
```diff
@@ -509,46 +508,68 @@ var marshalTests = []struct {
 	},
 }
 
-func (s *S) TestMarshal(c *C) {
+func TestMarshal(t *testing.T) {
 	defer os.Setenv("TZ", os.Getenv("TZ"))
 	os.Setenv("TZ", "UTC")
```
ccoVeille, line 513: Using t.Setenv should be considered

encode_test.go
```diff
@@ -509,46 +508,68 @@ var marshalTests = []struct {
...
 }
 
-func (s *S) TestEncoderWriteError(c *C) {
+func TestEncoderWriteError(t *testing.T) {
 	enc := yaml.NewEncoder(errorWriter{})
 	err := enc.Encode(map[string]string{"a": "b"})
-	c.Assert(err, ErrorMatches, `yaml: write error: some write error`) // Data not flushed yet
+	if err == nil || !strings.Contains(err.Error(), `yaml: write error: some write error`) {
+		t.Fatalf("Encode() returned %v, want error containing %q", err, `yaml: write error: some write error`)
+	}
```
ccoVeille, line 572: I would introduce a variable here


```suggestion
	expected := `yaml: write error: some write error`
	if err == nil || !strings.Contains(err.Error(), expected) {
		t.Fatalf("Encode() returned %v, want error containing %q", err, expected)
	}
```

I'm a bit worried error messages detected and report may divert over the time with code changes

node_test.go
```diff
@@ -2621,16 +2622,16 @@ var nodeTests = []struct {
 	},
 }
 
-func (s *S) TestNodeRoundtrip(c *C) {
+func TestNodeRoundtrip(t *testing.T) {
 	defer os.Setenv("TZ", os.Getenv("TZ"))
 	os.Setenv("TZ", "UTC")
```
ccoVeille, line 2627: Same here t.Setenv could be considered

node_test.go
```diff
@@ -2863,57 +2890,87 @@ var nodeEncodeDecodeTests = []struct {
...
+	}
+	if v != nil {
+		t.Fatalf("Decoded value mismatch.\nGot: %+v\nWant: nil", v)
+	}
 
 	// ... and even when looking for its tag.
-	c.Assert(n.ShortTag(), Equals, "!!null")
+	if n.ShortTag() != "!!null" {
+		t.Fatalf("ShortTag mismatch. Got: %q, Want: %q", n.ShortTag(), "!!null")
+	}
```
ccoVeille, line 2941: Here also I would recommend using a variable for "!!null", as it is used for detection and reporting

node_test.go
```diff
@@ -2863,57 +2890,87 @@ var nodeEncodeDecodeTests = []struct {
...
 	n.Line = 1
 	_, err = yaml.Marshal(&n)
-	c.Assert(err, ErrorMatches, "yaml: cannot encode node with unknown kind 0")
-	c.Assert(n.Decode(&v), ErrorMatches, "yaml: cannot decode node with unknown kind 0")
+	if err == nil || !regexp.MustCompile("yaml: cannot encode node with unknown kind 0").MatchString(err.Error()) {
+		t.Fatalf("Expected error matching %q, got %v", "yaml: cannot encode node with unknown kind 0", err)
+	}
+	err = n.Decode(&v)
+	if err == nil || !regexp.MustCompile("yaml: cannot decode node with unknown kind 0").MatchString(err.Error()) {
+		t.Fatalf("expected error matching %q, got %v", "yaml: cannot decode node with unknown kind 0", err)
```
ccoVeille, line 2951: Same remark, using a variable would be easier and safer to maintain

node_test.go
```diff
@@ -2863,57 +2890,87 @@ var nodeEncodeDecodeTests = []struct {
...
 	v.A = 1
 	data, err := yaml.Marshal(&v)
-	c.Assert(err, IsNil)
-	c.Assert(string(data), Equals, "a: 1\n")
+	if err != nil {
+		t.Fatalf("Decoded value mismatch. Got: %+v Want: nil", err)
+	}
+	if string(data) != "a: 1\n" {
+		t.Fatalf("Encoded YAML mismatch.\nGot: %q\nWant: %q", string(data), "a: 1\n")
+	}
```
ccoVeille, line 2967: Here too

node_test.go
```diff
@@ -2863,57 +2890,87 @@ var nodeEncodeDecodeTests = []struct {
...
 
 	v.B.Line = 1
 	_, err = yaml.Marshal(&v)
-	c.Assert(err, ErrorMatches, "yaml: cannot encode node with unknown kind 0")
+	if err == nil || !regexp.MustCompile("yaml: cannot encode node with unknown kind 0").MatchString(err.Error()) {
+		t.Fatalf("Expected error matching %q, got %v", "yaml: cannot encode node with unknown kind 0", err)
+	}
```
ccoVeille, line 2973: Here also

ccoVeille (approved):
I raised a few things, but LGTM 👍

Thanks for working on this

encode_test.go
carloslima: Yeah, let me fix that. ⏳

carloslima: @ccoVeille would you mind taking a look at #51 ? :)

andig:
That's why the whole thing felt so odd. I've also made good experience with `github.com/stretchr/testify` to complement `testing`. This pattern

```go
	if err != nil {
		t.Fatalf("Close() returned error: %v", err)
	}
```

then becomes

```go
    require.NoError(t, err, "Close() returned error: %v", err)
```

or simply

```go
    require.NoError(t, err, "Close() returned error")
```

ccoVeille:
@andig I'm unsure what you meant.

The way it sounds to me is "okay, but it was simpler with testify and I'm used to it."

While I agree, it doesn't sound logic to me.

Dolmen and @brackendawson testify maintainers asked yaml project to do not depends on testify because testify depends itself on (the old) yaml project. The idea is to avoid cyclic dependencies.

- https://github.com/yaml/go-yaml/issues/21#issuecomment-2939289751
- https://github.com/yaml/go-yaml/issues/21#issuecomment-2948446230
- https://github.com/yaml/go-yaml/issues/21#issuecomment-2966495487

So testify cannot be used.

Here I don't see a problem with this PR. A few files are updated, yes, it's not as simple as testify, yes, it is using vanilla testing.

But I would say it's expected because:
- it was what was asked the issue
- it is what this PR is about.

If there is a huge problem, nothing prevents to add a internal wrapper library that could check:
- for error
- if error is equal to another (using errors.Is)
- if error contains a string

And report test failure message returning the expected state.

It's easy. It can be extended if needed.

But this would be a NIH logic,
an/or at real choice of reducing the number of dependencies.

BTW Bracken suggested to use `is` which has no dependencies. It could be considered.

https://github.com/yaml/go-yaml/issues/21#issuecomment-3036429061

I feel like having a few wrappers could be considered, if vanilla become a problem.


This is the reply I'm doing about what I understood you say. But as I said, I might have misunderstood what you meant. So please apologize myself if I'm wrong and my reply is pointless.

andig:
> The idea is to avoid cyclic dependencies.

Sorry, I wasn‘t aware of that. I‘m confused by how that would be a bad thing for tests, but that‘s OT. No concerns, sorry for the noise.

ingydotnet:
This looks good enough to merge.

We can apply the optimization ideas in further PRs.

ingydotnet (approved):

encode_test.go
carloslima: I'll do in a separate MR.

encode_test.go
carloslima: I'll do in a separate MR.

PR #54, opened 2025-07-07
Move version to go.yaml.in/yaml/v4

carloslima:
We aren't going to release any non-security changes on v3, changes on main will be part of v4


ccoVeille:
What do you think about adding a section to the readme before/after the Compatibility one ?

It would explain that:

v1, v2, and v3 branches are imported from old project and are kept for maintenance/security status. 
v4 is about maintained version
main is about dev.

stefanprodan (commented):
I would postpone this change and merge it right before the v4.0.0 release. As discuss with @ingydotnet in Slack, we should also bump Go for the v4 release.

carloslima:
> What do you think about adding a section to the readme before/after the Compatibility one ?
> 
> It would explain that:
> 
> v1, v2, and v3 branches are imported from old project and are kept for maintenance/security status. v4 is about maintained version main is about dev.

@ccoVeille I added a note about v1-v3 being frozen. Does that cover what you expected?

carloslima:
> I would postpone this change and merge it right before the v4.0.0 release. As discuss with @ingydotnet in Slack, we should also bump Go for the v4 release.

I tend to see it a bit differently:

1. v3 is frozen and won't include any of the changes we've merged into main.

2. main will eventually become the v4 release.

So main is already much closer to v4.0.0-alpha than to v3. Given that, I think it makes sense to start referring to it as v4, since that's the direction it's heading. Any work related to v3 can go into the v3 branch.

Does that make sense?

ccoVeille (approved):

ingydotnet:
> I would postpone this change and merge it right before the v4.0.0 release. As discuss with @ingydotnet in Slack, we should also bump Go for the v4 release.

@stefanprodan I think this change should be merged sooner than later for 2 reasons:

* We don't want to give the false impression that changes on main or intended for v3 release
* We should put out a v4 alpha release soon, for downstreams to test

I'm fine with bumping the Go minimum to 1.24.0 as part of this and merging it now (even though 1.23 is still supported for a few more weeks). 
@carloslima do you want to add that change?
(We can also do that in a  separate PR)

carloslima:
> I'm fine with bumping the Go minimum to 1.24.0 as part of this and merging it now (even though 1.23 is still supported for a few more weeks). @carloslima do you want to add that change? 

Bumped

ccoVeille (approved):
LGTM 👍

I'm a bit worried that the minimal version would be 1.24

I face many people who cannot upgrade their stack for whatever reason.

I'm fine with supporting the v1.24 and above only, but please consider it may decrease the adoption.

Yes, v3 will be maintained for security purpose, but if no one migrates on v4 and gets bugs fixed, would this project be a success?

No one having an old Go version would be able to use the v4.

yq targets v1.24, but yq is a tool, not a module. Think about all old libraries using cobra, testify, viper.

Said otherwise, what is the best ? Adoption or have a recent Go versions?

But at same time, I know if v4 doesn't target a recent version, it would never get updated to a more recent one. It's not an easy topic.

Please note, I had this recent feedback on someone asking to downgrade Go version for a lib.

https://github.com/goforj/godump/issues/42#issuecomment-3072924544

If we are not using recent Go feature I wouldn't use a version above Go 1.18, that brought generics, or 1.21.

I'm OK with Go 1.24, but I wanted to share my thoughts about the consequences of having such recent Go versions for a low level lib like go-yaml

brackendawson:
As a Go module; you're entirely allowed to increment your major version voluntarily, and to update your minimum supported Go version to one newer than is actually needed for the language features used in the module. But both of these things cause unnecessary work for the users.

carloslima:
Personally, I prefer declaring the oldest possible version, like 1.18, and only updating it when necessary.

That said, I don't fully understand what forces some projects to stick with older versions. My naive assumption is that anyone rebuilding a project could simply use a newer version, but that clearly isn't the case for everyone.

brackendawson:
Projects which stick with older versions are most likely just using [semver](https://semver.org/) well and avoiding breaking changes. If you publish a Go module by having a valid go.mod file you should be using semver according to the [modules spec](https://go.dev/ref/mod#versions). 

Some projects want to use a different versioning system but also want to use go modules for their own vendoring, this can be done properly by opting out of publishing yourself as a module by not having a go.mod file called go.mod. Docker, who use calver, do this: https://github.com/moby/moby/blob/master/vendor.mod They can use modules for vendoring (with an extra command line arg) but their repo is not published as a go module, and is imported as `github.com/docker/docker v28.3.2+incompatible`.

It's also worth noting that incrementing the major version number of a go module practically creates a fork, albeit in the same repo. For users it's not just a matter of running a command to add the new version, they have to change every import line in their codebase to the new module name.

wwqgtxx:
As a basic library, I personally think that setting 1.24 as the minimum version is a bit too radical, especially when you decide to freeze the V3 branch and only develop new features on the V4 branch.

According to the version habits of golang, major version changes generally only occur when there are incompatible changes to the API. From the current development progress of the main branch, there is no such sign.

In addition, as many developers mentioned earlier and I mentioned in #39, many downstream projects cannot update the golang version for various reasons. The current v4 branch has no third-party dependencies, and there is no need to use the new version of the language features. Setting the go version to an older and widely used version is more conducive to the promotion of the project (such as the 1.18 that first supports generics).

ingydotnet:
Am I correct that this PR does not currently change the minimum supported Go version?

Let's do that in a separate PR.

I think we can merge this one as is.

stefanprodan (approved):
LGTM 🚀

andig:
I'm unable to get the module:

```
go get go.yaml.in/yaml/v4
go: unrecognized import path "go.yaml.in/yaml/v4": reading https://go.yaml.in/yaml/v4?go-get=1: 404 Not Found
```

carloslima:
> I'm unable to get the module:
> 
> ```
> go get go.yaml.in/yaml/v4
> go: unrecognized import path "go.yaml.in/yaml/v4": reading https://go.yaml.in/yaml/v4?go-get=1: 404 Not Found
> ```
> 

@ingydotnet we need to merge the changes to the vanity url. I have a PR open there. Can you merge?

PR #55, opened 2025-07-08
Add a CONTRIBUTING.md file

ingydotnet:
This is a stab at a Contributing rules file.

Please review and make suggestions.

Most of the content was AI generated, but I added a few things myself.

The point about `make review` will depend on another forthcoming PR.
That command will try to enforce all the rules locally.
And we can also invoke it from GitHub actions for PRS.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,124 @@
+Contributing to go-yaml
+=======================
+
+Thank you for your interest in contributing to go-yaml!
+
+This document provides guidelines and instructions for contributing to this
+project.
+
+
+## Code of Conduct
+
+By participating in this project, you agree to follow our Code of Conduct.
+
+We expect all contributors to:
+- Be respectful and inclusive
+- Use welcoming and inclusive language
+- Be collaborative and constructive
+- Focus on what is best for both the Go and YAML communities
+
+
+## How to Contribute
+
+
+### Reporting Issues
+
+Before submitting an issue, please:
+- Check if the issue already exists in our issue tracker
+- Use a clear and descriptive title
+- Provide detailed steps to reproduce the issue
+- Include relevant code samples and error messages
+- Specify your Go version and operating system
+
+
+### Pull Requests
+
+1. Fork the repository
+1. Create a new branch for your changes
+1. Make your changes following our coding conventions
+   - If you are not sure about the coding conventions, please ask
+   - Look at the existing code for examples
+1. Write clear commit messages
+1. Update tests and documentation
+1. Submit a pull request
+
+
+### Coding Conventions
+
+- Follow standard Go coding conventions
+- Use `gofmt` to format your code
+- Write descriptive comments for non-obvious code
+- Add tests for your work
+- Keep line length to 80 characters
+- Use meaningful variable and function names
+- Start doc and comment sentences on a new line
+
+
+### Commit Conventions
+
+- No merge commits
+- Commit subject line should:
+  - Start with a capital letter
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,123 @@
...
+## Code of Conduct
+
+By participating in this project, you agree to follow our Code of Conduct.
+
+We expect all contributors to:
+- Be respectful and inclusive
+- Use welcoming and inclusive language
+- Be collaborative and constructive
+- Focus on what is best for the communities
+  - Both the Go and YAML communities
```
ccoVeille: Here I'm nitpicking about style, but having sub list with one item, make it strange.

I talked about it because when I first read it, the different indentation caught my eyes.

So I would suggest this

```suggestion
- Focus on what is best for both the Go and YAML communities
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,123 @@
...
+  - Start with a capital letter
+  - Not end with a period
+  - Be no more than 50 characters
+
+
+### Testing
+
+- Run `make review` before submitting changes
+- Ensure all tests pass
+- Add new tests for new functionality
```
ccoVeille, line 69: I'm unsure how to write it clearly, but I would expect something like this:

```suggestion
- Add new tests for any functional additions or changes
```

Because it could be understood in a way people may argue later: "nah, you didn't request to provide this for a simple code change". Any change has consequence.

CONTRIBUTING.md
```diff
@@ -0,0 +1,123 @@
...
+- Update existing tests when modifying functionality
+
+
+## Development Setup
+
+1. Install Go (1.16 or later recommended)
```
ccoVeille: For development, I would expect people to use at least Go 1.22

BTW, this needs to be kept in sync with the minimal Go version present in the [go.mod](https://github.com/yaml/go-yaml/blob/main/go.mod) file

So maybe, there is something to do about it.

Old project is not defining a minimal version: https://github.com/go-yaml/yaml/blob/v3/go.mod

So maybe we will have to use 1.13, the version where go modules were added.

Here I am going out of the scope of the PR.

I feel like an issue should be created, to validate what Go version was expected for https://github.com/go-yaml/yaml, in the code, not in the readme or documentation.

So here I feel like the text should be about

```suggestion
1. Install Go (1.2X or later recommended), but code with Go 1.1X compatibility in mind.
```

stefanprodan:
```suggestion
1. Install Go (see [go.mod](github.com/yaml/go-yaml/blob/main/go.mod) for minimum required version)
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,123 @@
...
+2. Fork and clone the repository
+3. Run `go mod download` to install dependencies
```
stefanprodan: This is not needed as `go test` will download any dependencies (currently we have none).

CONTRIBUTING.md
ingydotnet: fixed

CONTRIBUTING.md
ingydotnet: fixed

CONTRIBUTING.md
ingydotnet: fixed

ccoVeille (approved):

dims (approved):
LGTM

(as long as we enforce most of these using automation! as it is hard to remember all the conventions in different repos)

PR #59, opened 2025-07-12
change error message about decoding mapping and sequence into Go `map`

garysferrao:
YAML allows sequences and mappings as a key for a mapping.
Go does not support slices or maps as a map key; and the default decoder returns an error `"yaml: invalid map key %#v"`. but i feel that this error message is misleading and may cause the user to think that it's an error in the document.

i updated the error message to be indicate this situation and ask the user to try `yaml.Node`

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -852,7 +852,7 @@ func (d *decoder) mapping(n *Node, out reflect.Value) (good bool) {
 				kkind = k.Elem().Kind()
 			}
 			if kkind == reflect.Map || kkind == reflect.Slice {
-				failf("invalid map key: %#v", k.Interface())
+				failf("cannot use '%#v' as a map key; try decoding into yaml.Node", k.Interface())
 			}
 			e := reflect.New(et).Elem()
 			if d.unmarshal(n.Content[i+1], e) || n.Content[i+1].ShortTag() == nullTag && (mapIsNew || !out.MapIndex(k).IsValid()) {
--- a/decode_test.go
+++ b/decode_test.go
@@ -1006,8 +1006,8 @@ var unmarshalErrorTests = []struct {
 	{"a: &a\n  b: *a\n", "yaml: anchor 'a' value contains itself"},
 	{"value: -", "yaml: block sequence entries are not allowed in this context"},
 	{"a: !!binary ==", "yaml: !!binary value contains invalid base64 data"},
-	{"{[.]}", `yaml: invalid map key: \[\]interface \{\}\{"\."\}`},
-	{"{{.}}", `yaml: invalid map key: map\[string]interface \{\}\{".":interface \{\}\(nil\)\}`},
+	{"{[.]}", `yaml: cannot use '\[\]interface \{\}\{"\."\}' as a map key; try decoding into yaml.Node`},
+	{"{{.}}", `yaml: cannot use 'map\[string]interface \{\}\{".":interface \{\}\(nil\)\}' as a map key; try decoding into yaml.Node`},
 	{"b: *a\na: &a {c: 1}", `yaml: unknown anchor 'a' referenced`},
 	{"%TAG !%79! tag:yaml.org,2002:\n---\nv: !%79!int '1'", "yaml: did not find expected whitespace"},
 	{"a:\n  1:\nb\n  2:", ".*could not find expected ':'"},
@@ -1525,7 +1525,7 @@ longTag:
   label: center/big
 
 inlineMap:
-  # Inlined map 
+  # Inlined map
   << : {"x": 1, "y": 2, "r": 10}
   label: center/big
 


decode.go
```diff
@@ -1016,3 +1020,101 @@ func (d *decoder) merge(parent *Node, merge *Node, out reflect.Value) {
...
+//   - Otherwise, it returns v unchanged (for primitives, structs, etc.).
+func makeComparableKey(v reflect.Value) (reflect.Value, error) {
+	switch v.Kind() {
+	case reflect.Interface:
+		if v.IsNil() {
+			// Zero value for interface{}
+			return reflect.Zero(reflect.TypeOf((*interface{})(nil)).Elem()), nil
+		} else {
+			return makeComparableKey(v.Elem())
+		}
```
ccoVeille:
```suggestion
		if v.IsNil() {
			// Zero value for interface{}
			return reflect.Zero(reflect.TypeOf((*interface{})(nil)).Elem()), nil
		}

		return makeComparableKey(v.Elem())
```

decode.go
```diff
@@ -1016,3 +1020,101 @@ func (d *decoder) merge(parent *Node, merge *Node, out reflect.Value) {
...
+//
+//   - If v.Kind() is Slice, it returns a fixed-length array [N]T, where
+//     N = v.Len() and T is the slice’s element type.
+//
+//   - If v.Kind() is Map, it returns a fixed-length array [M]Entry, where
+//     M = number of entries
+//     type Entry struct { Key K; Value V } with K and V the original map’s key and value types.
+//
+//   - Otherwise, it returns v unchanged (for primitives, structs, etc.).
+func makeComparableKey(v reflect.Value) (reflect.Value, error) {
```
ccoVeille: I read the code multiple times, and I feel like err is always nil. The code is recursive, but nothing return an error.

I might be wrong, but removing error from the signature if we don't need it would make the code way simpler

garysferrao: @ccoVeille thank you for your review.
you are correct, the only concrete error it returned was `nil`.
 i added the condition when it should return error. this should not happen during normal operation of the parser. but since Go language allows it, and we have to check it dynamically, i added the error for completeness. please see 593d859083a2f8af99756d6e1b9558791aa177ed.

i have unit tests for the unexported function `makeComparableKey`. but all the test files can only test exported functions (as they are in package `yaml_test`). please let me know whether / how i should include the unit test.

decode.go
garysferrao: thank you for your review. i simplified the `if…else` block that uses `return` statements in this commit 6c726ecc642c843a654168ab39d5fef930606e75.

decode.go
ccoVeille: It's all good now. I don't think there is a need of testing it

decode.go
```diff
@@ -1016,3 +1020,108 @@ func (d *decoder) merge(parent *Node, merge *Node, out reflect.Value) {
...
+			arrVal.Index(i).Set(entries.Index(i))
+		}
+		return arrVal, nil
+
+	default:
+		// base case (string, int, etc.)
+		// this check needs to come after converting slice and map to a comparable array.
+		if v.Type().Comparable() {
+			return v, nil
+		}
```
ccoVeille: I'm unsure this default statement is *only* about string, int ...

I feel like the code is OK, but the comment is not.

For me, the comment should be in the if comparable

garysferrao: @ccoVeille 
i see. i usually put comments above the target statement. but it seems it's a comment about the `default` case.

how about this?
```suggestion
	default:
		if v.Type().Comparable() {	
			// slice and map are not comparable. so this check needs to come after processing them into comparable arrays.
			// returns comparable Go types (string, int, float64, time.Time, bool, etc.) that are parsed from YAML scalars.
			// also includes other Go comparable types that wouldn't be parsed.
			return v, nil
		}
```

ccoVeille: It would be better, thanks

garysferrao: thank you so much. 😄
in addition, i went over my other messages/comments as well, and found they could be improved:
* "Node" is now more specifically called "Mapping" in the comment.
* changed helper function name to `makeComparableType`, as it doesn't necessarily relate to map keys.
* error message `"cannot use %s as a map key"` is used only when actually setting a map key. the rest are shortened to `"type %s is not comparable"`.

the changes are in commit aa26e9e1f1908ae880c91c7201531e2bece4f854.

ccoVeille (approved):

WGH-:
The resulting type is extremely awkward to use. Is there a use case for this? Have you ever seen such YAML documents, even? :)

garysferrao:
@WGH- 
> Have you ever seen such YAML documents

https://yaml.org/spec/1.2.2/#example-mapping-between-sequences
who am _i_ to judge if Go haven't caught up 🙃

<details>
<summary>Java's <code>snakeyaml</code> can do it</summary>

```java
import org.yaml.snakeyaml.Yaml;
import java.io.InputStream;
import java.util.Map;

public class Example {
    public static void main(String[] args) {
        Yaml yaml = new Yaml();
        try (InputStream inputStream = Example.class
                .getClassLoader()
                .getResourceAsStream("test.yaml")) {
            Map<Object, Object> data = yaml.load(inputStream);
            System.out.println(data);
        } catch (Exception e) {
            e.printStackTrace();
        }
    }
}
```

```
￥ java -cp .:snakeyaml-2.4.jar Example
{[Detroit Tigers, Chicago cubs]=[Mon Jul 23 00:00:00 UTC 2001]}

￥ java -cp .:snakeyaml-2.4.jar Example
{{null=false}={42=3.1416}}
```

</details>

‌

my main issue is that the previous error message is misleading: seems like a input problem, when in fact the key is valid.
https://go.dev/play/p/wylC-DerUM6
```
2009/11/10 23:00:00 error: yaml: invalid map key: []interface {}{"Detroit Tigers", "Chicago cubs"}
```
i don't mind the second option
> or change the error message to something like `"yaml: cannot use %#v as a Go map key"`

WGH-:
I mean being defined in the spec in one thing, and being actually useful and encountered in the wild is another.

It is isn't that go-yaml doesn't support it at all: it works fine when unmarshalling into `yaml.Node`, after all.

> my main issue is that the previous error message is misleading: seems like a input problem, when in fact the key is valid.
>
> i don't mind the second option
>> or change the error message to something like "yaml: cannot use %#v as a Go map key"

_Personally_, I think this is the better option. But I'm not a maintainer, so I'd say let's wait for their opinion.

WGH-:
And I think this is a pretty cool exercise in Go type system, and the implementation is quite clever and readable... but I don't see much use of it.

garysferrao:
@WGH- 

> It is isn't that go-yaml doesn't support it at all: it works fine when unmarshalling into `yaml.Node`, after all.

yes. i would prefer actually to return `yaml.Node` by default in such case, but the thing is the default option seems to be a map: https://go.dev/play/p/KR2L-tmE8f0

‌

> I mean being defined in the spec in one thing, and being actually useful and encountered in the wild is another.

i appreciate your comment, but i have to disagree, because the whole point of a library is to follow the specification. there was a case in the Go's XML library which seemed "not-to-spec" when compared with another library ([ref](https://github.com/golang/go/issues/12611#issuecomment-2830565949)), but that was a case where the specification itself allowed implementations to make a choice.

i would still argue for supporting recursive sequences into arrays for Go-map keys. recursively converting slices into arrays seems natural to me, and it already automatically works with `Marshal()`: https://go.dev/play/p/vDvrre0cZWp

although, i am not sure about converting mapping into an arbitrary `struct {Key, Value}`. but i would at least ask to change the error message to tell the user to try unmarshal into a `yaml.Node`, or it's a limitation of Go instead of an error in the document.  i'm open to hearing feedback from the maintainers.

ingydotnet:
> the whole point of a library is to follow the specification

I would say that is true for a "new" library.
go-yaml is a very important existing library.
The number one goal is to maintain it well.
That said I also want to bring it closer to the spec (without breaking existing usage).
This is doable.

Sadly very few YAML implementations were made directly from the spec.
Most are descendants of PyYAML including this one. (libyaml descended from PyYAML).

ingydotnet:
YAML itself supports any node as a mapping key.

This is not always usable within a particular language's native data types.

I recently created a go-yaml CLI: https://github.com/yaml/go-yaml-cli#readme

Let's try it out with container keys. Here's another valid way to use sequence keys in YAML:
```
$ cat a.yaml
[1, 2]: 3
[3, 4]: 7
```

Let's decode it in go-yaml:

```
kind: Document
content:
  - kind: Mapping
    content:
      - kind: Sequence
        style: Flow
        content:
          - kind: Scalar
            tag: '!!int'
            text: "1"
          - kind: Scalar
            tag: '!!int'
            text: "2"
      - kind: Scalar
        tag: '!!int'
        text: "3"
      - kind: Sequence
        style: Flow
        content:
          - kind: Scalar
            tag: '!!int'
            text: "3"
          - kind: Scalar
            tag: '!!int'
            text: "4"
      - kind: Scalar
        tag: '!!int'
        text: "7"
```

go-yaml handles this fine when loading into Node structs.

It even roundtrips it:

```
$ <a.yaml ./go-yaml -Y
? [1, 2]
: 3
? [3, 4]
: 7
```

Also:
```
$ cat b.yaml 
{1: 2}: 3
{3: 4}: 7
$ <b.yaml ./go-yaml -Y
? {1: 2}
: 3
? {3: 4}
: 7
```

ingydotnet:
What I'd suggest is making this PR only about clarifying the error messages.
I think that is a great idea.

People can use the Decode interface on YAML with collection keys just fine already.

I'm not yet sure we need to add the helper code for fixed length arrays, but feel free to submit that as a separate PR.

garysferrao:
@ingydotnet 
thank you for your comments

> I recently created a go-yaml CLI: https://github.com/yaml/go-yaml-cli#readme

nice!~ 

> This is not always usable within a particular language's native data types.

yes. sorry, i was being a bit sarcastic regarding Go. they do not support types that can dynamically change its hash as a map key for good reason.

> The number one goal is to maintain it well.

ah yes. and if breaking changes, would need to go into a "v5" (damn, it escalated since i saw it in June 🫠)

> (libyaml descended from PyYAML)

that's interesting!~ i tried to find the first commit, but it seems empty 😄
https://github.com/yaml/libyaml/commit/ceebf81b0b02e0bbe3e04979357022749349d6d9


> YAML itself supports any node as a mapping key.

so many questions to ask about designing specifications. but i'll keep the PR clean.

‌
---
sure, i will make 2 PRs then:
* a simple error message update
* support recursive arrays as when `Unmarshal` into Go map keys

thank you @WGH- and @ccoVeille as well for your reviews and comments so far on this PR. 🙇

ingydotnet:
> * a simple error message update

Will you repurpose this one for the improved error messages?

Look forward to it.

garysferrao:
@ingydotnet 

> Will you repurpose this one for the improved error messages?

yes.
there were some merge conflicts and the code was reorganised since i branched off main. so i pulled the new main and forced pushed the new change here 🙇

(+ @ccoVeille @WGH-)
please let me know if the error message is alright 🙏

ingydotnet:
Changed commit message to current rules.

PR #63, opened 2025-08-06
Better reporting of results on yts tests

carloslima:
1. output a clear count of PASS and FAIL
2. exit with error when FAIL count differs from known
3. when a known failing test passes, display the test name and error out


ingydotnet (approved):

stefanprodan (approved):
LGTM

PR #64, opened 2025-08-06
Yts tests improvements

carloslima:
1. DRY up and improve readability on yts tests
2. Fix flawed logic on MarshalTests by splitting it into two separate tests


ingydotnet:
What's the current status of this one @carloslima ?

ingydotnet:
What's the current status of this one @carloslima ? :)

yts/test_suite_test.go
```diff
@@ -89,30 +89,30 @@ func getEvents(in []byte) (string, error) {
 	return yaml.ParserGetEvents(in)
 }
 
+func mustRead(t *testing.T, path, name string) []byte {
+	data, err := os.ReadFile(filepath.Join(path, name))
```
SVilgelm, line 94: Please add `t.Helper()`
```suggestion
	t.Helper()
	
	data, err := os.ReadFile(filepath.Join(path, name))
```

yts/test_suite_test.go
```diff
@@ -90,30 +90,30 @@ func getEvents(in []byte) (string, error) {
...
 	// Read test description
-	descPath := filepath.Join(testPath, "===")
-	desc, err := os.ReadFile(descPath)
-	var testDescription string
-	if err == nil {
-		testDescription = string(desc)
-	} else {
-		testDescription = "No description available."
-	}
+	testDescription := mustRead(t, testPath, "===")
```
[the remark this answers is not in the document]
SVilgelm: I completely agree here with the copilot, we should not fail the test if the description is missing.
Basically you added around 200 tests as know failing tests just because the description is missing.
I don't think it is a god improvement.

yts/test_suite_test.go
```diff
@@ -90,30 +90,30 @@ func getEvents(in []byte) (string, error) {
...
-	if err != nil {
-		t.Fatalf("Test: %s\nDescription: %s\nError: Failed to read in.yaml: %v", testPath, testDescription, err)
-	}
-
-	errorPath := filepath.Join(testPath, "error")
-	_, err = os.Stat(errorPath)
-	expectError := err == nil
+	inYAML := mustRead(t, testPath, "in.yaml")
+	expectError := fileExists(testPath, "error")
+	expectedEvents := normalizeLineEndings(string(mustRead(t, testPath, "test.event")))
```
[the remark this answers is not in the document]
SVilgelm: also agree with the copilot's review

yts/test_suite_test.go
```diff
@@ -123,110 +123,135 @@ func runTest(t *testing.T, testPath string) {
...
-		expectedOutYAML, err := os.ReadFile(outYAMLPath)
-		if err != nil {
-			t.Errorf("Test: %s\nDescription: %s\nError: Failed to read out.yaml: %v", testPath, testDescription, err)
+	})
+
+	t.Run("RoundtripTest", func(t *testing.T) {
+		shouldSkipTest(t)
+		var unmarshaledValue interface{}
+
+		if expectError {
```
SVilgelm, line 201: Are you sure that the check for expectError should be done before any other steps, like nothing is done yet?

yts/test_suite_test.go
```diff
@@ -123,110 +123,135 @@ func runTest(t *testing.T, testPath string) {
...
+				testPath, testDescription, marshalErr)
 		}
-		var reUnmarshaledValue any
-		err = yaml.Unmarshal(marshaledYAML, &reUnmarshaledValue)
-		if err != nil {
-			t.Errorf("Test: %s\nDescription: %s\nError: Failed to re-unmarshal marshaled YAML: %v", testPath, testDescription, err)
-		} else if !reflect.DeepEqual(reUnmarshaledValue, expectedUnmarshaledValue) {
-			t.Errorf("Test: %s\nDescription: %s\nError: Marshal output mismatch\nExpected: %+v\nGot     : %+v", testPath, testDescription, expectedUnmarshaledValue, reUnmarshaledValue)
+
+		var roundtrippedValue interface{}
```
SVilgelm: lease use `any`

```suggestion
		var roundtrippedValue any
```

yts/test_suite_test.go
```diff
@@ -123,110 +123,135 @@ func runTest(t *testing.T, testPath string) {
...
 
-		currentUnmarshalErr := yaml.Unmarshal(inYAML, &currentUnmarshaledValue)
-
-		if !(currentUnmarshalErr == nil || expectError) {
-			return
-		}
-		inJSONPath := filepath.Join(testPath, "in.json")
-		if _, err := os.Stat(inJSONPath); err != nil {
+		// Nothing to test if unmarshall is expected to error
+		if expectError || !fileExists(testPath, "in.json") {
```
SVilgelm, line 235: again, the check for expectError happens before `yaml.Unmarshal` has been called.

yts/test_suite_test.go
```diff
@@ -123,110 +123,135 @@ func runTest(t *testing.T, testPath string) {
...
 			return
 		}
-		inJSON, err := os.ReadFile(inJSONPath)
-		if err != nil {
-			t.Errorf("Test: %s\nDescription: %s\nError: Failed to read in.json: %v", testPath, testDescription, err)
-			return
+		inJSON := mustRead(t, testPath, "in.json")
+
+		var unmarshaledValue interface{}
```
SVilgelm:
```suggestion
		var unmarshaledValue any
```
`any` is more readable and way shorter :)

yts/test_suite_test.go
```diff
@@ -123,110 +123,135 @@ func runTest(t *testing.T, testPath string) {
...
+			t.Errorf("Test: %s\nDescription: %s\nError: Failed to unmarshal in.yaml: %v", testPath, testDescription, err)
 		}
-		var jsonValue any
-		jsonErr := json.Unmarshal(inJSON, &jsonValue)
-		if jsonErr != nil {
-			t.Errorf("Test: %s\nDescription: %s\nError: Failed to unmarshal in.json: %v", testPath, testDescription, jsonErr)
-		} else if !reflect.DeepEqual(currentUnmarshaledValue, jsonValue) {
-			t.Errorf("Test: %s\nDescription: %s\nError: YAML unmarshal vs JSON unmarshal mismatch\nExpected: %+v\nGot     : %+v", testPath, testDescription, jsonValue, currentUnmarshaledValue)
+
+		var jsonValue interface{}
```
SVilgelm:
```suggestion
		var jsonValue any
```

yts/test_suite_test.go
```diff
@@ -123,110 +123,135 @@ func runTest(t *testing.T, testPath string) {
...
 
-	// Only proceed with marshal and JSON tests if unmarshal was successful and no expected error
-
-	t.Run("MarshalTest", func(t *testing.T) {
+	t.Run("MarshalCanonicalTest", func(t *testing.T) {
 		shouldSkipTest(t)
-		var currentUnmarshaledValue any
-
-		currentUnmarshalErr := yaml.Unmarshal(inYAML, &currentUnmarshaledValue)
+		var unmarshaledValue interface{}
```
SVilgelm: please use `any`, it is shorter and more readable

```suggestion
		var unmarshaledValue any
```

yts/test_suite_test.go
```diff
@@ -123,110 +123,135 @@ func runTest(t *testing.T, testPath string) {
...
 
-		if !(currentUnmarshalErr == nil || expectError) {
+		if expectError || !fileExists(testPath, "out.yaml") {
```
SVilgelm, line 170: check for `expectError` should be done after `yaml.Unmarshal`

yts/test_suite_test.go
```diff
@@ -123,110 +123,135 @@ func runTest(t *testing.T, testPath string) {
...
+				testPath, testDescription, string(expectedOutYAML), string(marshaledYAML))
 		}
-		expectedOutYAML, err := os.ReadFile(outYAMLPath)
-		if err != nil {
-			t.Errorf("Test: %s\nDescription: %s\nError: Failed to read out.yaml: %v", testPath, testDescription, err)
+	})
+
+	t.Run("RoundtripTest", func(t *testing.T) {
+		shouldSkipTest(t)
+		var unmarshaledValue interface{}
```
SVilgelm:
```suggestion
		var unmarshaledValue any
```

yts/test_suite_test.go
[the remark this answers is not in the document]
carloslima: Can you help me understand what test cases don't have a description? I don't see any and based on the data structure from the test suite I don't expect it to happen. That's why I made it required and a failure if the file is not there.

I can change it, but I would like to understand the 200 tests you are talking about. I see zero test directories without a description file.

yts/test_suite_test.go
carloslima: I disagree. Maybe I got the wrong understanding (the test files didn't come with an instruction manual) but when an error is expected, the only test I think we should run is the `UnmarshallTest`.

yts/test_suite_test.go
carloslima: I believe so. Reasoning in the comment above.

yts/test_suite_test.go
carloslima: This PR has been waiting for a long time, it's from before we bumped go versions and were able to use any. Fixing all instances.

yts/test_suite_test.go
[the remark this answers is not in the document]
carloslima: Based on the existing tests, every test has a `test.event` file and reading the file has no harm if the test is skipped.

Either way, I have moved it inside the EventComparisonTest

yts/test_suite_test.go
carloslima: For the most part, those changes are fixes to what I think I got wrong when I originally wrote the tests.

ccoVeille:
What is the status of this PR vs #159 ?

@SVilgelm asked if we could merge #159 first witg this comment

- https://github.com/yaml/go-yaml/pull/159#issuecomment-3478915425

I don't think there is a clear status here.

And I feel these 2 PRs are so close that I cannot spot the differences.

I would like one to be merged, so we can move on.

Here, it's about improving the test suite. It's not vital.

But I would prefer if we don't keep #159 and #64 open in parallel longer.

Carlos PR is 3 months old.

ingydotnet:
@carloslima can you look at the current code and test system and see if this still makes sense?

ingydotnet:
@carloslima I'm closing this simply because it is too old. Feel free to open up a new one for these things.

PR #66, opened 2025-08-15
Use helper functions to DRY up tests

carloslima:


carloslima:
@ccoVeille @stefanprodan Here are the changes I owed you from #52, can you take a look?

Sorry for the delay :sweat_smile:

go.mod
```diff
@@ -1,3 +1,3 @@
 module go.yaml.in/yaml/v4
 
-go 1.16
+go 1.20
```
carloslima: This is needed because of `comparable` in the generic type:

```
./decode_test.go:1256:22: interface{} to satisfy comparable requires go1.20 or later (-lang was set to go1.18; check go.mod)
./encode_test.go:675:22: error to satisfy comparable requires go1.20 or later (-lang was set to go1.18; check go.mod)
```

testutil/assert.go
ccoVeille, line 1: I would use a package in internal

We don't want people to use it directly, no?

go.mod
ccoVeille: Bumping the minimal version for adding tests doesn't sound fair

node_test.go (decode block)
ccoVeille, line 1: This file sounds odd

testutil/assert.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// Comparable types (numbers, strings, pointers to the same object, etc.).
+func AssertEqual[T comparable](tb miniTB, got, want T, msgAndArgs ...any) {
+	tb.Helper()
+	if got != want {
+		suffix := formatSuffix(msgAndArgs...)
+		tb.Fatalf("got %v; want %v%s", got, want, suffix)
+	}
+}
```
ccoVeille: Can't this work with any?


```suggestion
// Comparable types (numbers, strings, pointers to the same object, etc.).
func AssertEqual(tb miniTB, got, want any, msgAndArgs ...any) {
	tb.Helper()
	if got != want {
		suffix := formatSuffix(msgAndArgs...)
		tb.Fatalf("got %v; want %v%s", got, want, suffix)
	}
}
```

testutil/assert.go
```diff
@@ -0,0 +1,131 @@
...
+// Anything else (slices, maps, structs with slices...).
+func AssertDeepEqual(tb miniTB, got, want any, msgAndArgs ...any) {
+	tb.Helper()
+	if !reflect.DeepEqual(got, want) {
+		suffix := formatSuffix(msgAndArgs...)
+		tb.Fatalf("got %+v; want %+v%s", got, want, suffix)
+	}
+}
+
+func AssertErrorMatches(tb miniTB, err error, pattern string, msgAndArgs ...any) {
```
ccoVeille: Here and everywhere, I would use two methods for each


```suggestion
func AssertErrorMatches(tb miniTB, err error, pattern string) {
```

And


```suggestion
func AssertErrorMatchesf(tb miniTB, err error, pattern string, msg string, args ...any) {
```

Avoiding the oddities with playing with type assertion

testutil/assert.go
ccoVeille: If not, then adding as many asserter needed would be better than bumping Go minimal version

Here I'm suggesting adding

AssertEqualInt
AssertEqualFloat64
...

decode_test.go
```diff
@@ -995,29 +977,21 @@ func (errReader) Read([]byte) (int, error) {
...
 	value := map[string]interface{}{}
 	err := yaml.Unmarshal([]byte("notanum: .NaN"), &value)
-	if err != nil {
-		t.Fatalf("Unmarshal() returned error: %v", err)
-	}
-	if !math.IsNaN(value["notanum"].(float64)) {
-		t.Fatalf("Unmarshal() returned %v, want NaN", value["notanum"])
-	}
+	testutil.AssertNoError(t, err)
+	testutil.AssertEqual(t, math.IsNaN(value["notanum"].(float64)), true)
```
ccoVeille: Here having an AssertTrue would be better

decode_test.go
```diff
@@ -1224,54 +1162,33 @@ func TestUnmarshalerValueField(t *testing.T) {
...
-	}
+	testutil.AssertNoError(t, err)
 	value, ok := obj.value.(map[string]interface{})
-	if !ok {
-		t.Fatalf("obj.value is not a map[string]interface{}. Value: %#v", obj.value)
-	}
-	if !reflect.DeepEqual(value["_"], unmarshalerTests[0].value) {
-		t.Fatalf("Unmarshal() returned %#v but expected %#v", value["_"], unmarshalerTests[0].value)
-	}
+	testutil.AssertEqual(t, ok, true, "value: %#v", obj.value)
```
ccoVeille: Same here, using AssertTrue would be better

decode_test.go
```diff
@@ -1296,35 +1213,14 @@ func TestUnmarshalerTypeError(t *testing.T) {
...
-	if !reflect.DeepEqual(v.M["ghi"].value, 3) {
-		t.Fatalf("v.M[\"ghi\"].value is %v, want %v", v.M["ghi"].value, 3)
-	}
+	testutil.AssertErrorMatches(t, err, expectedError)
+	testutil.AssertNotNil(t, v.M["abc"])
+	testutil.AssertIsNil(t, v.M["def"])
+	testutil.AssertNotNil(t, v.M["ghi"])
+	testutil.AssertIsNil(t, v.M["jkl"])
+
+	testutil.AssertEqual(t, v.M["abc"].value, 1)
```
ccoVeille: You could have used AssertDeepEqual as in the code you replace.

But then I wonder how the code would suffer from having AssertEqual using reflect.DeepEqual and having no AssertDeepEqual.

It would be slower, and more restrictive. But I feel like it could be OK, and would avoid to bump Go minimal version

testutil/assert.go
carloslima: 👍
Will update when I get back to my desk

testutil/assert.go
carloslima: I took the pattern from the popular testify module https://github.com/stretchr/testify/blob/master/assert/assertions.go#L495

I tend to prefer their approach with the optional args at the end

ccoVeille: I'm a testify maintainer and we agree it was a bad idea 😁

Testifylint was created and it reports it.

decode_test.go
carloslima: The original code used Equals:

```
c.Assert(v.M["abc"].value, Equals, 1)
```

I somehow moved it to DeepEqual when removing gocheck and now I moved it back to what it was :-)

testutil/assert.go
ccoVeille: You will find explanation here

https://github.com/Antonboom/testifylint?tab=readme-ov-file#3

go.mod
carloslima: I moved it back to 1.18 which is what I remember we decided to go with for v4

testutil/assert.go
carloslima: done, let me know if you thought of something different

testutil/assert.go
carloslima: Nope, it can't. I aliased it to AssertDeepEqual until we bump the minimal version up.

carloslima: Sorry, my mistake, it can. Fixing :hourglass:

carloslima: Fixed

testutil/assert.go
ccoVeille: But I could live with what you implemented

carloslima: Thanks for the pointer. Let me implement your suggestion :hourglass:

go.mod
```diff
@@ -1,3 +1,3 @@
 module go.yaml.in/yaml/v4
 
-go 1.16
+go 1.18
```
ccoVeille: I feel like you rolled back manually

internal/testutil/assert.go
```diff
@@ -0,0 +1,131 @@
+package testutil
```
ccoVeille: I would rename the package to assert abd move the file in internal/testutil/assert


```suggestion
package assert
```

Then remove Assert prefix from all methods

This way it would look like this in _test.go files

```diff
-     testutil.AssertNoError(t, err)
+     assert.NoError(t, err)
```

decode_test.go
carloslima: done

internal/testutil/assert.go
carloslima: the file is already named `internal/testutil/assert.go`

should I just change the package name then or did I misunderstand your suggestion?

go.mod
carloslima: I moved it down to 1.18 because that's what I remember we had agreed to go with for v4 and I didn't want to unnecessarily rewrite all instances of `any` to `interface{}` if we are going with 1.18 at the end :-)

Did I misremember?

internal/testutil/assert.go
ccoVeille: I'm sorry my suggestion wasn't complete

So using something like this

`internal/testutil/assert/assert.go`

with `package assert`

And remove the Assert prefix from all methods

go.mod
ccoVeille: The discussion was a rollercoaster

But here again, if you only need Go 1.18 for being able to use `any` in a test helper, it's not sufficient.

Using `interface{}` and keeping Go 1.16 should be considered

internal/testutil/assert.go
carloslima: :+1:  Got it, thanks. Let me do that

go.mod
carloslima: I can definitely do that, wouldn't be a problem. I was just avoiding if the version we are using on v4 is going to be 1.18

Let me move it back to 1.16 and we can talk about it somewhere else then

internal/testutil/assert.go
carloslima: done, please take a look :)

go.mod
carloslima: done

encode_test.go
```diff
@@ -713,26 +672,18 @@ func (ft *failingMarshaler) MarshalYAML() (interface{}, error) {
 
 func TestMarshalerError(t *testing.T) {
 	_, err := yaml.Marshal(&failingMarshaler{})
-	if err != failingErr {
-		t.Fatalf("Marshal() returned %v, want %v", err, failingErr)
-	}
+	assert.Equal(t, err, failingErr)
```
ccoVeille: Error should be tested with ErrorIs and not Equal.

Error could be wrapped, and here it would fail.

So adding a `ErrorIs` asserter could be a good option.

We can leave it like this, but I prefer to raise the point.

And, yes, I know you are only refactoring, but I would prefer if this could be fixed

decode_test.go
```diff
@@ -1183,39 +1137,23 @@ func TestUnmarshalerPointerField(t *testing.T) {
...
-			}
+			assert.IsNil(t, obj.Field)
 		} else {
-			if obj.Field == nil {
-				t.Fatalf("Unmarshal() returned nil, want non-nil. Value: %#v", item.value)
-			}
-			if !reflect.DeepEqual(obj.Field.value, item.value) {
-				t.Fatalf("Unmarshal() returned\n%#v\nbut expected\n%#v", obj.Field.value, item.value)
-			}
+			assert.NotNil(t, obj.Field, "Pointer not initialized (%#v)", item.value)
```
ccoVeille: You didn't work on having NotNilf

https://github.com/yaml/go-yaml/pull/66#discussion_r2279739002

internal/testutil/assert/assert.go
```diff
@@ -0,0 +1,147 @@
...
+func Equal(tb miniTB, got, want interface{}, msgAndArgs ...interface{}) {
+	tb.Helper()
+	if got != want {
+		suffix := formatSuffix(msgAndArgs...)
+		tb.Fatalf("got %v; want %v%s", got, want, suffix)
+	}
+}
+
+// interface{}thing else (slices, maps, structs with slices...).
+func DeepEqual(tb miniTB, got, want interface{}, msgAndArgs ...interface{}) {
```
ccoVeille: Usually test asserter are like this


```suggestion
func Equal(tb miniTB, want, got interface{}, msgAndArgs ...interface{}) {
	tb.Helper()
	if got != want {
		suffix := formatSuffix(msgAndArgs...)
		tb.Fatalf("got %v; want %v%s", got, want, suffix)
	}
}

// interface{}thing else (slices, maps, structs with slices...).
func DeepEqual(tb miniTB, want, got interface{}, msgAndArgs ...interface{}) {
```

testutil/assert.go
ccoVeille: Not resolved

carloslima: I was working on it. It's done now, can you check?

decode_test.go
carloslima: It's done

internal/testutil/assert/assert.go
carloslima: Outch. It makes more sense in my head the other way around, but I'll adjust it to match testify's order. :hourglass:

encode_test.go
```diff
@@ -521,17 +521,12 @@ var marshalTests = []struct {
 }
 
 func TestMarshal(t *testing.T) {
-	defer os.Setenv("TZ", os.Getenv("TZ"))
-	os.Setenv("TZ", "UTC")
+	t.Setenv("TZ", "UTC")
```
ccoVeille: t.SetEnv is only available since Go 1.17

So here this change might break things. But only the test. I would keep the old code

encode_test.go
ccoVeille: What do you think about this?

carloslima: I will look into making the assertion after I get the argument order resolved :hourglass_flowing_sand:

carloslima: done, please take a look

encode_test.go
carloslima: I reverted the change but I have a bad feeling about supporting versions too low. Do you know where I can learn about the issues preventing some people from using a supported version of go?

ingydotnet:
I just read through #39 which goes over minimum version stuff in detail.

Then I looked at the minimum version for various branches:

* v1 - no `go.mod`
* v2 - 1.15 - https://github.com/yaml/go-yaml/blob/v2/go.mod
* v3 - 1.16 - https://github.com/yaml/go-yaml/blob/v3/go.mod
* v4 - 1.22 - https://github.com/yaml/go-yaml/blob/v4/go.mod
* main - 1.16 - https://github.com/yaml/go-yaml/blob/main/go.mod

Currently main is acting as the development/integration branch for code intended to land in v4. 
main and v4 should be in sync wrt minimum version.

@carloslima what is the minimum version you need for the features you'd like to use here?
I think it's 1.20.

**What do people think about using 1.20 for main and v4?**

We could also go with 1.24 (see https://endoflife.date/go) but 1.20 seems safer and gets the job done for now.

^^ @wwqgtxx @andig @liggitt @stefanprodan @dims feel free to weigh in.

----
PS I'm not sure we need both a main and v4 branch, but let's discuss that in [#go-yaml on CNCF Slack](https://cloud-native.slack.com/archives/C08PPAT8PS7).

wwqgtxx:
In fact, there is a suggestion in my previous discussion to place test as a subpackage in a separate folder so that it can have its own go.mod and can use any external packages and modern golang versions you want without affecting the main module.

ccoVeille:
> In fact, there is a suggestion in my previous discussion to place test as a subpackage in a separate folder so that it can have its own go.mod and:

> can use any external packages

Yes, this could help to import some method of testify without recoding it, and without having a need to bloat main go.mod with dependencies

> modern golang versions you want without affecting the main module.

I'm pretty sure you cannot import a Go 1.20 module in Go 1.16 project.

You can do the opposite, but then you won't be able to test code with version below 1.20

ccoVeille:
But testify requires Go 1.17

I would say that for now the PR is fine, and could stay with current code and PR.

Note: I'm OK with bumping minimal Go version to something higher, but as I said, I don't think it should be done for a small refactoring about cleaning up test files

wwqgtxx:
In my personal use, maintaining compatibility with 1.20 is indeed sufficient, but I'm not sure if others require a lower version.

Based on the content of this PR, combined with the discussion in #54, setting the minimum version to 1.18 may be less controversial.

ingydotnet:
> In my personal use, maintaining compatibility with 1.20 is indeed sufficient, but I'm not sure if others require a lower version.
> 
> Based on the content of this PR, combined with the discussion in #54, setting the minimum version to 1.18 may be less controversial.

I'm in favor of that. 
@carloslima can you update go.mod to use 1.18 and apply the changes to this PR that this would allow?

ccoVeille:
Here moving to Go 1.18 would only bring the ability to use `any` instead of `interface{}`

This change would bring clarity to code and can be done easily with a PR that would replace the `interface{}` in all code base.

We already stated this project shouldn't depend on testify.

So, I would suggest these:
- acknowledging the assertions helpers in this PR are enough for what we need.
- keep Go 1.16 for this PR
- merge the PR as-is
- open a PR to bump to Go 1.18 and fix all `interface{}`

ccoVeille:
I started working on bumping minimal version to Go 1.18, and it's not as simple as I thought

- https://github.com/yaml/go-yaml/pull/69

So I would say, let's move discussion about minimal version there, and keep this PR focused on bringing the feature of having test helper

ccoVeille:
Currently, `main` branch cannot work with version under Go < 1.20

- https://github.com/yaml/go-yaml/issues/71

Let's continue discussion there

carloslima:
Only `any` and `t.Setenv` would be changed with a bump to 1.18. To use `comparable` generic type we'd need to jump to 1.20. So, all in all, I think this PR can go as is and we discuss the versioning somewhere else.

I've rebased this on top of latest main and squashed the changes

carloslima:
@wwqgtxx since you do have lower version requirements, would you mind sharing a pointer so I can learn about the challenges preventing you from running a higher version?

internal/testutil/assert/assert.go
```diff
@@ -0,0 +1,196 @@
...
+
+func Equalf(tb miniTB, want, got interface{}, msgFormat string, args ...interface{}) {
+	tb.Helper()
+	if got != want {
+		suffix := formatSuffix(msgFormat, args...)
+		tb.Fatalf("got %v; want %v%s", got, want, suffix)
+	}
+}
+
+// interface{}thing else (slices, maps, structs with slices...).
```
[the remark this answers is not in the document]
carloslima: fixed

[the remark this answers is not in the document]
ccoVeille: If you ever need it again, instead of doing a blind search and replace, you can use the following command

```console
$ gofmt -w -r 'interface{} -> any' path/
```

@carloslima

ccoVeille (approved):

internal/testutil/assert/assert.go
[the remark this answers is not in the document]
carloslima: TIL `gofmt -r`.  Thanks! :-)

carloslima:
@ingydotnet @stefanprodan can you take a look? :-)

ingydotnet (approved):

ingydotnet (approved):

ccoVeille (approved):

PR #72, opened 2025-08-18
Fix repository URL in package comment

ccoVeille:
- **chore: fix repository URL in package comment**

### DIFF
--- a/yaml.go
+++ b/yaml.go
@@ -17,7 +17,7 @@
 //
 // Source code and other details for the project are available at GitHub:
 //
-//	https://github.com/go-yaml/yaml
+//	https://github.com/yaml/go-yaml
 package yaml
 
 import (


ingydotnet:
This looks like some standard Go stuff, but I'm not familiar.

Can you point to docs about what you are doing and leave a bit of explanation about what/why in the PR description?

ccoVeille:
> Packages with long package documentation may choose to dedicate one file called doc.go that has the package documentation, a package declaration, and nothing else, but this is not required.
Source: https://google.github.io/styleguide/go/best-practices.html


> Manage lengthy documentation using doc.go

If your package-level introductory documentation is very long, you can create a doc.go file. It only contains comments that are written following Godoc conventions.
Source: https://nirdoshgautam.medium.com/a-guide-to-effective-go-documentation-952f346d073f

> They can also be detailed like the gob package’s overview. That package uses another convention for packages that need large amounts of introductory documentation: the package comment is placed in its own file, doc.go, which contains only those comments and a package clause.
Source: https://go.dev/blog/godoc

It's more a suggestion in fact.

We can live without it

ccoVeille:
I'll fix the commit messages

EDIT: done :heavy_check_mark:

dolmen (changes requested):
The license block is missing from the new doc.go so this is a blocker.

But in general I don't think that moving the package doc out of yaml.go is necessary.

ccoVeille:
> The license block is missing from the new doc.go so this is a blocker.

I hesitated about the license header.

Here was the result of my reflections.

- Apache 2 license doesn't require having a license header
- MIT neither
- the requirements are about having a LICENSE file for both and a NOTICE file for Apache2
- we satisfy these requirements
- existing files imported MIT project have a license header, because they need it, as stated in the NOTICE file
- here I created a new file, so I didn't need it.

From my understanding, we didn't have to.

I'm more inclined to move to SPDX identifiers instead of these headers.

Here is what Gemini provided when I ask a detailed status.

```
The Apache 2.0 license doesn't legally require a header in every single file, but it is highly recommended and considered a best practice.
Here's a breakdown of the requirements and recommendations:
 * The License Itself: The Apache 2.0 license states that when you distribute the work (or derivative works), you must give recipients a copy of the license. It also requires you to retain the original copyright, patent, trademark, and attribution notices.
 * The "Why" Behind File Headers: The purpose of putting a license header in each source file is to ensure that if a file is separated from the main project's LICENSE file, its licensing terms are still clear. Without an explicit header, a third party who uses that file might not know the terms of its use. This is particularly important when files are copied and used in other projects.
 * Apache Software Foundation's Stance: The Apache Software Foundation (ASF) itself recommends a header in each file. They provide a specific boilerplate header to use. This header is typically a short notice indicating that the file is licensed under the Apache License 2.0.
 * Requirements for Modified Files: If you modify a file that is under the Apache 2.0 license, you are required to add a prominent notice stating that you have changed the file. A license header is the most common and effective way to fulfill this requirement.
 * SPDX Identifiers: A growing practice, often used in conjunction with or as an alternative to the full header, is to include an SPDX license identifier. This is a machine-readable tag that makes it easy for automated tools to identify the license of a file. For Apache 2.0, the identifier is SPDX-License-Identifier: Apache-2.0.
In summary, while the license text itself doesn't explicitly say "you must put a header in every file," the requirements of the license and general best practices in open-source development make it a very strong recommendation. It's the most reliable way to ensure proper attribution and compliance when your code is distributed and reused.
```

So files could have a header, I didn't put one. Please also note my PR was opened before #6 was merged

ccoVeille:
Anyway … 

> But in general I don't think that moving the package doc out of yaml.go is necessary.

I already state this here

https://github.com/yaml/go-yaml/pull/72#issuecomment-3215241381

I will rollback this change as it's not necessary, and keep only the first commit.

ccoVeille:
I have dropped the "new file without license header" commit.

I opened https://github.com/yaml/go-yaml/issues/98, let's continue the discussion about license header there.

- https://github.com/yaml/go-yaml/issues/98

dolmen (approved):

ingydotnet (approved):

PR #73, opened 2025-08-18
fix: code behavior with old Go versions

### LINKED ISSUE #71: TypeError introduction broke compatibility with Go < 1.20
- https://github.com/yaml/go-yaml/pull/19

The code uses Unwrap() []error signature, which is only expected with Go 1.20+

ccoVeille:
- **ci: update CI to make sure we test code with all supported versions**
- **fix: Remove support of TypeErrors unwrapping before Go 1.20**

Fixes #71


yaml_go120.go
```diff
@@ -0,0 +1,12 @@
...
+
+package yaml
+
+func (e *TypeError) Unwrap() []error {
+	errs := make([]error, 0, len(e.Errors))
+	for _, err := range e.Errors {
+		errs = append(errs, err)
+	}
+	return errs
+}
```
ccoVeille: Surprisingly, this is not tested at all.

Tests should be added before merging

yaml.go
```diff
@@ -368,14 +368,6 @@ func (e *TypeError) Error() string {
 	return b.String()
 }
 
-func (e *TypeError) Unwrap() []error {
```
ccoVeille, line 371: we will need to document that TypeError cannot be Unwrapped before Go 1.20

ccoVeille, line 371: here also, moving all errors to an `errors.go` file could be a good thing.

so we would have an `errors_go1.20.go` file

It would be clearer

yaml_go120.go
ccoVeille: Also a good point was raised by @WGH- here:

> We could also provide [`Is(error)bool`](https://cs.opensource.google/go/go/+/refs/tags/go1.19.13:src/errors/wrap.go;l=50-52) and [`As(any)bool`](https://cs.opensource.google/go/go/+/refs/tags/go1.19.13:src/errors/wrap.go;l=96-98) methods manually. This would support `errors.Is` and `errors.As` for all nested errors on old Go versions, too.
> 
> (I still think it's a waste of efforts, though) 

 _Originally posted by @WGH- in [#71](https://github.com/yaml/go-yaml/issues/71#issuecomment-3197067592)_

ccoVeille:
This PR was replaced by #74

- #74

You can find the reason here

- https://github.com/yaml/go-yaml/issues/71#issuecomment-3197313189

yaml_go120.go
ccoVeille: This is what I implemented in fact.

It simplified everything

ccoVeille:
Based on the discussion we had on #74 

- #74 

And especially this comment

> It seems that Unwrap is indeed not strictly required since `Errors` is public: https://github.com/yaml/go-yaml/commit/3e63b6ac2aadb02a1945f9410d69c9aa9d6cad48#diff-acfe93d173e77894feaede834d0927584abe0fc551dba248d191e16edaa63611L327-L328.
> 
> However, not only `errors.Is()` uses `Unwrap`, but also `errors.As()`. And that check does imho make perfect sense. We're today doing this in our application:
> 
> ```go
> 	var (
> 		ype *yaml.ParserError
> 		yue *yaml.UnmarshalError
> 	)
> 	switch {
> 	case errors.As(err, &ype):
> 		res.Line = ype.Line
> 	case errors.As(err, &yue):
> 		res.Line = yue.Line
> 	}
> ```
> 
> This means we need the `Unwrap()` for this pattern?
_Originally posted by @andig in https://github.com/yaml/go-yaml/pull/74#issuecomment-3199938768_

I feel like this PR is enough.

I would love to get feedbacks from you guys @andig @WGH- @wwqgtxx

Thanks

yaml_go120.go
ccoVeille: TODO add test like @andig suggested

https://github.com/yaml/go-yaml/pull/73#issuecomment-3201969932

yaml.go
```diff
@@ -368,12 +368,28 @@ func (e *TypeError) Error() string {
...
+func (e *TypeError) As(target interface{}) bool {
 	for _, err := range e.Errors {
-		errs = append(errs, err)
+		if errors.As(err, target) {
+			return true
+		}
 	}
-	return errs
+	return false
 }
```
ccoVeille, line 393: TODO add a Co-authored-by tag to the commit, as it's the idea of @WGH- 

> We could also provide [`Is(error)bool`](https://cs.opensource.google/go/go/+/refs/tags/go1.19.13:src/errors/wrap.go;l=50-52) and [`As(any)bool`](https://cs.opensource.google/go/go/+/refs/tags/go1.19.13:src/errors/wrap.go;l=96-98) methods manually. This would support `errors.Is` and `errors.As` for all nested errors on old Go versions, too.
> 
> (I still think it's a waste of efforts, though) 

 _Originally posted by @WGH- in [#71](https://github.com/yaml/go-yaml/issues/71#issuecomment-3197067592)_

ccoVeille:
If you are wondering what the PR is about 

The code relies in the old way to unwrap things.

So it doesn't need an `Unwrap` method

https://cs.opensource.google/go/go/+/refs/tags/go1.19.13:src/errors/wrap.go;l=50-52

https://cs.opensource.google/go/go/+/refs/tags/go1.19.13:src/errors/wrap.go;l=96-98

This allows us to keep the feature without having to use `Unwrap []error` and then need a `//go:build` tag.

Recent Go versions that support Unwrap []error still supports `func Is(error) bool` and `As(error) bool` interfaces

https://cs.opensource.google/go/go/+/refs/tags/go1.24.0:src/errors/wrap.go;l=58-60

https://cs.opensource.google/go/go/+/refs/tags/go1.24.0:src/errors/wrap.go;l=122-124

yaml.go
ccoVeille, line 393: TODO: update the commit message with something based on this comment

- https://github.com/yaml/go-yaml/pull/73#issuecomment-3202005866

andig:
I prefer #73 over #74 since it maintains 1.20 error handling capabilities on lower versions while #74 just removes it. If we absolutely have to have lower Go version (do we? is there any concrete demand?) #73 is imho fine.

.github/workflows/go.yaml
```diff
@@ -27,8 +27,9 @@ jobs:
   build:
     strategy:
       matrix:
-        go-versions: [1.21.x, 1.22.x, 1.23.x, 1.24.x]
+        go-versions: ["1.16", "1.17", "1.18", "1.19", "1.20", "1.21", "1.22", "1.23", "1.24", "1.25"]
```
ingydotnet: Use single quotes unless double quotes are needed.

ingydotnet: Drop 1.16 and 1.17.

yaml_go120.go
ingydotnet: @andig please add a test for this!

ccoVeille: I started working on it for records.

It's almost ready locally.

ccoVeille: Here is the PR https://github.com/yaml/go-yaml/pull/82

- #82

PR #74, opened 2025-08-18
chore: remove useless Unwrapper on TypeError

### LINKED ISSUE #71: TypeError introduction broke compatibility with Go < 1.20

ccoVeille:
- **ci: update CI to make sure we test code with all supported versions**
- **fix: remove Unwrap []error method**

Fix #71 

EDIT: Replaced by #73

### DIFF
--- a/yaml.go
+++ b/yaml.go
@@ -368,14 +368,6 @@ func (e *TypeError) Error() string {
 	return b.String()
 }
 
-func (e *TypeError) Unwrap() []error {
-	errs := make([]error, 0, len(e.Errors))
-	for _, err := range e.Errors {
-		errs = append(errs, err)
-	}
-	return errs
-}
-
 type Kind uint32
 
 const (


ccoVeille:
@carloslima @andig 

Could you review this PR ?

I would like to know if you agree with the analysis I made:

> I'm convinced the best solution is simply to remove the Unwrap []error method
> 
> There is no need for someone to look for an exact errors, TypeError.Errors is a []*UnmarshalError
> 
> No one might or should use, errors.Is with an *UnmarshalError, error.Is is used for matching sentinel error.
> 
> No one would use ErrorAs with an UnmarshalError on a TypeError, it would match only the first error.
> 
> Now I said that, it's OK to remove the Unwrap because the TypeError.Errors field is exported and available.
> 
> Please note, also that the Unwrap method wasn't tested, the code to test it would have shown how strange it is to try doing it.
> 
https://github.com/yaml/go-yaml/issues/71#issuecomment-3197313189



Thanks

ccoVeille:
Also @WGH- and @wwqgtxx if you have a chance to review, I would appreciate

andig:
It seems that Unwrap is indeed not strictly required since `Errors` is public: https://github.com/yaml/go-yaml/commit/3e63b6ac2aadb02a1945f9410d69c9aa9d6cad48#diff-acfe93d173e77894feaede834d0927584abe0fc551dba248d191e16edaa63611L327-L328.

However, not only `errors.Is()` uses `Unwrap`, but also `errors.As()`. And that check does imho make perfect sense. We're today doing this in our application:

```go
	var (
		ype *yaml.ParserError
		yue *yaml.UnmarshalError
	)
	switch {
	case errors.As(err, &ype):
		res.Line = ype.Line
	case errors.As(err, &yue):
		res.Line = yue.Line
	}
```

This means we need the `Unwrap()` for this pattern?

wwqgtxx:
My suggestion is to place this function under the `//go:build 1.20` tag. 

`Unwrap` is still useful for accessing internal errors via `errors.Is` and `errors.As` and should not be removed directly.

ccoVeille:
Thanks for your feedbacks @andig @wwqgtxx 

It's clearer for me.

Then I think #73 would be enough, and there won't be a need for a `go:build` tag

- #73

Please review and comment the #73 PR.

I'm switching this PR (#74) as draft.

carloslima:
I've toyed with this for a while and I went both ways. Overall I think adding `Is()/As()` and maybe also `Unwrap() []error` matches people expectations on errors and would cause the least surprise.

I think a clear example of actual expected usage would make this easier to reason about.

The only error we wrap is `errors.errorString{}`. Does it make sense to match on that with `Is()/As()`?:

```
> git grep -B7 -A4 '&UnmarshalError'
(..)
decode.go:      d.terrors = append(d.terrors, &UnmarshalError{
decode.go-              Err:    fmt.Errorf("cannot unmarshal %s%s into %s", shortTag(tag), value, out.Type()),
decode.go-              Line:   n.Line,
decode.go-              Column: n.Column,
decode.go-      })
--
decode.go-      switch e := err.(type) {
decode.go-      case nil:
decode.go-              return true
decode.go-      case *TypeError:
decode.go-              d.terrors = append(d.terrors, e.Errors...)
decode.go-              return false
decode.go-      default:
decode.go:              d.terrors = append(d.terrors, &UnmarshalError{
decode.go-                      Err:    err,
decode.go-                      Line:   n.Line,
decode.go-                      Column: n.Column,
decode.go-              })
--
decode.go-      switch e := err.(type) {
decode.go-      case nil:
decode.go-              return true
decode.go-      case *TypeError:
decode.go-              d.terrors = append(d.terrors, e.Errors...)
decode.go-              return false
decode.go-      default:
decode.go:              d.terrors = append(d.terrors, &UnmarshalError{
decode.go-                      Err:    err,
decode.go-                      Line:   n.Line,
decode.go-                      Column: n.Column,
decode.go-              })
--
(..)
decode.go-                      if err != nil {
decode.go:                              d.terrors = append(d.terrors, &UnmarshalError{
decode.go-                                      Err:    err,
decode.go-                                      Line:   n.Line,
decode.go-                                      Column: n.Column,
decode.go-                              })
--
(..)
decode.go:                                      d.terrors = append(d.terrors, &UnmarshalError{
decode.go-                                              Err:    fmt.Errorf("mapping key %#v already defined at line %d", nj.Value, ni.Line),
decode.go-                                              Line:   nj.Line,
decode.go-                                              Column: nj.Column,
decode.go-                                      })
--
(..)
decode.go-                              if doneFields[info.Id] {
decode.go:                                      d.terrors = append(d.terrors, &UnmarshalError{
decode.go-                                              Err:    fmt.Errorf("field %s already set in type %s", name.String(), out.Type()),
decode.go-                                              Line:   ni.Line,
decode.go-                                              Column: ni.Column,
decode.go-                                      })
--
(..)
decode.go:                      d.terrors = append(d.terrors, &UnmarshalError{
decode.go-                              Err:    fmt.Errorf("field %s not found in type %s", name.String(), out.Type()),
decode.go-                              Line:   ni.Line,
decode.go-                              Column: ni.Column,
decode.go-                      })
```

Additionally, would it make sense for `Is()` on `TypeError` and `ParserError` to match on the type alone?
```
if errors.Is(err, yaml.TypeError{}) {...}
```
As in..
```
func (e *TypeError) Is(target error) bool {
       _, ok := target.(*TypeError)
       return ok
(..)
}
```

andig:
> Additionally, would it make sense for Is() on TypeError and ParserError to match on the type alone?

I don't think so- this is not what `Is()` is supposed to do.

ccoVeille:
I split you message in multiple parts, the order is the one from your message, but I will quote them in a different order for demonstration purpose.


> part 2 
> The only error we wrap is `errors.errorString{}`. Does it make sense to match on that with `Is()/As()`?:
> 

This is why I came to bring #74 instead of #73

Because I won't see people trying to use Is or As here.

That's why I removed the Unwrap in this PR.

> part 3
> Additionally, would it make sense for `Is()` on `TypeError` and `ParserError` to match on the type alone?
> ```
> if errors.Is(err, yaml.TypeError{}) {...}
> ```
> As in..
> ```
> func (e *TypeError) Is(target error) bool {
>        _, ok := target.(*TypeError)
>        return ok
> (..)
> }
> ```

I don't think it would make sense.

> part 1
> I've toyed with this for a while and I went both ways. Overall I think adding `Is()/As()` and maybe also `Unwrap() []error` matches people expectations on errors and would cause the least surprise.
> 
> I think a clear example of actual expected usage would make this easier to reason about.

But then, here I'm unsure what is the conclusion you came to.

The first part of your message let think we could use define As and Is interfaces

But the next ones, where you ask questions let me think you came to the conclusion, there is no need to have something to unwrap.

So I'm lost, I'm sorry 😅

andig:
I prefer #73 over #74 since it maintains 1.20 error handling capabilities on lower versions while #74 just removes it. If we absolutely have to have lower Go version (do we? is there any concrete demand?) #73 is imho fine.

carloslima:
> So I'm lost, I'm sorry

No worries, it's my fault :joy: 
It's late here and I think my message was unclear :sweat_smile: 

> The first part of your message let think we could use define As and Is interfaces

Correct, at the end that's what I think is the best option mostly because it matches expectations around errors. Convention more than actual usefulness.

I understand the point you made and I think it makes sense based on the errors we generate. I can't see them being very helpful to match against. I was just listing the info I was basing my answer on in case I misunderstood or missed something.

What I think was confusing in my message was the request to have actual expected usage scenarios in the form of test cases. Given that @andig and @wwqgtxx seem to have use cases in mind, I'd like for those to be codified as tests, making it easier for us to reason about and identify what's being supported without having to scratch our heads again in the future.

andig:
UC: use `errors.As` against set of multiple `UnmarshalErrors`, matching the first one.

ccoVeille:
> I prefer #73 over #74 since it maintains 1.20 error handling capabilities on lower versions while #74 just removes it. If we absolutely have to have lower Go version (do we? is there any concrete demand?) #73 is imho fine.
Source @andig :  https://github.com/yaml/go-yaml/pull/74#issuecomment-3202396233

Good then. Could you review it and approve it? Thanks

> What I think was confusing in my message was the request to have actual expected usage scenarios in the form of test cases. Given that @andig and @wwqgtxx seem to have use cases in mind, I'd like for those to be codified as tests, making it easier for us to reason about and identify what's being supported without having to scratch our heads again in the future.
Source @carloslima https://github.com/yaml/go-yaml/pull/74#issuecomment-3202405222

I agree.

Do you think guys could add unit tests to my PR ?


> UC: use `errors.As` against set of multiple `UnmarshalErrors`, matching the first one.
Source @andig  https://github.com/yaml/go-yaml/pull/74#issuecomment-3202415331

Does anyone see a need for being able to use errors.Is ? If not, I would like to suggest to remove the `Is(error) bool` interface and only keep the `As(error) bool`but maybe I'm going too far, as there is no harm to leave it even if no one will use it.

andig:
I don‘t.

andig:
Btw, note https://github.com/golang/go/issues/51945

PR #75, opened 2025-08-19
test: reproducer for incorrect indentation hint that fails validation

earl-warren:
Refs https://github.com/yaml/go-yaml/issues/76

---

The tests shows that Unmarshal() on the following (note the empty line following the literal scalar):

```yaml
jobs:
      - run: |

             uname -a
```

Will be Marshal() as:

```yaml
jobs:
    - run: |4

        uname -a
```

Which fails Unmarshal() because there is a two space indentation although the [block indentation indicator](https://yaml.org/spec/1.2.2/#8111-block-indentation-indicator) is set to four.

`decode_test.go:...: Unmarshal() returned error: yaml: line 1: did not find expected key`

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -895,6 +895,82 @@ func TestUnmarshal(t *testing.T) {
 	}
 }
 
+func TestUnmarshalScalar(t *testing.T) {
+	var unmarshalTests = []struct {
+		data  string
+		value interface{}
+	}{
+		{
+			`run: |
+
+     uname -a
+`,
+			map[string]interface{}{
+				"run": "\nuname -a\n",
+			},
+		},
+		{
+			`run: |
+     uname -a
+`,
+			map[string]interface{}{
+				"run": "uname -a\n",
+			},
+		},
+		{
+			`run: |2
+     uname -a
+`,
+			map[string]interface{}{
+				"run": "   uname -a\n",
+			},
+		},
+		{
+			`jobs:
+      - run: |
+
+             uname -a
+`,
+			map[string]interface{}{
+				"jobs": []interface{}{
+					[]interface{}{
+						map[string]interface{}{
+							"run": "   uname -a\n",
+						},
+					},
+				},
+			},
+		},
+	}
+
+	for i, item := range unmarshalTests {
+		t.Run(fmt.Sprintf("test %d: %q", i, item.data), func(t *testing.T) {
+			typ := reflect.ValueOf(item.value).Type()
+			value := reflect.New(typ)
+			err := yaml.Unmarshal([]byte(item.data), value.Interface())
+			if _, ok := err.(*yaml.TypeError); !ok {
+				if err != nil {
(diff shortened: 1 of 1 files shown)


ccoVeille:
May I assume the `|4` is a typo in your example?

earl-warren:
It is a [block indentation indicator](https://yaml.org/spec/1.2.2/#8111-block-indentation-indicator), this is copy/pasted from the  `Marshal()` output, as shown in the test output.

ccoVeille:
It's indeed not that obvious.

Thanks for the explanation and documentation link

decode_test.go
```diff
@@ -899,6 +899,82 @@ func TestUnmarshal(t *testing.T) {
 	}
 }
 
+func TestUnmarshalScalar(t *testing.T) {
+	var unmarshalTests = []struct {
+		data  string
+		value interface{}
+	}{
+		{
```
ingydotnet, line 903: Can you write the raw inputs like
```
                {
`
run: |
     
     uname -a
`,
```
Then use something like `strings.TrimLeft(s1, "\n")` to remove the leading newline?

I think that would make the tests read much better.

ingydotnet:
> Which fails Unmarshal() because there is a two space indentation although the [block indentation indicator](https://yaml.org/spec/1.2.2/#8111-block-indentation-indicator) is set to four.

https://play.yaml.io/main/parser?input=am9iczoKICAgIC0gcnVuOiB8NAoKICAgICAgICB1bmFtZSAtYQ==

I had to double-check that myself :)

decode_test.go
earl-warren: I'll do that. 

I fully expect those tests to be trashed / replaced by a more idiomatic / appropriate way to verify a fix. Their value is solely to explain with code rather than plain English how this border case is not a figment of my imagination.

ingydotnet: Ah, I see. Is this better as an issue rather than a PR?

earl-warren: AFAIK issues cannot be run but if you enable the workflows for this PR it will show the test that fails. It is in support of the associated issue https://github.com/yaml/go-yaml/issues/76

ingydotnet:
This is interesting because using https://github.com/yaml/go-yaml-cli

```
$ ./go-yaml -Y <<<"\
jobs:
  - run: |2

      uname -a
"
jobs:
  - run: |2

      uname -a
```

I'd guess that is because this tool uses Decode rather then Unmarshall.

I'll add an option to load inputs with Unmarshal, and dump outputs with Marshal.

ingydotnet:
Closing this due to https://github.com/yaml/go-yaml/issues/76#issuecomment-3217258282

I updated https://github.com/yaml/go-yaml-cli to support `-u` and `-m` for using Unmarshal and Marshal instead of Decode and Encode respectively.

Using -m I was able to repro your problem. It exists even with equivalent JSON input because I determined that Marshal (or Decode) were not contributing to the issue.

PR #77, opened 2025-08-19
Forbid unmarshaling non-scalars into a TextUnmarshaler

WGH-:
This is a breaking change.

Previously an attempt to unmarshal a YAML mapping into a struct implementing `encoding.TextUnmarshaler` would succeed, but likely do the wrong thing.

Consider this example:

    type Foo struct {
        IP netip.Addr `yaml:"ip"`
    }

`netip.Addr` is a struct type implementing `TextUnmarshaler`, but containing no exported fields. So unmarshaling `{"ip": "127.0.0.1"}` would use `TextUnmarshaler` (as expected), but `{"ip": {"whatever": 123}}` would silently do nothing.

With this commit, it will now result in a type error.

`encoding/json/v2` documentation states:

> If the value type implements encoding.TextUnmarshaler, then the input
> is decoded as a JSON string and the UnmarshalText method is called with
> the decoded string value. This fails with a SemanticError if the input
> is not a JSON string.

`encoding/json` documentation is somewhat ambiguous on this, but it also behaves the same way (https://go.dev/play/p/z9okIXkUzvQ).

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -476,6 +476,24 @@ func allowedAliasRatio(decodeCount int) float64 {
 	}
 }
 
+func isTextUnmarshaler(out reflect.Value) bool {
+	// Dereference pointers to check the underlying type,
+	// similar to how prepare() handles Unmarshaler checks.
+	for out.Kind() == reflect.Pointer {
+		if out.IsNil() {
+			// Create a new instance to check the type
+			out = reflect.New(out.Type().Elem()).Elem()
+		} else {
+			out = out.Elem()
+		}
+	}
+	if out.CanAddr() {
+		_, ok := out.Addr().Interface().(encoding.TextUnmarshaler)
+		return ok
+	}
+	return false
+}
+
 func (d *decoder) unmarshal(n *Node, out reflect.Value) (good bool) {
 	d.decodeCount++
 	if d.aliasDepth > 0 {
@@ -488,6 +506,23 @@ func (d *decoder) unmarshal(n *Node, out reflect.Value) (good bool) {
 		out.Set(reflect.ValueOf(n).Elem())
 		return true
 	}
+
+	// When out type implements [encoding.TextUnmarshaler], ensure the node is
+	// a scalar. Otherwise, for example, unmarshaling a YAML mapping into
+	// a struct having no exported fields, but implementing TextUnmarshaler
+	// would silently succeed, but do nothing.
+	//
+	// Note that this matches the behavior of both encoding/json and encoding/json/v2.
+	if n.Kind != ScalarNode && isTextUnmarshaler(out) {
+		err := fmt.Errorf("cannot unmarshal %s%s into %s (TextUnmarshaler)", shortTag(n.Tag), n.Value, out.Type())
+		d.terrors = append(d.terrors, &UnmarshalError{
+			Err:    err,
+			Line:   n.Line,
+			Column: n.Column,
+		})
+		return false
+	}
+
 	switch n.Kind {
 	case DocumentNode:
 		return d.document(n, out)
--- a/decode_test.go
+++ b/decode_test.go
@@ -101,6 +101,30 @@ type (
 	}
 )
 
+// simpleTextUnmarshaler is a simple type implementing encoding.TextUnmarshaler
+// for testing TextUnmarshaler validation.
+type simpleTextUnmarshaler struct {
+	Value string
+}
+
+func (s *simpleTextUnmarshaler) UnmarshalText(text []byte) error {
+	s.Value = string(text)
+	return nil
+}
+
+// Test types for TextUnmarshaler validation
+type (
+	testStructA_TextUnmarshaler struct {
+		A simpleTextUnmarshaler
+	}
+	testStructA_TextUnmarshalerPtr struct {
+		A *simpleTextUnmarshaler
+	}
+	testStructA_TextUnmarshalerPtrPtr struct {
+		A **simpleTextUnmarshaler
+	}
+)
+
 // Type and value registries for data-driven tests
 var (
 	decodeTypes  = datatest.NewTypeRegistry()
@@ -204,6 +228,11 @@ func init() {
 	decodeTypes.Register("testStructA_Int_InlineBPtr", testStructA_Int_InlineBPtr{})
 	decodeTypes.Register("testStructA_Int_InlineDPtr", testStructA_Int_InlineDPtr{})
 	decodeTypes.Register("testStructA_Int_InlineMapStringInt", testStructA_Int_InlineMapStringInt{})
+
+	// Register TextUnmarshaler test types
+	decodeTypes.Register("testStructA_TextUnmarshaler", testStructA_TextUnmarshaler{})
+	decodeTypes.Register("testStructA_TextUnmarshalerPtr", testStructA_TextUnmarshalerPtr{})
+	decodeTypes.Register("testStructA_TextUnmarshalerPtrPtr", testStructA_TextUnmarshalerPtrPtr{})
 
 	// Register math constants
 	decodeValues.Register("+Inf", math.Inf(+1))
@@ -717,12 +746,24 @@ func runUnmarshalErrorTest(t *testing.T, tc map[string]any) {
 	yamlInput := datatest.RequireString(t, tc, "yaml")
 	expectedError := datatest.RequireString(t, tc, "want")
 
-	var value any
-	err := yaml.Unmarshal([]byte(yamlInput), &value)
-	if err == nil {
-		t.Fatalf("got nil; want error %q - Partial unmarshal: %#v", expectedError, value)
+	var target any
+	// If a type is specified, use it; otherwise default to any
+	if typeName, ok := tc["output_type"].(string); ok {
+		var err error
+		target, err = decodeTypes.NewPointerInstance(typeName)
+		if err != nil {
+			t.Fatalf("failed to create instance of type %s: %v", typeName, err)
+		}
+	} else {
+		var value any
+		target = &value
 	}
-	assert.Equalf(t, expectedError, err.Error(), "Partial unmarshal: %#v", value)
+
+	err := yaml.Unmarshal([]byte(yamlInput), target)
+	if err == nil {
+		t.Fatalf("got nil; want error %q - Partial unmarshal: %#v", expectedError, target)
+	}
+	assert.Equalf(t, expectedError, err.Error(), "Partial unmarshal: %#v", target)
 }
 
 func TestDecoderErrors(t *testing.T) {
@@ -734,12 +775,24 @@ func TestDecoderErrors(t *testing.T) {
 			yamlInput := datatest.RequireString(t, tc, "yaml")
 			expectedError := datatest.RequireString(t, tc, "want")
 
-			var value any
-			err := yaml.NewDecoder(strings.NewReader(yamlInput)).Decode(&value)
-			if err == nil {
-				t.Fatalf("got nil; want error %q - Partial decode: %#v", expectedError, value)
+			var target any
+			// If a type is specified, use it; otherwise default to any
+			if typeName, ok := tc["output_type"].(string); ok {
+				var err error
+				target, err = decodeTypes.NewPointerInstance(typeName)
+				if err != nil {
+					t.Fatalf("failed to create instance of type %s: %v", typeName, err)
+				}
+			} else {
+				var value any
+				target = &value
 			}
-			assert.Equalf(t, expectedError, err.Error(), "Partial decode: %#v", value)
+
+			err := yaml.NewDecoder(strings.NewReader(yamlInput)).Decode(target)
+			if err == nil {
+				t.Fatalf("got nil; want error %q - Partial decode: %#v", expectedError, target)
+			}
+			assert.Equalf(t, expectedError, err.Error(), "Partial decode: %#v", target)
 		},
 	})
 }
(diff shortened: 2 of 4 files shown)


decode_test.go
```diff
@@ -1575,6 +1575,25 @@ func TestTextUnmarshalerError(t *testing.T) {
...
+
+	for _, input := range inputs {
+		err := yaml.Unmarshal([]byte(input), &dst)
+		t.Logf("%s -> err=%v", input, err)
+		var target *yaml.TypeError
+		if !errors.As(err, &target) {
+			t.Errorf("expected yaml.TypeError, got %v", err)
+		}
+	}
+}
```
ccoVeille, line 1294: Could you also add a test with net.Ip the example will help I feel.

decode.go
```diff
@@ -791,6 +791,18 @@ func (d *decoder) mapping(n *Node, out reflect.Value) (good bool) {
 			return false
 		}
 	}
+	if out.CanAddr() {
+		_, ok := out.Addr().Interface().(encoding.TextUnmarshaler)
+		if ok {
```
ccoVeille: The breaking change is the fact it's now impossible to unmarshal a "ip": "127.0.0.1" to a net.Ip ?

I mean it would be forbidden, even if it was working before.

WGH-: It has never been possible to unmarshal `{"ip": "127.0.0.1"}` into `net.IP` because `net.IP` is a slice.  It falls into this case: https://github.com/yaml/go-yaml/blob/cb1771ea6b0c7dde0172714ae121d453839ad663/decode.go#L807-L809

decode_test.go
WGH-: How is this different from other `enconding.TextUnmarshaler` tests which use the locally defined `textUnmarshaler` type?

decode.go
ccoVeille: OK then

decode_test.go
ccoVeille: I assumed it would be clearer to see somehow a real life example that was working before.

decode.go
ccoVeille: But what is the breaking change then?

Is there something that was working before this PR and that would no longer work after merging it?

I feel I'm lost here.

Because if the breaking change is about having something that never worked that will now report an error instead of not failing. I would call it a bug fix, more than a breaking change.

But as I said, I'm unsure I got what this was about.

Usually when I work on a "breaking" change I tend to add a commit that add a test that will show how it was "working" before, and then I update the test to comply with the new code in the commit that change the code. This way it's easier to understand.

But that's my way to do, I don't expect people to do it 😅

But one last time, I might have misunderstood the whole purpose of this change.

WGH-: My example was `netip.Addr`, which, unlike `net.IP`,  is a struct. Maybe that was a confusing example.

I can give you an example of some possibly legitimate use that will break.

```go
type Foo struct {
	X string
	Y string
}

func (f *Foo) UnmarshalText(b []byte) error {
	f.X, f.Y, _ = strings.Cut(string(b), "/")
	return nil
}
```

You can unmarshal `{"x": 1, "y": 2}` into `Foo` right now, but my PR forbids it. Note, however, that `encoding/json` doesn't allow it as well, and returns a similar error.

ccoVeille: OK, so the fact you replied me something I didn't get is because I wrongly used net.Ip in my example instead of net.Addr

ingydotnet:
This PR _seems_ good to me but I'd like to see more discussion.
It's technically breaking but do people think it will actually break anything?
Is there any way to search for public code that it would possibly break?

WGH-:
> This PR _seems_ good to me but I'd like to see more discussion. It's technically breaking but do people think it will actually break anything? Is there any way to search for public code that it would possibly break?

I've been thinking about this for a bit, but I have no good ideas how to do so.

We could port this patch to v3, and run tests across popular repositories with `go mod replace`d dependency, but I doubt that YAML unmarshalling is covered by unit tests well.

Going the conservative way, we could add a deprecation warning, and an option to opt in immediately.

SVilgelm:
I would suggest to do same as json/v2 - unmarshal to a string first, then pass it to `TextUnmarshaler`

SVilgelm:
@copilot create an issue based on this PR

WGH-:
> I would suggest to do same as json/v2 - unmarshal to a string first, then pass it to `TextUnmarshaler`

This PR already does the same thing as json/v2.

WGH-:
The only difference perhaps is that handling of `TextUnmarshaler` is somewhat less explicit. [`encoding/json/v2` switches on target type](https://cs.opensource.google/go/go/+/refs/tags/go1.25.3:src/encoding/json/v2/arshal_methods.go;l=217-249), whereas `go-yaml` switches on YAML value type first, and only then methods `(*decoder).mapping`, `(*decoder).sequence`, etc. check the target type.

...now that I spelled this out, I think odd cases like unmarshaling a YAML sequence into a slice type implementing `TextUnmarshaler` might be handled differently to `encoding/json/v2`. Thanks for pointing that out, I'll check it again.

decode.go
```diff
@@ -791,6 +791,18 @@ func (d *decoder) mapping(n *Node, out reflect.Value) (good bool) {
...
+				Err:    err,
+				Line:   n.Line,
+				Column: n.Column,
+			})
+			return false
+		}
+	}
 	switch out.Kind() {
 	case reflect.Struct:
 		return d.mappingStruct(n, out)
```
ccoVeille: I think we have to be careful here. I might be wrong here but it's not because a struct implement a TextUnmarshaler that it's OK to return an error.

A struct could have exported fields, and could work, but there is a TextUnmarshaler defined on the struct for some reason.

WGH-: Erm, no, that's what I actually intended, and that's how both `encoding/json` and `encoding/json/v2` behave. Implements `TextUnmarshaler` = accepts only YAML strings. No heuristics like "there're some exported fields, maybe it's OK to bypass `TextUnmarshaler`".

ccoVeille: OK, then. Thanks for checking

WGH-:
I've updated this by making the check more explicit.

ingydotnet:
I've been asking claude to explain some of the go-yaml PRs to me. 
Here's one for this: https://gist.github.com/ingydotnet/d4d933a9ba7513ac7728b6047fa47b0c

That helps me a lot. Assuming it is correct, I approve.

The new options system will let us deal with any unexpected problems, should they ever arise.

ingydotnet (approved):
👍

decode.go
```diff
@@ -488,6 +496,23 @@ func (d *decoder) unmarshal(n *Node, out reflect.Value) (good bool) {
...
 		return true
 	}
+
+	// When out type implements [encoding.TextUnmarshaler], ensure the node is
+	// a scalar. Otherwise, for example, unmarshaling a YAML mapping into
+	// a struct having no exported fields, but implementing TextUnmarshaler
+	// would silently succeed, but do nothing.
+	//
+	// Note that this matches the behavior of both encoding/json and encoding/json/v2.
+	if n.Kind != ScalarNode && isTextUnmarshaler(out) {
```
ccoVeille, line 516: I'm a bit worried about the unserialization of YAML NULL into a pointer to a struct that would handle a TextUnmarshaler

ingydotnet:
@ccoVeille didn't notice your comment here. Will address later today.

decode_test.go
```diff
@@ -1218,6 +1218,25 @@ func TestUnmarshalError_Unwrapping(t *testing.T) {
...
```
ccoVeille, line 1294: Because of https://github.com/yaml/go-yaml/pull/77/changes#r2638137604

I worked on adding some example to tests

```suggestion
func TestTextUnmarshalerNonScalar(t *testing.T) {
	type testStruct struct {
		A textUnmarshaler
		B *textUnmarshaler
	}

	inputs := []struct {
		input string
		want  *testStruct
	}{
		// Invalid cases
		{input: `a: {}`},
		{input: `a: {"foo": "bar"}`},
		{input: `a: []`},
		{input: `a: NULL`},

		{input: `b: {}`},
		{input: `b: {"foo": "bar"}`},
		{input: `b: []`},

		// Valid cases for control
		{input: `a: "hello"`, want: &testStruct{A: textUnmarshaler{S: "hello"}}},
		{input: `b: "hello"`, want: &testStruct{B: &textUnmarshaler{S: "hello"}}},
		{input: `b: NULL`, want: &testStruct{B: nil}},
	}

	for _, tt := range inputs {
		var dst testStruct
		err := yaml.Unmarshal([]byte(tt.input), &dst)
		if tt.want != nil {
			if err != nil {
				t.Errorf("unexpected error for input %#q: %v", tt.input, err)
			}

			if !reflect.DeepEqual(dst.A, tt.want.A) {
				t.Errorf("for input %#q, dst.A = %#v; want %#v", tt.input, dst.A, tt.want.A)
			}
			if !reflect.DeepEqual(dst.B, tt.want.B) {
				t.Errorf("for input %#q, dst.B = %#v; want %#v", tt.input, dst.B, tt.want.B)
			}
			continue
		}

		if err == nil {
			t.Errorf(
				"expected error for input %#q, got nil\n"+
					"dst.A: %#v\n"+
					"dst.B: %#v",
				tt.input, dst.A, dst.B,
			)
			continue
		}
		var target *yaml.TypeError
		if !errors.As(err, &target) {
			t.Errorf("expected yaml.TypeError for input %#q, got %v", tt.input, err)
		}
	}
}

```

Here is what I got without this PR

```
--- FAIL: TestTextUnmarshalerNonScalar (0.00s)
    decode_test.go:1241: expected error for input `a: {}`, got nil
        dst.A: yaml_test.textUnmarshaler{S:""}
        dst.B: (*yaml_test.textUnmarshaler)(nil)
    decode_test.go:1241: expected error for input `a: {"foo": "bar"}`, got nil
        dst.A: yaml_test.textUnmarshaler{S:""}
        dst.B: (*yaml_test.textUnmarshaler)(nil)
    decode_test.go:1241: expected error for input `a: NULL`, got nil
        dst.A: yaml_test.textUnmarshaler{S:""}
        dst.B: (*yaml_test.textUnmarshaler)(nil)
    decode_test.go:1241: expected error for input `b: {}`, got nil
        dst.A: yaml_test.textUnmarshaler{S:""}
        dst.B: &yaml_test.textUnmarshaler{S:""}
    decode_test.go:1241: expected error for input `b: {"foo": "bar"}`, got nil
        dst.A: yaml_test.textUnmarshaler{S:""}
        dst.B: &yaml_test.textUnmarshaler{S:""}
FAIL
```

Here is what I get with this PR

```
--- FAIL: TestTextUnmarshalerNonScalar (0.00s)
    decode_test.go:1241: expected error for input `a: NULL`, got nil
        dst.A: yaml_test.textUnmarshaler{S:""}
        dst.B: (*yaml_test.textUnmarshaler)(nil)
    decode_test.go:1241: expected error for input `b: {}`, got nil
        dst.A: yaml_test.textUnmarshaler{S:""}
        dst.B: &yaml_test.textUnmarshaler{S:""}
    decode_test.go:1241: expected error for input `b: {"foo": "bar"}`, got nil
        dst.A: yaml_test.textUnmarshaler{S:""}
        dst.B: &yaml_test.textUnmarshaler{S:""}
```

I feel like there is something strange that still need to be addressed.

- Why would `a: NULL` won't lead to an error ?

This one might be valid, but it's a bit strange to me.

Are these two exact equivalent in YAML ?

```yaml
a: NULL
b: whatever
```

and 

```yaml
b: whatever
```

- `b: {}` and `b: {"foo": "bar"}` leads to no error, but why ?

It's not logic to be to have `&yaml_test.textUnmarshaler{S:""}` for B

The good news is that `b: NULL` seems safe

decode.go
ccoVeille: It seems safe apparently

colinjlacy:
Quick sanity check question here - is it worth packing this as part of the upcoming v4 release, rather than releasing a breaking change to v3?

ingydotnet:
> Quick sanity check question here - is it worth packing this as part of the upcoming v4 release, rather than releasing a breaking change to v3?

From a slack convo with @colinjlacy 

My point Is that we want to make the defaults be a lot more clean and YAML idiomatic. We don't need to add an option for every little thing unless it becomes a problem.

We should just make the default behavior for go-yaml to be best practice; But we have an escape hatch if it causes real problems for people

People using load and dump with yaml.v3, obviously aren't using v3. they are using v4 with a certain set of options to get v3 behavior. It's unclear how granular that will have to be. But I would like to lean towards breaking eggs since it's easy to clean up the mess now

I think this will be a huge win for us because now our hands are no longer tied.

I think we'll just have to make it clear in the docs that it's possible v4 will break things for users transitioned v3 code, but just let us know and we'll take care of it. (or join us with a PR!)

decode_test.go
ccoVeille: @ingydotnet @colinjlacy @carloslima 

What do you think about what I have reported ?

ingydotnet: I don't really understand it. Can you boil it down to a couple sentences, that I can either agree with or disagree with?

ccoVeille: This PR addresses

`v: {foo: bar}` reporting no error when using Load to a struct with TextUnmarshaler 

But I noticed that:

- using Load on `v: {foo: bar}` to a TextUnmarshaler with a pointer receiver reports no error

It doesn't look OK. There is something else to fix in this PR

- using Load on `v: NULL` to a TextUnmarshaler reports no error

It doesn't sound normal to me. But yet, it's a different issue than the one this PR tries to address. But I found it while trying to understand the changes involved in the PR

ingydotnet: @ccoVeille see bb19c74 for fixes and tests for this.

ccoVeille: It sounds good. Thanks

PR #79, opened 2025-08-21
Multiline scalars w/ trailing ws can emit literal

### LINKED ISSUE #78: Trailing line spaces in string block changes to a quoted string
Note: this is a clone of https://github.com/go-yaml/yaml/issues/880, with slight modification to make it more succinct.

data1.yaml

```yaml
multi-line-string-with-trailing-space-chars: |
  this line does not have trailing whitespace
  this line has trailing space characters   
  this line does not have trailing whitespace
multi-line-string-with-trailing-tab-chars: |
  this line does not have trailing whitespace
  this line has trailing tab characters			
  this line does not have trailing whitespace
multi-line-string-without-trailing-whitespace: |
  this line does not have trailing whitespace
  this line does not have trailing whitespace
  this line does not have trailing whitespace
```

example code:

```go
package main

import (
	"fmt"
	"os"

	"go.yaml.in/yaml/v4"
)

func main() {
	data, _ := os.ReadFile("data1.yaml")

	ymlData := make(map[string]any)

	yaml.Unmarshal(data, &ymlData)

	newData, _ := yaml.Marshal(ymlData)
	fmt.Println(string(newData))
}
```

output yaml:

```yaml
multi-line-string-with-trailing-space-chars: "this line does not have trailing whitespace\nthis line has trailing space characters   \nthis line does not have trailing whitespace\n"
multi-line-string-with-trailing-tab-chars: |
    this line does not have trailing whitespace
    this line has trailing tab characters			
    this line does not have trailing whitespace
multi-line-string-without-trailing-whitespace: |
    this line does not have trailing whitespace
    this line does not have trailing whitespace
    this line does not have trailing whitespace


```

Observed behaviors:
- Multi-line strings that have lines with trailing space (0x20) characters will be converted into double-quoted strings. Trailing spaces are preserved.
- Multi-line strings that have lines with trailing tab (0x09) characters will remain as a multi-line string. Trailing tabs are preserved.

Expectations:
- Multi-line strings that have lines with trailing space (0x20) characters would remain multi-line strings, and their trailing spaces would be preserved.

ingydotnet:
Fixes #78

### DIFF
--- a/emitterc.go
+++ b/emitterc.go
@@ -1026,7 +1026,6 @@ func yaml_emitter_check_simple_key(emitter *yaml_emitter_t) bool {
 
 // Determine an acceptable scalar style.
 func yaml_emitter_select_scalar_style(emitter *yaml_emitter_t, event *yaml_event_t) bool {
-
 	no_tag := len(emitter.tag_data.handle) == 0 && len(emitter.tag_data.suffix) == 0
 	if no_tag && !event.implicit && !event.quoted_implicit {
 		return yaml_emitter_set_emitter_error(emitter, "neither tag nor implicit flags are specified")
@@ -1283,7 +1282,7 @@ func yaml_emitter_analyze_tag(emitter *yaml_emitter_t, tag []byte) bool {
 }
 
 // Check if a scalar is valid.
-func yaml_emitter_analyze_scalar(emitter *yaml_emitter_t, value []byte) bool {
+func yaml_emitter_analyze_scalar(emitter *yaml_emitter_t, event *yaml_event_t) bool {
 	var (
 		block_indicators   = false
 		flow_indicators    = false
@@ -1304,6 +1303,7 @@ func yaml_emitter_analyze_scalar(emitter *yaml_emitter_t, value []byte) bool {
 		previous_break         = false
 	)
 
+	value := event.value
 	emitter.scalar_data.value = value
 
 	if len(value) == 0 {
@@ -1407,8 +1407,13 @@ func yaml_emitter_analyze_scalar(emitter *yaml_emitter_t, value []byte) bool {
 		emitter.scalar_data.flow_plain_allowed = false
 		emitter.scalar_data.block_plain_allowed = false
 	}
-	if trailing_space {
+	if trailing_space || special_characters {
 		emitter.scalar_data.block_allowed = false
+	}
+	if space_break {
+		if event.scalar_style() != yaml_LITERAL_SCALAR_STYLE {
+			emitter.scalar_data.block_allowed = false
+		}
 	}
 	if break_space {
 		emitter.scalar_data.flow_plain_allowed = false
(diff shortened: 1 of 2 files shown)


emitterc.go
```diff
@@ -1283,7 +1282,7 @@ func yaml_emitter_analyze_tag(emitter *yaml_emitter_t, tag []byte) bool {
 }
 
 // Check if a scalar is valid.
-func yaml_emitter_analyze_scalar(emitter *yaml_emitter_t, value []byte) bool {
+func yaml_emitter_analyze_scalar(emitter *yaml_emitter_t, event *yaml_event_t) bool {
```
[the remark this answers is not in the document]
ingydotnet: Team, this is a private function only used within this file. 

Is there any chance that this could be problematic?

ingydotnet:
I have some issues with this fix.
It changes how YAML is emitted.
> See how it broke 3 existing tests, that I needed to change.

That means it also changes `yq` even though it is reported as a bug in yq: https://github.com/mikefarah/yq/issues/566

In v4 I'd like to have a _lot_ more options on how things are emitted.
We might want to make this new formatting only available as an option.

I'd like to hear your thoughts. (Yes, you!)

justfalter:
I get the impression from `Node.Style`'s docs that the encoder would obey the style when non-zero, so maybe there should be tests validating as much? https://github.com/yaml/go-yaml/blob/7f7bed120dc891d76a7b9b965d56ea272c8fe56a/yaml.go#L433-L434

After all, it seems that `Node.SetString` exists to assist with at least partially determining the appropriate style https://github.com/yaml/go-yaml/blob/7f7bed120dc891d76a7b9b965d56ea272c8fe56a/yaml.go#L540-L554

And, when decoding a literal string to a `Node`, the node's Style is `LiteralStyle` ...  so it would make sense to reflect that when encoding.

emitterc.go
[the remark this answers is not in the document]
ccoVeille: Nope. Unexported methods can be renamed and changed at anytime.

encode_test.go
```diff
@@ -1104,19 +1104,24 @@ func TestWhitespaceWithContent(t *testing.T) {
...
 			"  hello  \n",
-			"\"  hello  \\n\"\n",
+			"|4\n      hello  \n",
 			"Multiple spaces + text + spaces + newline - should be double quoted (ends with spaces)",
 		},
 		{
 			"hello  \n",
-			"\"hello  \\n\"\n",
+			"|\n    hello  \n",
 			"Text + spaces + newline - should be double quoted (ends with spaces)",
```
ccoVeille, line 1113: You talked about the change here

https://github.com/yaml/go-yaml/pull/79#issuecomment-3208952435

There is clear problem indeed

Not only the format changes (why not, after all) , but here the trailing spaces before the \n are lost

I feel like it's not a good thing.

ingydotnet: > Not only the format changes (why not, after all) , but here the trailing spaces before the \n are lost

I'm confused.
How are they lost?

ccoVeille: I'm sorry. I feel like my comment was based on a misunderstanding.

Now, I feel like I didn't understand clearly what your change is about.

The spaces are indeed there. I was confused by the presence of \n as a first character of the marshaled.

I feel like I didn't understand the change involved here. So I'll pass on this PR.

PR #82, opened 2025-08-21
chore: improve errors unwrapping

ccoVeille:
- **fix: assert.ErrorIs arguments order**
- **chore: simplify error message in assertion tests**
- **chore: add ErrorAs assertion helper**
- **chore: add missing tests for errors unwrapping**
- **fix: improve the error unwrapping**


stefanprodan (approved):
LGTM

internal/testutil/assert/assert.go
```diff
@@ -73,12 +73,48 @@ func ErrorMatchesf(tb miniTB, pattern string, err error, msgFormat string, args
 	}
 }
 
-func ErrorIs(tb miniTB, want, got error) {
+func ErrorIs(tb miniTB, got, target error) {
```
ingydotnet: I prefer `want` to `target` here.

ingydotnet (changes requested):

internal/testutil/assert/assert.go
ccoVeille: I used the errors.Is idiomatic arguments

https://pkg.go.dev/errors#Is

But I can change for sure

ingydotnet (approved):

internal/testutil/assert/assert.go
ccoVeille: I now use `got, want`

ingydotnet: thanks

PR #83, opened 2025-08-21
Add a GitHub Actions workflow to check commit messages

ccoVeille:
The following rules were implemented:

1. Subject must be shorter than 50 characters
2. Subject must not end with a period
3. Subject must start with a capital letter

Closes #81

PR #86, opened 2025-08-23
Add a yamllint GitHub Action

### LINKED ISSUE #84: CI: link YAML files in repository
We have various YAML file out of `testdata`

We should lint them.

Here are solid candidates

- [yamlfmt](https://github.com/google/yamlfmt)
- [yamlfix](https://github.com/lyz-code/yamlfix)
- [yamllint](https://github.com/adrienverge/yamllint)

yamllint is the one that provided me fixes for everything @ingydotnet reported me in a PR.

ccoVeille:
- **Add CI for linting YAML files**
- **Fix go.yaml linting issues against yamllint rules**

Fixes #84

- #84

The rules come from various feedback made by @ingydotnet https://github.com/yaml/go-yaml/pull/83#pullrequestreview-3144726521

- https://github.com/yaml/go-yaml/pull/83#pullrequestreview-3144726521

### DIFF
--- a/.github/workflows/files.yaml
+++ b/.github/workflows/files.yaml
@@ -1,0 +1,19 @@
+name: Files Lint
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+permissions:
+  contents: read  # for actions/checkout to fetch code
+
+jobs:
+  yamllint:
+    runs-on: ubuntu-latest
+    steps:
+    - name: checkout-action
+      uses: actions/checkout@v5
+
+    - name: yamllint
+      uses: ibiqlik/action-yamllint@v3
--- a/.yamllint.yaml
+++ b/.yamllint.yaml
@@ -1,0 +1,19 @@
+extends: default
+
+ignore: /**/testdata/  # ignore testdata files that are not relevant for linting
+
+rules:
+  brackets:
+    min-spaces-inside: 0
+    max-spaces-inside: 1  # allow us to use space in brackets
+  document-start:
+    present: false
+  indentation:
+    spaces: 2
+    indent-sequences: false  # make sure there is no indentation for sequences
+  truthy:
+    check-keys: false  # there is a problem with the "on" key in GitHub Actions
+  quoted-strings:
+    check-keys: true  # by default, only values are checked
+    quote-type: single
+    required: only-when-needed
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -30,7 +30,7 @@ jobs:
         go-versions:
           [ '1.18', '1.19', '1.20', '1.21', '1.22', '1.23', '1.24', '1.25' ]
     runs-on: ubuntu-latest
-    continue-on-error: true # make sure to run all versions, even if one fails
+    continue-on-error: true  # make sure to run all versions, even if one fails
     steps:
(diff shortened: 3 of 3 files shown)


ccoVeille:
> Please prefer `.yaml` over `.yml` for file extensions where possible.

I changed it

.github/workflows/files.yaml
```diff
@@ -0,0 +1,19 @@
+name: Files Lint
+on:
+  push:
+    branches: [ main ]
```
ingydotnet, line 4: I think you can just use `branches: main` here.

I _think_ most GHA values that take a sequence can take a single value instead.

.yamllint.yml
```diff
@@ -0,0 +1,18 @@
+extends: default
+ignore:
```
ingydotnet: Try `ignore: ./**/testdata/`.

ingydotnet: or if that doesn't work, try `ignore: [ ./**/testdata/ ]`

.github/workflows/files.yaml
ccoVeille: Image

.yamllint.yml
ccoVeille: this worked

```yaml
ignore: /**/testdata/
```

while `./` variants didn't

.yamllint.yaml
```diff
@@ -0,0 +1,17 @@
+extends: default
```
ingydotnet, line 1: Add blank lines between top level pairs

.yamllint.yaml
```diff
@@ -0,0 +1,17 @@
...
+ignore: /**/testdata/  # ignore testdata files that are not relevant for linting
```
ingydotnet, line 3: Did that ignore line actually work?! Cool!

.yamllint.yaml
```diff
@@ -0,0 +1,17 @@
...
+  document-start:
+    present: false
+  indentation:
+    spaces: 2
+    indent-sequences: false  # make sure there is no indentation for sequences
+  truthy:
+    check-keys: false  # there is a problem with the "on" key in GitHub Actions
+  quoted-strings:
+    check-keys: true  # by default, only values are checked
+    quote-type: single
```
ingydotnet, line 18: I can't believe this config works the way I prefer!! Awesome!

ingydotnet (changes requested):

.yamllint.yaml
ccoVeille: Yes, it's pretty close.

2 things are not supported by the yaml lint and its config

- Don't over-indent block sequences as mapping values.
- Let's put blank lines before top level pairs.

.yamllint.yaml
ccoVeille: Yes.

We will have to use an array again if we want to have more exclusions

ccoVeille (commented):
please review and let me know

.yamllint.yaml
ingydotnet: > * Don't over-indent block sequences as mapping values.

I thought that's what `indent-sequences: false ` does

> * Let's put blank lines before top level pairs.

Makes sense. Not really a best practice. I just think it's nice here.

ingydotnet (approved):

dims (approved):

PR #92, opened 2025-08-23
Add a GitHub Action to check commit messages

### LINKED ISSUE #81: CI: lint commit messages
As per @ingydotnet request

> commit messages begin with capital letter and <= 50 chars and no . at end

Let's implement this.

Note: we might have to handle dependabot specifically. Either with tuning its commit message via .GitHub/dependabot.yml, or with an exception in the linting rule.

I hope to find existing GitHub Action for this

ccoVeille:
The following rules were implemented:

The subject should not use the conventional commit format (e.g., feat, fix, etc.).                                                                                                         
The subject should not start with square brackets.                                                                                                                                         
The subject should start with a capital letter.                                                                                                                                            
The subject should not end with a period.                                                                                                                                                  
The subject should not contain consecutive spaces.                                                                                                                                         
The subject should not have trailing spaces.                                                                                                                                               
The subject should be longer than 20 characters.                                                                                                                                           
The subject should be shorter than 50 characters.                                                                                                                                          
The subject and body should be separated by a single blank line.                                                                                                                           
Each line in the body should not have trailing spaces.                                                                                                                                     

Fixes #81 
Closes and replace #83


ccoVeille:
Here is what the script is able to report

image

.github/workflows/check-commit-message.yaml
```diff
@@ -0,0 +1,22 @@
...
+  check-commit-message:
+    name: Check Commit Message
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.user.login != 'dependabot[bot]'
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v5
+      with:
+        fetch-depth: 0  # Fetch all history to ensure all SHAs are available
+    - name: Check PR Commits
```
ingydotnet, line 20: Better as:
```
      run: 
        ./tooling/check-commit-messages.sh
        ${{ github.event.pull_request.base.sha }}..${{
               github.event.pull_request.head.sha }}
```
I think that should work fine but try variations if you want.

tooling/check-commit-messages.sh
```diff
@@ -0,0 +1,151 @@
+#!/bin/bash
```
ingydotnet: This is fairly well done but there's a lot I'd change. I'll push a series of commits to show how I'd do this.

ingydotnet (changes requested):

.github/workflows/check-commit-message.yaml
ingydotnet: I pushed this change.

ingydotnet:
@ccoVeille I'm done refactoring the Bash. 

I did it as a series of small commits but you can squash into 1 commit when we are done discussing them.

I hope you find this useful.

Let me know if you have questions or disagreements.

There's a lot to learn. I even learned new stuff doing this.

ingydotnet:
Also test things out. I'm not sure if I broke the logic.

ingydotnet:
> Also test things out. I'm not sure if I broke the logic.

I tested some bad commits and seemed to work.

util/check-commit-messages.sh
```diff
@@ -0,0 +1,128 @@
...
+
+# shellcheck disable=1091
+source "$(dirname "${BASH_SOURCE[0]}")"/common.bash || exit
+
+usage() (
+	cat <<-...
+	Usage: $0 <commit-range>|<file>
+	  <commit-range>: A range of commits in the form hash..hash
+	  <file>: A file containing a list of commit hashes, one per line"
+	...
```
ccoVeille, line 13: I'm curious: Why using cat instead of echo here?

util/check-commit-messages.sh
```diff
@@ -0,0 +1,128 @@
...
+	elif [[ $length -gt 50 ]]; then
+		errors+=("subject should be shorter than 50 characters (current: $length) on line 1")
+	fi
+
+	if [[ ${#errors[*]} -gt 0 ]]; then
+		lines_with_errors[1]=true
+		last_line_with_error=1
+	fi
+
+	if [[ $(sed -n '2p' <<<"$message") ]]; then
```
ccoVeille, line 83: I should use <<< indirection instead of using `echo "whatever" | sed` thanks

util/common.bash
ccoVeille, line 1: I tend to do optimize early. So here, I wouldn't add a another file to import before we get another Bash script in the same folder.

But I can live with this, anyway

util/common.bash
```diff
@@ -0,0 +1,33 @@
...
+shopt -s compat43 2>/dev/null ||
+	{ echo "Bash 4.4 required"; exit 1; }
+
+set -euo pipefail
+shopt -s inherit_errexit
+
+# ANSI color code vars:
+R="\033[31m"
+Y="\033[33m"
+Z="\033[0m"
```
ccoVeille: I tend to follow Go specs about variable names

> The basic rule: the further from its declaration that a name is used, the more descriptive the name must be. For a method receiver, one or two letters is sufficient. Common variables such as loop indices and readers can be a single letter (i, r). More unusual things and global variables need more descriptive names.
Source: https://go.dev/wiki/CodeReviewComments#variable-names

So I prefer color_red color_yellow color_reset

util/check-commit-messages.sh
```diff
@@ -0,0 +1,128 @@
...
+		# read the message and add the line number in front of each line, and use
+		# warn_color to display a line with an error based on line_with_errors
+		i=0
+		C=$Y
+		while IFS= read -r line && [[ $((++i)) -lt $last_line_with_error ]]; do
+			(
+				[[ ${lines_with_errors[$i]:-} ]] || C=''
+				echo -e "${C}Line $i: $line$Z"
+			)
+		done <<<"$message"
```
ccoVeille: Based on the error I saw in the GitHub Action log. I feel like this is no longer working.

https://github.com/yaml/go-yaml/actions/runs/17192738057/job/48770576703#step:3:1

util/check-commit-messages.sh
```diff
@@ -0,0 +1,128 @@
+#!/usr/bin/env bash
+
+# shellcheck disable=1091
```
ccoVeille, line 4: Here 1091 is a magical number.

I would explain what rule you are disabling and why

util/check-commit-messages.sh
ccoVeille: It was reported like this before you refactored

https://github.com/yaml/go-yaml/actions/runs/17191391015/job/48767481778#step:3:1

I wonder if the issue is not with the stderr used for errors that GHA might not support.

One solution could be simply to add a 2>&1 in the files.yaml when calling the script.

But it could be something else

ccoVeille: I also found this

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-an-error-message

That could be interesting. But we can consider this later

util/check-commit-messages.sh
ingydotnet: The point is to use a heredoc which is the cleanest way to do a multiline. `<<-` supports indenting the whole thing, but only using tabs. 
I've never used tabs to indent bash, but now I like it, mostly due to that feature.

You don't echo a heredoc, you cat them.

util/check-commit-messages.sh
ingydotnet: I think it's cleaner. I try not to use echo cat and pipes where there is better bash idioms.
Like `$(< file.txt)` instead of `$(cat file.txt)`

util/common.bash
ingydotnet: Fair point. I was just trying to show you things I would do. I do think we'll have more scripts.
Though I like scripting in yamlscript a lot these days :)

util/common.bash
ingydotnet: I'd rather have those longer names as functions.
R Y and Z are vars I've used a lot in the past.

util/check-commit-messages.sh
ingydotnet: you can probably change that line to
```
# shellcheck disable=1091 - https://www.shellcheck.net/wiki/SC1091
```

util/check-commit-messages.sh
ingydotnet: I'll look into it.

util/check-commit-messages.sh
ingydotnet: Looks like it would have to be 
```
# shellcheck disable=1091 # https://www.shellcheck.net/wiki/SC1091
```
else shellcheck barfs on it.

I'd rather leave it out as I'll add a shellcheck test soon that deals with 1091 itself so we don't need disables in scripts.

util/common.bash
ingydotnet: Changed to RED and RESET,

util/check-commit-messages.sh
ingydotnet: Fixed!

See https://github.com/yaml/go-yaml/actions/runs/17214596559/job/48834932487?pr=92 for a full example report.

ingydotnet:
@ccoVeille I think I've fixed everything now.

ingydotnet:
> image

Do we want this style output with the explanations interleaved, or do you like what we have in https://github.com/yaml/go-yaml/actions/runs/17214596559/job/48834932487?pr=92 ?

util/check-commit-messages.sh
ccoVeille: What is the status now?

util/common.bash
```diff
@@ -0,0 +1,32 @@
+#!/bash
```
ccoVeille: This script is included, OK.

But this shebang looks suspicious to me

ccoVeille:
> Do we want this style output with the explanations interleaved, or do you like what we have in https://github.com/yaml/go-yaml/actions/runs/17214596559/job/48834932487?pr=92 ?

Let's merge and we might iterate later

ccoVeille:
I'm still reviewing BTW.

Sorry for the confusion by saying let's merge

util/common.bash
```diff
@@ -0,0 +1,32 @@
...
+
+# Strict settings:
+# -u works poorly with arrays until Bash 4.4
+shopt -s compat43 2>/dev/null ||
+	{ echo "Bash 4.4 required"; exit 1; }
```
ccoVeille: This file contains nothing about arrays.

I would move this in the check-commit-messages.sh

I'm a bit afraid of someone removing it from here, because a script cannot work with Bash < 4.4 while he needs to import some other helpers from this file. Then it would break check-commit-messages.sh

Another way would be to add a function checkBash44 here, that encapsulates this.
And check-commit-messages.sh would call it, and any future script that may requires it.

ccoVeille: I tend to also display the current Bash version in such failure.

ingydotnet: I copied that comment from another place.

I'll make some quick adjustments.

ingydotnet:
@ccoVeille https://github.com/yaml/go-yaml/commit/734b011be286c717d7015df87e6a3f89fbef4a76

ccoVeille (commented):
LGTM 👍

ingydotnet (approved):

PR #80, opened 2025-08-21
Renamed types to match Go naming conventions

inteon:
The types were copied from the c project this implementation was based on.
However, in Go the convention is to use [MixedCaps names](https://go.dev/doc/effective_go#mixed-caps).
I hope that this rename will make the code a bit less scary for Go developers.

I renamed the types using this script:
```bash
gofmt -r 'yaml_emitter_t -> yamlEmitter' -w ./

gofmt -r 'yaml_version_directive_t -> yamlVersionDirective' -w ./
gofmt -r 'yaml_tag_directive_t -> yamlTagDirective' -w ./
gofmt -r 'yaml_mark_t -> yamlMark' -w ./
gofmt -r 'yaml_token_t -> yamlToken' -w ./
gofmt -r 'yaml_event_t -> yamlEvent' -w ./
gofmt -r 'yaml_node_pair_t -> yamlNodePair' -w ./
gofmt -r 'yaml_node_t -> yamlNode' -w ./
gofmt -r 'yaml_document_t -> yamlDocument' -w ./
gofmt -r 'yaml_simple_key_t -> yamlSimpleKey' -w ./
gofmt -r 'yaml_alias_data_t -> yamlAliasData' -w ./
gofmt -r 'yaml_parser_t -> yamlParser' -w ./
gofmt -r 'yaml_comment_t -> yamlComment' -w ./

gofmt -r 'yaml_read_handler_t -> yamlReadHandler' -w ./
gofmt -r 'yaml_write_handler_t -> yamlWriteHandler' -w ./

gofmt -r 'yaml_encoding_t -> yamlEncoding' -w ./
gofmt -r 'yaml_break_t -> yamlBreak' -w ./
gofmt -r 'yaml_error_type_t -> yamlErrorType' -w ./
gofmt -r 'yaml_style_t -> yamlStyle' -w ./
gofmt -r 'yaml_token_type_t -> yamlTokenType' -w ./
gofmt -r 'yaml_event_type_t -> yamlEventType' -w ./
gofmt -r 'yaml_node_type_t -> yamlNodeType' -w ./
gofmt -r 'yaml_node_item_t -> yamlNodeItem' -w ./
gofmt -r 'yaml_parser_state_t -> yamlParserState' -w ./
gofmt -r 'yaml_emitter_state_t -> yamlEmitterState' -w ./

gofmt -r 'yaml_scalar_style_t -> yamlScalarStyle' -w ./
[...]

PR #93, opened 2025-08-29
Add typos GitHub action

ccoVeille:
This will report typos in the codebase and documentation.

### DIFF
--- a/yamlh.go
+++ b/yamlh.go
@@ -771,7 +771,7 @@ type yamlEmitter struct {
 	scalar_data struct {
 		value                 []byte          // The scalar value.
 		multiline             bool            // Does the scalar contain line breaks?
-		flow_plain_allowed    bool            // Can the scalar be expessed in the flow plain style?
+		flow_plain_allowed    bool            // Can the scalar be expressed in the flow plain style?
 		block_plain_allowed   bool            // Can the scalar be expressed in the block plain style?
 		single_quoted_allowed bool            // Can the scalar be expressed in the single quoted style?
 		block_allowed         bool            // Can the scalar be expressed in the literal or folded styles?


ccoVeille:
It looks like to be a conflict was badly resolved with #80 and 64132a759f891171537e6913f72ff857e08bba72

- #80 

https://github.com/yaml/go-yaml/commit/64132a759f891171537e6913f72ff857e08bba72#diff-ff8aaff94a2db49556e62ef7a6566f79fc7dfe9a40a46de407cd9290e1d83e63R35-L781

ccoVeille:
Not a real issue, but I will have to fix it

EDIT: done :heavy_check_mark:

.typos.toml
```diff
@@ -0,0 +1,22 @@
+# this is the configuration file of typos (spell checker)
+# https://github.com/crate-ci/typos
+
+[files]
+# excluded file
+extend-exclude = [
+	"go.sum", "go.mod", # go.sum and go.mod contains URLs with hash, they can provide false positive
+	"yts/testdata",    # third-party test data
```
ingydotnet: maybe line up this comment with the one above it?

.typos.toml
```diff
@@ -0,0 +1,22 @@
...
+	]
```
ingydotnet: why is this indented and the one below not?

.typos.toml
```diff
@@ -0,0 +1,22 @@
...
+
+# setting for Go files configuration
+[type.go]
+extend-ignore-re = [
+	'ba-dum-tss\W+',
```
ingydotnet: Explain these.

.typos.toml
```diff
@@ -0,0 +1,22 @@
+# this is the configuration file of typos (spell checker)
+# https://github.com/crate-ci/typos
+
+[files]
+# excluded file
+extend-exclude = [
+	"go.sum", "go.mod", # go.sum and go.mod contains URLs with hash, they can provide false positive
```
ingydotnet: hashes are all same length so we could ignore them with regex, without ignoring the files.

ingydotnet (changes requested):

.typos.toml
ccoVeille: Indeed. I will give it a try.

For records, there is a pending discussion

- https://github.com/crate-ci/typos/issues/1363

.typos.toml
ccoVeille: I removed one line, so it's OK now

.typos.toml
ccoVeille: done

.typos.toml
ccoVeille: It should be better, let me know if I didn't understand what you meant

.typos.toml
ccoVeille: in fact, there is no problem with go.mod and go.sum, so I removed the exceptions.

PR #95, opened 2025-08-30
Renamed functions to match Go naming conventions

inteon:
Similar to https://github.com/yaml/go-yaml/pull/80, trying to make the code a bit less scary for Go developers.

The PR includes the following type of diffs:
```diff
-func yaml_insert_token(parser *yamlParser, pos int, token *yamlToken) {
+func (parser *yamlParser) insertToken(pos int, token *yamlToken) {
```
^^ also dropped the `yaml_` prefix in the name

```diff
-func yaml_parser_initialize(parser *yamlParser) bool {
+func (parser *yamlParser) initialize() bool {
```
^^ also dropped the `yaml_parser_` prefix in the name

```diff
-func yaml_string_read_handler(parser *yamlParser, buffer []byte) (n int, err error) {
+func yamlStringReadHandler(parser *yamlParser, buffer []byte) (n int, err error) {
```

```diff
-func flush(emitter *yamlEmitter) bool {
+func (emitter *yamlEmitter) flushIfNeeded() bool {
```
^^ renamed the function due to a naming conflict

### DIFF
--- a/apic.go
+++ b/apic.go
@@ -26,7 +26,7 @@ import (
 	"io"
 )
 
-func yaml_insert_token(parser *yamlParser, pos int, token *yamlToken) {
+func (parser *yamlParser) insertToken(pos int, token *yamlToken) {
 	//fmt.Println("yaml_insert_token", "pos:", pos, "typ:", token.typ, "head:", parser.tokens_head, "len:", len(parser.tokens))
 
 	// Check if we can move the queue at the beginning of the buffer.
@@ -46,7 +46,7 @@ func yaml_insert_token(parser *yamlParser, pos int, token *yamlToken) {
 }
 
 // Create a new parser object.
-func yaml_parser_initialize(parser *yamlParser) bool {
+func (parser *yamlParser) initialize() bool {
 	*parser = yamlParser{
 		raw_buffer: make([]byte, 0, input_raw_buffer_size),
 		buffer:     make([]byte, 0, input_buffer_size),
@@ -55,12 +55,12 @@ func yaml_parser_initialize(parser *yamlParser) bool {
 }
 
 // Destroy a parser object.
-func yaml_parser_delete(parser *yamlParser) {
+func (parser *yamlParser) delete() {
 	*parser = yamlParser{}
 }
 
 // String read handler.
-func yaml_string_read_handler(parser *yamlParser, buffer []byte) (n int, err error) {
+func yamlStringReadHandler(parser *yamlParser, buffer []byte) (n int, err error) {
 	if parser.input_pos == len(parser.input) {
 		return 0, io.EOF
 	}
@@ -70,31 +70,31 @@ func yaml_string_read_handler(parser *yamlParser, buffer []byte) (n int, err err
 }
 
 // Reader read handler.
-func yaml_reader_read_handler(parser *yamlParser, buffer []byte) (n int, err error) {
+func yamlReaderReadHandler(parser *yamlParser, buffer []byte) (n int, err error) {
 	return parser.input_reader.Read(buffer)
 }
 
 // Set a string input.
-func yaml_parser_set_input_string(parser *yamlParser, input []byte) {
+func (parser *yamlParser) setInputString(input []byte) {
 	if parser.read_handler != nil {
 		panic("must set the input source only once")
 	}
-	parser.read_handler = yaml_string_read_handler
+	parser.read_handler = yamlStringReadHandler
 	parser.input = input
 	parser.input_pos = 0
 }
 
 // Set a file input.
-func yaml_parser_set_input_reader(parser *yamlParser, r io.Reader) {
+func (parser *yamlParser) setInputReader(r io.Reader) {
 	if parser.read_handler != nil {
 		panic("must set the input source only once")
 	}
-	parser.read_handler = yaml_reader_read_handler
+	parser.read_handler = yamlReaderReadHandler
 	parser.input_reader = r
 }
 
 // Set the source encoding.
-func yaml_parser_set_encoding(parser *yamlParser, encoding yamlEncoding) {
+func (parser *yamlParser) setEncoding(encoding yamlEncoding) {
 	if parser.encoding != yaml_ANY_ENCODING {
 		panic("must set the encoding only once")
 	}
@@ -102,7 +102,7 @@ func yaml_parser_set_encoding(parser *yamlParser, encoding yamlEncoding) {
 }
 
 // Create a new emitter object.
-func yaml_emitter_initialize(emitter *yamlEmitter) {
+func (emitter *yamlEmitter) initialize() {
 	*emitter = yamlEmitter{
 		buffer:     make([]byte, output_buffer_size),
 		raw_buffer: make([]byte, 0, output_raw_buffer_size),
@@ -113,43 +113,43 @@ func yaml_emitter_initialize(emitter *yamlEmitter) {
 }
 
 // Destroy an emitter object.
-func yaml_emitter_delete(emitter *yamlEmitter) {
+func (emitter *yamlEmitter) delete() {
 	*emitter = yamlEmitter{}
 }
 
 // String write handler.
-func yaml_string_write_handler(emitter *yamlEmitter, buffer []byte) error {
+func yamlStringWriteHandler(emitter *yamlEmitter, buffer []byte) error {
 	*emitter.output_buffer = append(*emitter.output_buffer, buffer...)
 	return nil
 }
 
 // yaml_writer_write_handler uses emitter.output_writer to write the
 // emitted text.
-func yaml_writer_write_handler(emitter *yamlEmitter, buffer []byte) error {
+func yamlWriterWriteHandler(emitter *yamlEmitter, buffer []byte) error {
 	_, err := emitter.output_writer.Write(buffer)
 	return err
 }
 
 // Set a string output.
-func yaml_emitter_set_output_string(emitter *yamlEmitter, output_buffer *[]byte) {
+func (emitter *yamlEmitter) setOutputString(output_buffer *[]byte) {
 	if emitter.write_handler != nil {
 		panic("must set the output target only once")
 	}
-	emitter.write_handler = yaml_string_write_handler
+	emitter.write_handler = yamlStringWriteHandler
 	emitter.output_buffer = output_buffer
 }
 
 // Set a file output.
-func yaml_emitter_set_output_writer(emitter *yamlEmitter, w io.Writer) {
+func (emitter *yamlEmitter) setOutputWriter(w io.Writer) {
 	if emitter.write_handler != nil {
 		panic("must set the output target only once")
 	}
-	emitter.write_handler = yaml_writer_write_handler
+	emitter.write_handler = yamlWriterWriteHandler
 	emitter.output_writer = w
 }
 
 // Set the output encoding.
-func yaml_emitter_set_encoding(emitter *yamlEmitter, encoding yamlEncoding) {
+func (emitter *yamlEmitter) setEncoding(encoding yamlEncoding) {
 	if emitter.encoding != yaml_ANY_ENCODING {
 		panic("must set the output encoding only once")
 	}
@@ -157,12 +157,12 @@ func yaml_emitter_set_encoding(emitter *yamlEmitter, encoding yamlEncoding) {
 }
 
 // Set the canonical output style.
-func yaml_emitter_set_canonical(emitter *yamlEmitter, canonical bool) {
+func (emitter *yamlEmitter) setCanonical(canonical bool) {
 	emitter.canonical = canonical
 }
 
 // Set the indentation increment.
-func yaml_emitter_set_indent(emitter *yamlEmitter, indent int) {
+func (emitter *yamlEmitter) setIndent(indent int) {
 	if indent < 2 || indent > 9 {
 		indent = 2
 	}
@@ -170,7 +170,7 @@ func yaml_emitter_set_indent(emitter *yamlEmitter, indent int) {
 }
 
 // Set the preferred line width.
-func yaml_emitter_set_width(emitter *yamlEmitter, width int) {
+func (emitter *yamlEmitter) setWidth(width int) {
 	if width < 0 {
 		width = -1
 	}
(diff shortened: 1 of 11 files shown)


apic.go
```diff
@@ -55,12 +55,12 @@ func yaml_parser_initialize(parser *yamlParser) bool {
...
 
 // Destroy a parser object.
-func yaml_parser_delete(parser *yamlParser) {
+func (parser *yamlParser) delete() {
 	*parser = yamlParser{}
 }
 
 // String read handler.
-func yaml_string_read_handler(parser *yamlParser, buffer []byte) (n int, err error) {
+func yamlStringReadHandler(parser *yamlParser, buffer []byte) (n int, err error) {
```
ccoVeille, line 62: This name was strange and is still strange.
Wording thing is never fun.

But I feel like this method that accepts a yamlParser and a []byte should become a method of yamlParser accepting []byte as argument.

It could help to simplify the wording

So it would become readString

```suggestion
func (parser *yamlParser) readString(buffer []byte) (n int, err error) {
```

apic.go
```diff
@@ -70,39 +70,39 @@ func yaml_string_read_handler(parser *yamlParser, buffer []byte) (n int, err err
 }
 
 // Reader read handler.
-func yaml_reader_read_handler(parser *yamlParser, buffer []byte) (n int, err error) {
+func yamlReaderReadHandler(parser *yamlParser, buffer []byte) (n int, err error) {
```
ccoVeille, line 72: Same as

https://github.com/yaml/go-yaml/pull/95#discussion_r2311923108

It could become


```suggestion
func (parser *yamlParser) readReader(buffer []byte) (n int, err error) {
```

apic.go
```diff
@@ -113,77 +113,77 @@ func yaml_emitter_initialize(emitter *yamlEmitter) {
...
 
 // Destroy an emitter object.
-func yaml_emitter_delete(emitter *yamlEmitter) {
+func (emitter *yamlEmitter) delete() {
 	*emitter = yamlEmitter{}
 }
 
 // String write handler.
-func yaml_string_write_handler(emitter *yamlEmitter, buffer []byte) error {
+func yamlStringWriteHandler(emitter *yamlEmitter, buffer []byte) error {
```
ccoVeille, line 120: Same here being a method of emitter would be better

apic.go
```diff
@@ -113,77 +113,77 @@ func yaml_emitter_initialize(emitter *yamlEmitter) {
...
 	*emitter.output_buffer = append(*emitter.output_buffer, buffer...)
 	return nil
 }
 
 // yaml_writer_write_handler uses emitter.output_writer to write the
 // emitted text.
-func yaml_writer_write_handler(emitter *yamlEmitter, buffer []byte) error {
+func yamlWriterWriteHandler(emitter *yamlEmitter, buffer []byte) error {
```
ccoVeille, line 127: Same

apic.go
```diff
@@ -113,77 +113,77 @@ func yaml_emitter_initialize(emitter *yamlEmitter) {
...
 
 // Set if unescaped non-ASCII characters are allowed.
-func yaml_emitter_set_unicode(emitter *yamlEmitter, unicode bool) {
+func (emitter *yamlEmitter) setUnicode(unicode bool) {
 	emitter.unicode = unicode
 }
 
 // Set the preferred line break character.
-func yaml_emitter_set_break(emitter *yamlEmitter, line_break yamlBreak) {
+func (emitter *yamlEmitter) setBreak(line_break yamlBreak) {
```
ccoVeille: SetLineBreak would be clearer

decode.go
```diff
@@ -89,7 +89,7 @@ func (p *parser) expect(e yamlEventType) {
 		p.parser.problem = fmt.Sprintf("expected %s event but got %s", e, p.event.typ)
 		p.fail()
 	}
-	yaml_event_delete(&p.event)
+	yamlEventDelete(&p.event)
```
ccoVeille: Can't we make p.event.delete() ?

emitterc.go
```diff
@@ -47,8 +47,8 @@ func put(emitter *yamlEmitter, value byte) bool {
 }
 
 // Put a line break to the output buffer.
-func put_break(emitter *yamlEmitter) bool {
-	if emitter.buffer_pos+5 >= len(emitter.buffer) && !yaml_emitter_flush(emitter) {
+func (emitter *yamlEmitter) putBreak() bool {
```
ccoVeille: putLineBreak would be better

yamlprivateh.go
```diff
@@ -122,46 +122,46 @@ func is_printable(b []byte, i int) bool {
 }
 
 // Check if the character at the specified position is NUL.
-func is_z(b []byte, i int) bool {
+func isZ(b []byte, i int) bool {
```
ccoVeille: Use isNullChar because here the z is not clear

Or isZeroChar

yamlprivateh.go
```diff
@@ -122,46 +122,46 @@ func is_printable(b []byte, i int) bool {
...
 // Check if the character at the specified position is blank (space or tab).
-func is_blank(b []byte, i int) bool {
+func isBlank(b []byte, i int) bool {
 	//return is_space(b, i) || is_tab(b, i)
 	return b[i] == ' ' || b[i] == '\t'
 }
 
 // Check if the character at the specified position is a line break.
-func is_break(b []byte, i int) bool {
+func isBreak(b []byte, i int) bool {
```
ccoVeille: isLineBreak

yamlprivateh.go
ccoVeille: You also have to rename the is_z comment you can find in comment in this file

yamlprivateh.go
```diff
@@ -122,46 +122,46 @@ func is_printable(b []byte, i int) bool {
...
 }
 
-func is_crlf(b []byte, i int) bool {
+func isCRLF(b []byte, i int) bool {
 	return b[i] == '\r' && b[i+1] == '\n'
 }
 
 // Check if the character is a line break or NUL.
-func is_breakz(b []byte, i int) bool {
+func isBreakz(b []byte, i int) bool {
```
ccoVeille: isLineBreakOrZero

yamlprivateh.go
```diff
@@ -175,7 +175,7 @@ func is_breakz(b []byte, i int) bool {
 }
 
 // Check if the character is a line break, space, or NUL.
-func is_spacez(b []byte, i int) bool {
+func isSpacez(b []byte, i int) bool {
```
ccoVeille: isSpaceOrZero

And so on

ccoVeille: isSpaceOrZero

And so on

ccoVeille (commented):
I can see you had a lot of fun

apic.go
ccoVeille: Same yamlBreak type could become `yamlLineBreak` or simply `lineBreak`

apic.go
```diff
@@ -321,7 +318,7 @@ func yaml_alias_event_initialize(event *yamlEvent, anchor []byte) bool {
 }
 
 // Create SCALAR.
-func yaml_scalar_event_initialize(event *yamlEvent, anchor, tag, value []byte, plain_implicit, quoted_implicit bool, style yamlScalarStyle) bool {
+func yamlScalarEventInitialize(event *yamlEvent, anchor, tag, value []byte, plain_implicit, quoted_implicit bool, style yamlScalarStyle) bool {
```
ccoVeille: About being idiomatic in Go I would use


```suggestion
func newScalarEventInitialize(anchor, tag, value []byte, plain_implicit, quoted_implicit bool, style yamlScalarStyle) *yamlEvent  {
 	event = &yamlEvent{
 		typ:             yaml_SCALAR_EVENT,
```

And return only yamlEvent as there is need to always return true

Or having a method like this

```suggestion
func (event *yamlEvent) resetScalar(anchor, tag, value []byte, plain_implicit, quoted_implicit bool, style yamlScalarStyle) {
 	*event = yamlEvent{
 		typ:             yaml_SCALAR_EVENT,
```

But it doesn't sound logic maybe

See also

https://github.com/yaml/go-yaml/pull/95#discussion_r2311931792

decode.go
```diff
@@ -39,22 +39,22 @@ type parser struct {
 
 func newParser(b []byte) *parser {
 	p := parser{}
-	if !yaml_parser_initialize(&p.parser) {
+	if !(&p.parser).initialize() {
```
ccoVeille: In Go, there is no need to use the pointer

This would work

```suggestion
	if !p.parser.initialize() {
```

decode.go
```diff
@@ -102,7 +102,7 @@ func (p *parser) peek() yamlEventType {
 	// It's curious choice from the underlying API to generally return a
 	// positive result on success, but on this case return true in an error
 	// scenario. This was the source of bugs in the past (issue #666).
-	if !yaml_parser_parse(&p.parser, &p.event) || p.parser.error != yaml_NO_ERROR {
+	if !(&p.parser).parse(&p.event) || p.parser.error != yaml_NO_ERROR {
```
ccoVeille: This is not idiomatic

Here I feel like parse should return an error

encode.go
```diff
@@ -39,17 +39,17 @@ type encoder struct {
 
 func newEncoder() *encoder {
 	e := &encoder{}
-	yaml_emitter_initialize(&e.emitter)
-	yaml_emitter_set_output_string(&e.emitter, &e.out)
-	yaml_emitter_set_unicode(&e.emitter, true)
+	(&e.emitter).initialize()
+	(&e.emitter).setOutputString(&e.out)
+	(&e.emitter).setUnicode(true)
```
ccoVeille:
```suggestion
	e.emitter.initialize()
	e.emitter.setOutputString(&e.out)
	e.emitter.setUnicode(true)
```

encode.go
```diff
@@ -61,24 +61,24 @@ func (e *encoder) init() {
 		e.indent = 4
 	}
 	e.emitter.best_indent = e.indent
-	yaml_stream_start_event_initialize(&e.event, yaml_UTF8_ENCODING)
+	yamlStreamStartEventInitialize(&e.event, yaml_UTF8_ENCODING)
```
ccoVeille: See https://github.com/yaml/go-yaml/pull/95#discussion_r2311929936

I feel like code should be

```suggestion
	e.event = newStreamStartEvent(yaml_UTF8_ENCODING)
```

ccoVeille:
Thanks a lot for fighting this dragon 🥰🎉

decode.go
inteon: probably OOS for this PR?

apic.go
inteon: This function is actually used to overwrite the `read_handler` property in `setInputString`.

I think it will become too confusing if we make this a method of yamlParser.
We could also just move this function into the `setInputString` function, but that is probably OOS.

apic.go
inteon: see https://github.com/yaml/go-yaml/pull/95#discussion_r2311993268

apic.go
inteon: see https://github.com/yaml/go-yaml/pull/95#discussion_r2311993268

apic.go
inteon: see https://github.com/yaml/go-yaml/pull/95#discussion_r2311993268

decode.go
ccoVeille: Indeed

apic.go
ccoVeille: You are right

emitterc.go
```diff
@@ -104,24 +104,24 @@ func write(emitter *yamlEmitter, s []byte, i *int) bool {
...
+		if !emitter.write(s, &i) {
 			return false
 		}
 	}
 	return true
 }
 
 // Copy a line break character from a string into buffer.
-func write_break(emitter *yamlEmitter, s []byte, i *int) bool {
+func (emitter *yamlEmitter) writeBreak(s []byte, i *int) bool {
```
ccoVeille:
```suggestion
func (emitter *yamlEmitter) writeLineBreak(s []byte, i *int) bool {
```

yamlprivateh.go
```diff
@@ -104,12 +104,12 @@ func as_hex(b []byte, i int) int {
 }
 
 // Check if the character is ASCII.
-func is_ascii(b []byte, i int) bool {
+func isAscii(b []byte, i int) bool {
```
ccoVeille: Maybe this to respect the acronym

```suggestion
func isASCII(b []byte, i int) bool {
```

apic.go
```diff
@@ -70,40 +69,40 @@ func yaml_string_read_handler(parser *yamlParser, buffer []byte) (n int, err err
...
 
 // Set a file input.
-func yaml_parser_set_input_reader(parser *yamlParser, r io.Reader) {
+func (parser *yamlParser) setInputReader(r io.Reader) {
 	if parser.read_handler != nil {
 		panic("must set the input source only once")
 	}
-	parser.read_handler = yaml_reader_read_handler
+	parser.read_handler = yamlReaderReadHandler
 	parser.input_reader = r
```
ccoVeille, line 92: Any plan to rename them to


```suggestion
	parser.readHandler = yamlReaderReadHandler
 	parser.inputReader = r
```

There are many of these.

In this PR or another?

apic.go
```diff
@@ -382,7 +372,7 @@ func yaml_event_delete(event *yamlEvent) {
 // */
 //
 //YAML_DECLARE(int)
-//yaml_document_initialize(document *yaml_document_t,
+//yaml_document_(document *yaml_document_t,
```
ccoVeille: This one looks like the result of bad search replace

ccoVeille (commented):
The result is impressive it looks like Go code now. Thanks a lot.

A few suggestion, feel free to ignore

ccoVeille (approved):

apic.go
inteon: Indeed, this can be another PR.

apic.go
inteon: Fixed

ccoVeille (approved):

ccoVeille:
Thanks again

stefanprodan (approved):
LGTM

Awesome refactoring 🏅

decode.go
dolmen: This is out of scope.

yamlprivateh.go
dolmen: This is out of scope.

apic.go
dolmen: Changing function names is out of scope for this PR.
Please keep it boring.

PR #29, opened 2025-06-04
Add staticcheck to the Linter configuration

zostay:
This builds on the work in #26 and #27, so ensure those are merged first.

This is the first of these linter configs where the fixes might be controversial. It removes some unused code and applies De Morgan's law in a couple cases, removes some redundant parentheses, turns some if-else statements into switch statements (which is the preferred style in Go), changes one instance of `error.New(fmt.Sprintf(...))` to `fmt.Errorf(...)`, changes a use of `strings.Index` to `strings.Contains`, changing a use of `reflect.PtrTo` to the new `reflect.PointerTo` method, and things like that.

The controversial bit is all the cases where it suggests De Morgan's law be applied. I just added `//nolint:staticcheck` comments with explanations for the majority of these. Most of them are cases where the change would make no appreciable difference to readability or might actually make the expression less readable. At least, that was my judgement call on these.

I'd be happy to defer to the reviewer if the reviewer would prefer a different solution in any of these cases.

PR #106, opened 2025-09-08
Enable staticcheck in golangci and fix everything reported

ccoVeille:
- **Enable staticcheck in golangci-lint**
- **Fix code according to staticcheck**
- **Fix code according to staticcheck**
- **Remove golangci-lint v1 default exclusions**
- **Fix all remaining staticcheck errors**
- **Add missing go doc to asserters**

This PR supersedes and replaces #29 that was left unfinished and was partially addressing staticcheck issues.

Closes #29 

- #29

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -1515,12 +1515,12 @@ func TestObsoleteUnmarshalerTypeErrorProxying(t *testing.T) {
 	assert.ErrorMatches(t, expectedError, err)
 }
 
-var failingErr = errors.New("failingErr")
+var errFailing = errors.New("failingErr")
 
 type failingUnmarshaler struct{}
 
 func (ft *failingUnmarshaler) UnmarshalYAML(node *yaml.Node) error {
-	return failingErr
+	return errFailing
 }
 
 func TestUnmarshalerError(t *testing.T) {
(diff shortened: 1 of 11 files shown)


internal/testutil/assert/assert.go
```diff
@@ -73,14 +89,15 @@ func ErrorMatchesf(tb miniTB, pattern string, err error, msgFormat string, args
...
+// ErrorIs asserts that two errors are equal by using [errors.Is].
 func ErrorIs(tb miniTB, got, want error) {
 	tb.Helper()
 	if !errors.Is(got, want) {
 		tb.Fatalf("got %#v; want %#v", got, want)
 	}
 }
 
-// errorAsNoPanic calls errors.As, but catch possible panic and returns it as an error
+// errorAsNoPanic calls [errors.As], but catch possible panic and returns it as an error
```
ingydotnet, line 100: Explain the square brackets there.

decode_test.go
```diff
@@ -1502,12 +1502,12 @@ func TestObsoleteUnmarshalerTypeErrorProxying(t *testing.T) {
 	assert.ErrorMatches(t, expectedError, err)
 }
 
-var failingErr = errors.New("failingErr")
+var errFailing = errors.New("failingErr")
```
ingydotnet, line 1518: Why the rename?

internal/testutil/assert/assert.go
ccoVeille: The square brackets is the way to render HTML links in the documentation.

go doc automatically add links by resolving the links.

If you add `[whatever]` it would be rendered without a link, but when there is a function that is available it adds links.

More information here

https://tip.golang.org/doc/comment#doclinks

decode_test.go
ccoVeille: staticcheck reports it as it's a violation of the Go guidelines

> Error types end in "Error" and error variables start with "Err" or "err"
https://go.dev/wiki/Errors#naming

Issue #96, opened 2025-08-31
migration from gopkg.in/yaml feedback


ccoVeille:
> quick clarification on the maintained go-yaml module path. Running `go get go.yaml.in/yaml/v3@v3.0.1` fails because the module declares `module
gopkg.in/yaml.v3`.

_Originally posted by @ben-ranford in https://github.com/bmf-san/ggc/issues/149#issuecomment-3239903775_

The presence of old tags create confusion

None of them could be used as a Go module.

ccoVeille:
I feel like we should do something about it.

They are already cached in gomod proxies, so deleting the tags will have no effects

I would like to suggest using go mod retract

stefanprodan:
+1 for go mod retract

ingydotnet:
@ccoVeille @stefanprodan Can you be more specific about exactly what you mean by `go mod retract`?

I currently have no idea how to act on this.

ccoVeille:
A go mod retract command is necessary to invalidate the go.yaml.in/yaml/v3 module for versions v3.0.0 and v3.0.1. The module name was incorrect in these versions and was only fixed in v3.0.2. Using go mod retract will prevent users from accidentally depending on these broken versions. This is a crucial step to maintain the integrity of our module and avoid build failures for our users.

The idea is to do this in the latest version of the head branch

```
module go.yaml.in/yaml/v4

go 1.18

require (
	...
)

// These versions are retracted to prevent users from depending on them.
retract [v3.0.0, v3.0.1] // The module name was incorrect in v3.0.0 and v3.0.1 and was fixed in v3.0.2.
```

ingydotnet:
Got it. Can we get a PR for that?

PR #107, opened 2025-09-08
Retract v3 tags that cannot be installed

ccoVeille:
- **Retract v3 tags that cannot be installed**

Related to #96

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,3 +3,8 @@ module go.yaml.in/yaml/v3
 go 1.16
 
 require gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405
+
+// these tags come from gopkg.in/yaml.v3
+// they cannot be installed from go.yaml.in/yaml/v3 as it doesn't match
+// so they are invalid and are retracted.
+retract [v3.0.0, v3.0.1] // v3.0.2 is the first one with go.yaml.in/yaml/v3 module.


ccoVeille:
Please note we will have to create a v3.0.5 tag to make it available

ccoVeille:
> Should we delete the old tags after we merge this?

We should as I think they might create confusion, yes.

But go proxies used by go.pk.dev might have already cached these versions. They will stay here for ever. That's why retract exists. It tells go.pkg.dev to invalidate them.

Here is an example on a package I created.

https://pkg.go.dev/github.com/ccoveille/go-safecast?tab=versions

I had retracted the v1.0.0

Here is what it looks alike on go.pkg.dev

https://pkg.go.dev/github.com/ccoveille/go-safecast@v1.0.0

ingydotnet:
I'm not sure why the Analyze and CodeQL checks aren't running.

ccoVeille:
🤔

I'm unsure there is a need to retract them (at least the v3)

They are not listed here

https://pkg.go.dev/github.com/yaml/go-yaml?tab=versions

Abd they are not here

https://pkg.go.dev/go.yaml.in/yaml/v3?tab=versions

I cannot test easily as I don't want see them cached on the Go proxies.

I will have to play with GOPROXY env variable to check.

For now, deleting the invalid tags might be a better first move

ingydotnet:
I merge the v2 retractions and pushed a new `v.2.4.3` tag.

I'll hold off on this one for now.

@ccoVeille can you make a list of the v2 and v3 tags that we should delete from the repo?

go.mod
```diff
@@ -3,3 +3,8 @@ module go.yaml.in/yaml/v3
 go 1.16
 
 require gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405
+
+// these tags come from gopkg.in/yaml.v3
+// they cannot be installed from go.yaml.in/yaml/v3 as it doesn't match
+// so they are invalid and are retracted.
+retract [v3.0.0, v3.0.1] // v3.0.2 is the first one with go.yaml.in/yaml/v3 module.
```
ccoVeille: Note may have to retract the tags that will contain only the retract directive.

See 
https://github.com/yaml/go-yaml/pull/108#issuecomment-3277873461

ccoVeille:
> @ccoVeille can you make a list of the v2 and v3 tags that we should delete from the repo?

I move the discussion out of the PR to the following issue.

- https://github.com/yaml/go-yaml/issues/176

Let's continue there.

I don't abandon the current PR anyway.

I will try to work on this

ccoVeille:
I have force push my changes as a recent rebase caused some issues.

PR #108, opened 2025-09-08
Retract v2 tags that cannot be installed

ccoVeille:
Related to https://github.com/yaml/go-yaml/issues/96

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,3 +3,8 @@ module go.yaml.in/yaml/v2
 go 1.15
 
 require gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405
+
+// these tags come from gopkg.in/yaml.v2
+// they cannot be installed from go.yaml.in/yaml/v2 as it doesn't match
+// so they are invalid and are retracted.
+retract [v2.0.0, v2.4.0] // v2.4.1 is the first one with go.yaml.in/yaml/v2 module.


ccoVeille:
Please note we will have to create a v2.4.3 tags to make it available

go.mod
```diff
@@ -3,3 +3,8 @@ module go.yaml.in/yaml/v2
 go 1.15
```
SVilgelm, line 3: You, probably, need to change version of go as well, because the retract directive was added in v16: https://go.dev/ref/mod#go-mod-file-retract

ccoVeille: Good catch

It would require some tests maybe.

While what you said is right, I feel like.

If the lib is used with Go 1.15, the user won't get warned, you are right. But the go get will fail anyway because package name won't satistify its import path.

If the lib is used with Go 1.16+, the retract will be parsed and it will be OK.

Also, the retract thing is about informing about the status on pkg.go.dev

But good catch.

ccoVeille:
I feel like I should have used what is documented on Go spec and I didn't get until now.

https://go.dev/ref/mod#go-mod-file-retract

```
 version containing retractions may retract itself. If the highest release or pre-release version of a module retracts itself, the @latest query resolves to a lower version after retracted versions are excluded.

As an example, consider a case where the author of module example.com/m publishes version v1.0.0 accidentally. To prevent users from upgrading to v1.0.0, the author can add two retract directives to go.mod, then tag v1.0.1 with the retractions.

retract (
    v1.0.0 // Published accidentally.
    v1.0.1 // Contains retractions only.
)

When a user runs go get example.com/m@latest, the go command reads retractions from v1.0.1, which is now the highest version. Both v1.0.0 and v1.0.1 are retracted, so the go command will upgrade (or downgrade!) to the next highest version, perhaps v0.9.5.
```

I should also have retracted the v2.4.3, the next tag that would be created for retracting the old versions.

The issue I see now is that dependabot will suggest upgrading repository using go-yaml to v2.4.3 while there is no need.

Please note, it's done. It's OK. No real harm.
It's just it could have been done differently. Let's think about it for the v3 retractions

ccoVeille:
Please note that what we could do would be this:

- add a go mod retract directive for 2.4.3
- add a go mod retract directive for 2.4.4
- create the 2.4.4

But there is rush or need. I'm just describing how to do it, if we want to do it.

ccoVeille:
> The issue I see now is that dependabot will suggest upgrading repository using go-yaml to v2.4.3 while there is no need.

Here are the effect of what I described

https://github.com/search?q=go-yaml+v2.4.3&type=pullrequests

212 opened PR for now, PR that are not needed

image

ccoVeille:
Here is the result of the retract

Before

```console
$ GOPROXY=direct go list -m -versions -json
{
	"Path": "go.yaml.in/yaml/v2",
	"Versions": [
		"v2.0.0",
		"v2.1.0",
		"v2.1.1",
		"v2.2.0",
		"v2.2.1",
		"v2.2.2",
		"v2.2.3",
		"v2.2.4",
		"v2.2.5",
		"v2.2.6",
		"v2.2.7",
		"v2.2.8",
		"v2.3.0",
		"v2.4.0",
		"v2.4.1",
		"v2.4.2",
		"v2.4.3"
	],
	"Main": true,
	"Dir": "~/dev/perso/oss/go-yaml",
	"GoMod": "~/dev/perso/oss/go-yaml/go.mod",
	"GoVersion": "1.15"
}
```

After

```console
$ GOPROXY=direct go list -m -versions -json
{
	"Path": "go.yaml.in/yaml/v2",
	"Versions": [
		"v2.4.1",
		"v2.4.2",
		"v2.4.3"
	],
	"Main": true,
	"Dir": "~/dev/perso/oss/go-yaml",
	"GoMod": "~/dev/perso/oss/go-yaml/go.mod",
	"GoVersion": "1.15"
}
```

ccoVeille:
@ingydotnet I think the following tags could be deleted from the repository

		"v2.0.0",
		"v2.1.0",
		"v2.1.1",
		"v2.2.0",
		"v2.2.1",
		"v2.2.2",
		"v2.2.3",
		"v2.2.4",
		"v2.2.5",
		"v2.2.6",
		"v2.2.7",
		"v2.2.8",
		"v2.3.0",
		"v2.4.0",

@stefanprodan please confirn you agree

If so, then we may proceed to their deletion

PR #110, opened 2025-09-09
Add options system

HadrienPatte:
This PR is a proposal for a new options based configuration framework for yaml functions. One of the objectives is to bring new behaviors without breaking change by making the new behaviors gated by an option.

See individual commit messages:

---

Remove unused strict argument for unmarshal function

This change is safe to do as `unmarshal` is an internal function and the
`strict` argument is unused.

---

Add options arguments to public functions

Add a variadic `Options` argument to public functions:
* `Unmarshal`/`Marshal`
* `NewDecoder`/`NewEncoder`
* `Node.Decode`/`Node.Encode`

`Options` is a struct that will have unexported fields representing
various marshalling and unmarshalling parameters. Instances of `Options`
will be created via helper functions.

---

Adopt new options framework for indent and knownFields

This commit builds upon the previous one to adopt the new `Options`
framework for two existing parameters: `indent` and `knownFields` which
are currently controlled respectively through the `Encoder.SetIndent()`
and `Decoder.KnownFields()` methods.

With this commit, they can now additionally be set via `OptionIndent()`
and `OptionKnownFields()` which return an `Options` that can be passed to
any public method and will get propagated down to the relevant place.


options.go
ccoVeille, line 1: What do you guys think about the idea of having an option package?

There is no such a thing in json/v2 standard lib, bit I feel I would prefer to isolate them in a dedicated packages this way IDE autocompletion and public documentation on pkg.go.dev is not polluted with options

options.go
```diff
@@ -0,0 +1,72 @@
+package yaml
+
+// Options holds configuration options for YAML processing
+type Options struct {
```
ccoVeille: I would use an unexported struct.

ccoVeille: I would prefer 

type Option func(*config) 

With config being your Options struct

yaml.go
```diff
@@ -85,30 +85,33 @@ type Marshaler interface {
...
+		parser:  newParserFromReader(r),
+		options: options,
 	}
 }
 
 // KnownFields ensures that the keys in decoded mappings to
 // exist as fields in the struct being decoded into.
 func (dec *Decoder) KnownFields(enable bool) {
-	dec.knownFields = enable
+	dec.options.knownFields = &enable
```
ccoVeille: Here usually there should have a default configuration. With a loop on the variadic made of func(*config)

options.go
HadrienPatte: I updated the PR to move the option logic to a separate package and use option functions, but one issue I found is that with a separate options package, I don't think I can have `config` be an unexported struct

ingydotnet:
One thing I'd like is a really easy way to specify options in YAML file.

Possibly even having an env var that specifies such a file.

options.go
ccoVeille: Indeed.

config could be moved to an internal package, maybe.

So it would be exported but not not of the module

yaml.go
```diff
@@ -85,30 +87,33 @@ type Marshaler interface {
...
 //
 // The decoder introduces its own buffering and may read
 // data from r beyond the YAML values requested.
-func NewDecoder(r io.Reader) *Decoder {
+func NewDecoder(r io.Reader, opts ...option.Option) *Decoder {
+	config := option.NewConfig(opts...)
 	return &Decoder{
 		parser: newParserFromReader(r),
+		config: config,
 	}
```
ccoVeille, line 110: Maybe this


```suggestion
 	return &Decoder{
 		parser: newParserFromReader(r),
		config: option.NewConfig(opts...),
 	}
```

yaml.go
```diff
@@ -85,30 +87,33 @@ type Marshaler interface {
...
 }
 
 // KnownFields ensures that the keys in decoded mappings to
 // exist as fields in the struct being decoded into.
 func (dec *Decoder) KnownFields(enable bool) {
-	dec.knownFields = enable
+	dec.config.Apply(option.WithKnownFields(enable))
```
ccoVeille, line 116: You can't be sure the config field is set here.

You will have to init if missing. This should be done in a dedicated function of this struct that could be reused

ccoVeille (commented):
Some random feedback.

ingydotnet:
I fixed the commit message errors. Commit message subject lines need to be 20-50 chars long.

option/option.go
```diff
@@ -0,0 +1,61 @@
...
+// GetKnownFields returns the Config's knownFields if set or the default value
+func (c *Config) GetKnownFields() bool {
+	if c.knownFields != nil {
+		return *c.knownFields
+	}
+	return defaultKnownFields
+}
+
+// NewConfig creates a new Config with the provided options
+func NewConfig(opts ...Option) *Config {
```
inteon, line 48:
```suggestion
func newConfig(opts ...Option) *Config {
```

This function probably does not need to be public.

option/option.go
```diff
@@ -0,0 +1,61 @@
+package option
+
+// Config holds configuration options for YAML processing
+type Config struct {
```
inteon, line 4:
```suggestion
type config struct {
```

This type probably does not need to be public.

yaml.go
```diff
@@ -85,30 +87,33 @@ type Marshaler interface {
...
+	config := option.NewConfig(opts...)
+	return unmarshal(in, out, config)
 }
 
 // A Decoder reads and decodes YAML values from an input stream.
 type Decoder struct {
-	parser      *parser
-	knownFields bool
+	parser *parser
+	config *option.Config
```
inteon, line 98:
```suggestion
	config option.Config
```

I believe it will be more efficient & easier to keep this as a non-pointer.

HadrienPatte:
Closing in favor of #162

PR #112, opened 2025-09-10
Add new public token and parser access functions

ingydotnet:
These are needed by https://github.com/yaml/go-yaml-cli which is an essential tool for understanding and debugging go-yaml.

### DIFF
--- a/yaml.go
+++ b/yaml.go
@@ -899,3 +899,131 @@ func formatEvent(e *yamlEvent) string {
 	}
 	return b.String()
 }
+
+// Token represents a YAML token
+type Token struct {
+	Type      string
+	Value     string
+	Style     string
+	StartLine int
+	StartCol  int
+	EndLine   int
+	EndCol    int
+}
+
+// Convert a yamlScalarStyle to a string representation
+func scalarStyleToString(style yamlScalarStyle) string {
+	switch style {
+	case yaml_PLAIN_SCALAR_STYLE:
+		return "Plain"
+	case yaml_SINGLE_QUOTED_SCALAR_STYLE:
+		return "Single"
+	case yaml_DOUBLE_QUOTED_SCALAR_STYLE:
+		return "Double"
+	case yaml_LITERAL_SCALAR_STYLE:
+		return "Literal"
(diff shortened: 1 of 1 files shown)


yaml.go
```diff
@@ -899,3 +899,131 @@ func formatEvent(e *yamlEvent) string {
...
+		token.Type = "UNKNOWN"
+	}
+
+	return token, nil
+}
+
+// Close releases the parser resources
+func (p *Parser) Close() {
+	p.parser.delete()
+}
```
ccoVeille, line 1029: I would move all this to

internal/debug/ folder

Or any other names, but something under internal/

Then this folder will have headers like

```
package go.yaml.in/yaml/v4internal/debug
```

And then I would move go-yaml-cli files into this repository to `/cmd/go-yaml-cli/` folder. The files in this folder would have the `package main` that prevents code to be exported.

And then go-yaml and go-yaml-cli could use the code in go.yaml.in/yaml/v4internal/debug without having it exported

ccoVeille: Another option is to have

`cmd/go-yaml-cli/main.go` (and only this file)
`internal/debug/*.go`
`internal/cmd/go-yaml-cli/*.go`

The first one imports code from go.yaml.in/yaml/v4/internal/cmd/go-yaml-cli and from go.yaml.in/yaml/v4/internal/debug

And go.yaml.in/yaml/v4 code imports from go.yaml.in/yaml/v4/internal/debug package

yaml.go
```diff
@@ -899,3 +899,131 @@ func formatEvent(e *yamlEvent) string {
...
+	case yaml_FOLDED_SCALAR_STYLE:
+		return "Folded"
+	default:
+		return ""
+	}
+}
+
+// Parser provides access to the internal YAML parser
+type Parser struct {
+	parser yamlParser
```
ccoVeille, line 934: So here is what you need in the go-yaml-cli I think

It means the yamlaParser that is currently unexported needs to be moved to internal/ and make it exported (but then limited to the module only)

And so the yamlParser and all its method will need to be moved to internal/parser/

ingydotnet:
Replaced by #113

PR #115, opened 2025-09-11
Move libyaml to an internal package

ccoVeille:
- **Move the libyaml to an internal package**
- **Apply the Go minimal naming convention**

### DIFF
--- a/internal/libyaml/api.go
+++ b/internal/libyaml/api.go
@@ -45,7 +45,7 @@ func (parser *Parser) insertToken(pos int, token *Token) {
 	parser.tokens[parser.tokens_head+pos] = *token
 }
 
-// Create a new parser object.
+// NewParser creates a new parser object.
 func NewParser() Parser {
 	return Parser{
 		raw_buffer: make([]byte, 0, input_raw_buffer_size),
@@ -53,7 +53,7 @@ func NewParser() Parser {
 	}
 }
 
-// Destroy a parser object.
+// Delete a parser object.
 func (parser *Parser) Delete() {
 	*parser = Parser{}
 }
@@ -73,7 +73,7 @@ func yamlReaderReadHandler(parser *Parser, buffer []byte) (n int, err error) {
 	return parser.input_reader.Read(buffer)
 }
 
-// Set a string input.
+// SetInputString sets a string input.
 func (parser *Parser) SetInputString(input []byte) {
 	if parser.read_handler != nil {
 		panic("must set the input source only once")
@@ -83,7 +83,7 @@ func (parser *Parser) SetInputString(input []byte) {
 	parser.input_pos = 0
 }
 
-// Set a file input.
+// SetInputReader sets a file input.
 func (parser *Parser) SetInputReader(r io.Reader) {
 	if parser.read_handler != nil {
 		panic("must set the input source only once")
@@ -92,7 +92,7 @@ func (parser *Parser) SetInputReader(r io.Reader) {
 	parser.input_reader = r
 }
 
-// Set the source encoding.
+// SetEncoding sets the source encoding.
 func (parser *Parser) SetEncoding(encoding Encoding) {
 	if parser.encoding != ANY_ENCODING {
 		panic("must set the encoding only once")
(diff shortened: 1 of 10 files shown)


ccoVeille:
@HadrienPatte @ingydotnet please review

This PR prepares what is needed to split out the code, and introduce go-yaml-cli

internal/libyaml/api.go
HadrienPatte, line 1: small nit/question: this broad move could be the occasion to drop the `c`/`h` suffix from those file names, eg `parserc.go` -> `parser.go`, what do you think?

ccoVeille: I don't feel a need in this PR, but later maybe

ingydotnet: I think we should rename them in this PR.

ingydotnet:
What are the lock-in implications of all the exported identifiers that are under `internal/`?

ccoVeille:
> What are the lock-in implications of all the exported identifiers that are under `internal/`?

They become available as exported but only within the module.

They cannot be used out of the module, by someone importing the module.

Said otherwise, they are exported for us when we code, so we can use them in any of our package, but they are unexported and thus unavailable to anyone using go-yaml as a lib

internal is a special folder in Go. Its purpose is what I described.

internal/libyaml/yaml.go
```diff
@@ -0,0 +1,804 @@
...
+	Type EventType
+
+	// The start and end of the event.
+	StartMark, EndMark Mark
+
+	// The document encoding (for STREAM_START_EVENT).
+	encoding Encoding
+
+	// The version directive (for DOCUMENT_START_EVENT).
+	version_directive *VersionDirective
```
inteon, line 300: @ccoVeille the casing of this field is totally different from the casing of eg. the StartMark field.

internal/libyaml/yaml.go
```diff
@@ -0,0 +1,804 @@
...
+// [in]       size        The size of the buffer.
+// [out]      size_read   The actual number of bytes read from the source.
+//
+// On success, the handler should return 1.  If the handler failed,
+// the returned value should be 0. On EOF, the handler should set the
+// size_read to 0 and return 1.
+type ReadHandler func(parser *Parser, buffer []byte) (n int, err error)
+
+// This structure holds information about a potential simple key.
+type SimpleKey struct {
```
inteon, line 449: @ccoVeille I think we shouldn't make all structs public, only the structs that we use outside of this package.

internal/libyaml/yaml.go
ccoVeille: Yes, I was hesitant to update them all at once.

I asked to other maintainers and they suggested me to keep the ones we don't need to update in this PR unchanged.

This way the PR is clearer.

I plan to rename all remaining variable in a future PR

internal/libyaml/yaml.go
ccoVeille: You are right.

The risk is limited because everything is now in internal packages, so it's not exported outside the module.

I might need to double check, maybe the SimpleKey is now available via an exported field that become exported because of the changes.

I can fix that if needed.

Maybe it was a simple search and replace that went too far.

Thanks for catching this

ccoVeille:
@inteon @HadrienPatte 

These are the thread that remains to be addressed.

https://github.com/yaml/go-yaml/pull/115#discussion_r2342163809

https://github.com/yaml/go-yaml/pull/115#discussion_r2358815451

Thanks for your feedback.

For your information, I don't plan to work any further on this PR, until I have a GO from @ingydotnet that he is OK for the PR.

internal/libyaml/api.go
ccoVeille: I've done it

internal/libyaml/yaml.go
ccoVeille: I don't think I will fix this unless someone else requests for it.

yamlh.go
ccoVeille, line 1: There is a problem here with the way Git understand the changes.

I need to fix this as I want git to be able to follow the renaming, which is useful when using git log or git grep

ccoVeille: I tried, it cannot be done :smirk:

ingydotnet:
@ccoVeille I resolved the conflicts, but now we get new linting errors. Can you make the adjustments? I want to merge this one next....

ccoVeille:
> @ccoVeille I resolved the conflicts, but now we get new linting errors. Can you make the adjustments? I want to merge this one next....

I solved them.

PR #121, opened 2025-09-14
Adds make format target, closes #120

### LINKED ISSUE #120: gofmt or gofumpt?
The [CONTRIBUTING](https://github.com/yaml/go-yaml/blob/main/CONTRIBUTING.md?plain=1#L49) doc says to use `gofmt`. But [.golangci.yaml:](https://github.com/yaml/go-yaml/blob/dcff9c48c5df0c383e6e0ea68b3be147c905f6b3/.golangci.yaml#L39-L44) is configured to use `gofumpt`.

I'm happy to open a PR to align these, just not sure which one is preferred. Additionally, I typically leverage a Makefile target:

```make
.PHONY: fmt
fmt:
	go tool -modfile=go.tool.mod golines --base-formatter=gofumpt -w .
	go tool -modfile=go.tool.mod gofumpt -l -w -extra .
```

This uses `go tool` which is Go 1.24+ feature. The [proposal](https://go.googlesource.com/proposal/+/54d6775ff71ccbc00c276db2a4e4841d67011cf4/design/48429-go-tool-modules.md) and the [discussionl](https://github.com/golang/go/issues/48429) and a more digestible blog post [How to manage tool dependencies in go 1.24+](https://www.alexedwards.net/blog/how-to-manage-tool-dependencies-in-go-1.24-plus). The adoption of this pattern should definitely be in a unique PR (if its adoption exists at all). Which is just a long way of saying, in addition to aligning the documentation and tooling on `gofmt` or `gofumpt`, I'd love to add a Make target `make fmt` that does either of these things if you're open to it. Thanks!

esacteksab:
This adds a `make format` target. While the intent was to use `make fmt` this is already used by `.cache/makes/go.mk` which already has a `fmt` target. Using the `fmt` target, results in the following error:

```bash
GNUmakefile:44: *** target file 'fmt' has both : and :: entries.  Stop.
```

I can extend it with `fmt::` but it runs `go fmt`, then it runs `golangci-lint fmt ./...` and the `go fmt` felt unnecessary, so I chose `format` instead.

I also updated the `CONTRIBUTING.md`, changes include:

- Properly format Markdown
- Reorder things for a better flow
- Update references to use the new `make format` target
- Included additional documentation on `make test` target for testing locally
- Add a mention of how to lint locally with `golangci-lint run`

I also removed the `fmt` job from `.github/workflows/go.yaml`.


CONTRIBUTING.md
```diff
@@ -90,19 +84,19 @@ targets:
 - `make shell` opens a shell with the project's dependencies set up
 - `make shell GO-VERSION=1.23.4` opens a shell with a specific Go version
 - `make fmt` runs `go fmt`
+- `make format` runs `golangci-lint fmt ./...`
```
ccoVeille: See https://github.com/yaml/go-yaml/pull/121#discussion_r2347991106


```suggestion
- `make fmt` runs `golangci-lint fmt ./...`
```

GNUmakefile
```diff
@@ -40,6 +40,8 @@ include $(MAKES)/shell.mk
 v ?=
 count ?= 1
 
+format:
+	golangci-lint fmt ./...
```
ccoVeille:
```suggestion
fmt:
	golangci-lint fmt ./...
```

Then add `fmt` to GO-CMDS-SKIP variable defined line 27.

Then the go.mk file will skip it

https://github.com/makeplus/makes/blob/488bced7b1d50e3addc25826d933f79d94a43416/go.mk#L28

And yours would be accepted.

I'm unsure about the format to provide multiple commands to ignore. But I feel like it should be something like 

```
GO-CMDS-SKIP = test fmt
```

Maybe with quotes.

CONTRIBUTING.md
```diff
@@ -1,59 +1,43 @@
...
-1. Write clear commit messages
-1. Update tests and documentation
-1. Submit a pull request
-
-
 ### Coding Conventions
 
 - Follow standard Go coding conventions
-- Use `gofmt` to format your code
+- Use `make format` to format your code
```
ccoVeille: See https://github.com/yaml/go-yaml/pull/121#discussion_r2347991106


```suggestion
- Use `make fmt` to format your code
```

GNUmakefile
esacteksab: I wasn't aware of this functionality. Thank you!

esacteksab:
> The code changes are fine.
> 
> Please update the PR description that refers to previous commit and mention `make format` and why you had to add it.
> 
> It can be rewritten now you managed to use make fmt directly

Sorry for missing that. Updated.

CONTRIBUTING.md
```diff
@@ -1,59 +1,43 @@
-Contributing to go-yaml
-=======================
```
ingydotnet, line 2: Please revert.

.github/workflows/go.yaml
```diff
@@ -7,23 +7,6 @@ on:
     branches: [ main, v3, v2, v1 ]
 
 jobs:
-  fmt:
```
ingydotnet, line 10: Why are you removing this formatting check?

CONTRIBUTING.md
```diff
@@ -1,59 +1,43 @@
...
+# Contributing to go-yaml
 
 Thank you for your interest in contributing to go-yaml!
 
 This document provides guidelines and instructions for contributing to this
 project.
 
-
```
ingydotnet, line 9: Please keep 2 blank lines before headings.

.github/workflows/go.yaml
ccoVeille: It's my request

More information here

https://github.com/yaml/go-yaml/issues/120#issuecomment-3289683266

There is no need for such job, golangci-lint action is enough

ingydotnet:
Please don't make markdown style changes in this PR.
Let's just keep it to adding the `make` target and removing the unneeded CI check.

esacteksab:
> Please don't make markdown style changes in this PR. Let's just keep it to adding the `make` target and removing the unneeded CI check.

Understood. I reverted the style changes. I did keep the ordering of the sections as it felt like it flowed better, but if you'd rather I not make that change either, happy to undo that also. Thanks!

ccoVeille:
I fixed the conflicts, the PR is now OK, and ready to be reviewed and merged.

ingydotnet (approved):

ingydotnet:
I'm a bit disappointed with this.
The GNUmakefile is set up so that all targets always work, with no need to install anything (including go!).

```
$ make fmt
golangci-lint fmt ./...
bash: line 1: golangci-lint: command not found
```

I'll submit a PR but we reviewers (including me!) need to look for that.

ccoVeille:
> I'm a bit disappointed with this.
> The GNUmakefile is set up so that all targets always work, with no need to install anything (including go!).
> 
> ```
> $ make fmt
> golangci-lint fmt ./...
> bash: line 1: golangci-lint: command not found
> ```
> 
> I'll submit a PR but we reviewers (including me!) need to look for that.

#131 was opened

- #131

PR #123, opened 2025-09-18
Improve empty document handling

### LINKED ISSUE #122: Comments before document start get misplaced
When decoding
```yaml
# foo
---
key: value
```
to a `yaml.Node`, the result is a single document which contains the comment attached to the non-empty map.

I prepared a fix for this in go-yaml/yaml some time ago (it's already 4.5 years: https://github.com/go-yaml/yaml/pull/690). I can re-do the PR here if there's interest in contributions, or feel free to salvage that PR yourself if you want :)

felixfontein:
Rebased from https://github.com/go-yaml/yaml/pull/690.

~~Fixes #122.~~

Allows to correctly parse and serialize
```
# comment
```
without losing the comment (parsing), or getting some error (serialization).

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -160,6 +160,15 @@ func (p *parser) parse() *Node {
 		return p.document()
 	case libyaml.STREAM_END_EVENT:
 		// Happens when attempting to decode an empty buffer.
+		if len(p.event.HeadComment) > 0 {
+			// The buffer is empty, but there is a comment.
+			// Create a document node which contains this comment.
+			// Without this, the comment would get lost since
+			// Node.Decode() would simply return io.EOF.
+			n := p.node(DocumentNode, "", "", "")
+			p.event.HeadComment = nil
+			return n
+		}
 		return nil
 	case libyaml.TAIL_COMMENT_EVENT:
 		panic("internal error: unexpected tail comment event (please report)")
--- a/decode_test.go
+++ b/decode_test.go
@@ -1102,6 +1102,244 @@ func TestDecoder(t *testing.T) {
 	}
 }
 
+var decoderNodeTests = []struct {
+	data  string
+	nodes []*yaml.Node
+}{
+	{
+		"# foo\nkey: value\n",
+		[]*yaml.Node{
+			{
+				Kind:   yaml.DocumentNode,
+				Line:   2,
+				Column: 1,
+				Content: []*yaml.Node{{
+					Kind:   yaml.MappingNode,
+					Tag:    "!!map",
+					Line:   2,
+					Column: 1,
+					Content: []*yaml.Node{{
+						Kind:        yaml.ScalarNode,
+						Value:       "key",
+						Tag:         "!!str",
+						Line:        2,
+						Column:      1,
+						HeadComment: "# foo",
+					}, {
+						Kind:   yaml.ScalarNode,
+						Value:  "value",
+						Tag:    "!!str",
+						Line:   2,
+						Column: 6,
+					}},
+				}},
+			},
+		},
+	},
+	{
+		"# foo\n---\n# bar\n---\nkey: value\n",
+		[]*yaml.Node{
+			{
+				Kind:        yaml.DocumentNode,
+				Line:        2,
+				Column:      1,
+				HeadComment: "# foo",
+				FootComment: "# bar",
+				Content: []*yaml.Node{{
+					Kind:   yaml.ScalarNode,
+					Value:  "",
+					Tag:    "!!null",
+					Line:   4,
+					Column: 1,
+				}},
+			}, {
+				Kind:   yaml.DocumentNode,
+				Line:   4,
+				Column: 1,
+				Content: []*yaml.Node{{
+					Kind:   yaml.MappingNode,
+					Tag:    "!!map",
+					Line:   5,
+					Column: 1,
+					Content: []*yaml.Node{{
+						Kind:   yaml.ScalarNode,
+						Value:  "key",
+						Tag:    "!!str",
+						Line:   5,
+						Column: 1,
+					}, {
+						Kind:   yaml.ScalarNode,
+						Value:  "value",
+						Tag:    "!!str",
+						Line:   5,
+						Column: 6,
+					}},
+				}},
+			},
+		},
+	},
+	{
+		"# foo\n",
+		[]*yaml.Node{
+			{
+				Kind:        yaml.DocumentNode,
+				Line:        2,
+				Column:      1,
+				HeadComment: "# foo",
+			},
+		},
+	},
+	{
+		"# foo\n---\n",
+		[]*yaml.Node{
+			{
+				Kind:        yaml.DocumentNode,
+				Line:        2,
+				Column:      1,
+				HeadComment: "# foo",
+				Content: []*yaml.Node{{
+					Kind:   yaml.ScalarNode,
+					Tag:    "!!null",
+					Value:  "",
+					Line:   3,
+					Column: 1,
+				}},
+			},
+		},
+	},
+	{
+		"# foo\n---\nkey: value\n",
+		[]*yaml.Node{
+			{
+				Kind:        yaml.DocumentNode,
+				Line:        2,
+				Column:      1,
+				HeadComment: "# foo",
+				Content: []*yaml.Node{{
+					Kind:   yaml.MappingNode,
+					Tag:    "!!map",
+					Line:   3,
+					Column: 1,
+					Content: []*yaml.Node{{
+						Kind:   yaml.ScalarNode,
+						Value:  "key",
+						Tag:    "!!str",
+						Line:   3,
+						Column: 1,
+					}, {
+						Kind:   yaml.ScalarNode,
+						Value:  "value",
+						Tag:    "!!str",
+						Line:   3,
+						Column: 6,
+					}},
+				}},
+			},
+		},
+	},
+	{
+		"# foo\n---\n# bar\nkey: value\n",
+		[]*yaml.Node{
+			{
+				Kind:        yaml.DocumentNode,
+				Line:        2,
+				Column:      1,
+				HeadComment: "# foo",
+				Content: []*yaml.Node{{
+					Kind:   yaml.MappingNode,
+					Tag:    "!!map",
+					Line:   4,
+					Column: 1,
+					Content: []*yaml.Node{{
+						Kind:        yaml.ScalarNode,
+						Value:       "key",
+						Tag:         "!!str",
+						Line:        4,
+						Column:      1,
+						HeadComment: "# bar",
+					}, {
+						Kind:   yaml.ScalarNode,
+						Value:  "value",
+						Tag:    "!!str",
+						Line:   4,
+						Column: 6,
+					}},
+				}},
+			},
+		},
+	},
+	{
+		"key: value\n\n# foo\n---\nkey: value\n",
+		[]*yaml.Node{
+			{
+				Kind:        yaml.DocumentNode,
+				Line:        1,
+				Column:      1,
+				FootComment: "# foo",
+				Content: []*yaml.Node{{
+					Kind:   yaml.MappingNode,
+					Tag:    "!!map",
+					Line:   1,
+					Column: 1,
+					Content: []*yaml.Node{{
+						Kind:   yaml.ScalarNode,
+						Value:  "key",
+						Tag:    "!!str",
+						Line:   1,
+						Column: 1,
+					}, {
+						Kind:   yaml.ScalarNode,
+						Value:  "value",
+						Tag:    "!!str",
+						Line:   1,
+						Column: 6,
+					}},
+				}},
+			}, {
+				Kind:   yaml.DocumentNode,
+				Line:   4,
+				Column: 1,
+				Content: []*yaml.Node{{
+					Kind:   yaml.MappingNode,
+					Tag:    "!!map",
+					Line:   5,
+					Column: 1,
+					Content: []*yaml.Node{{
+						Kind:   yaml.ScalarNode,
+						Value:  "key",
+						Tag:    "!!str",
+						Line:   5,
+						Column: 1,
+					}, {
+						Kind:   yaml.ScalarNode,
+						Value:  "value",
+						Tag:    "!!str",
+						Line:   5,
+						Column: 6,
+					}},
+				}},
+			},
+		},
+	},
+}
+
+func TestDecoderNodes(t *testing.T) {
+	for i, item := range decoderNodeTests {
+		var nodes []*yaml.Node
+		d := yaml.NewDecoder(bytes.NewReader([]byte(item.data)))
+		for {
+			node := &yaml.Node{}
+			err := d.Decode(node)
+			if err == io.EOF {
+				break
+			}
+			assert.NoError(t, err)
+			nodes = append(nodes, node)
+		}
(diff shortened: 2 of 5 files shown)


decode_test.go
```diff
@@ -1067,6 +1067,164 @@ func TestDecoder(t *testing.T) {
...
+		d := yaml.NewDecoder(bytes.NewReader([]byte(item.data)))
+		for true {
+			node := &yaml.Node{}
+			err := d.Decode(node)
+			if err == io.EOF {
+				break
+			}
+			assert.NoError(t, err)
+			nodes = append(nodes, node)
+			// walkTree(0, node)
```
ccoVeille: remove this too, please


```suggestion
```

decode_test.go
```diff
@@ -1067,6 +1067,164 @@ func TestDecoder(t *testing.T) {
...
+	fmt.Printf("%s{%d %d %#v:%#v anchor:%#v head:%#v line:%#v foot:%#v %d:%d}\n", strings.Repeat("  ", indent), node.Kind, node.Style, node.Tag, node.Value, node.Anchor, node.HeadComment, node.LineComment, node.FootComment, node.Line, node.Column)
+	for _, item := range node.Content {
+		walkTree(indent + 1, item)
+	}
+	if node.Alias != nil {
+		walkTree(indent + 1, node.Alias)
+	}
+}
+*/
+
```
ccoVeille: If you don't need it, please remove it

```suggestion
```

decode_test.go
```diff
@@ -1067,6 +1067,164 @@ func TestDecoder(t *testing.T) {
...
+func TestDecoderNodes(t *testing.T) {
+	for i, item := range decoderNodeTests {
+		var nodes []*yaml.Node
+		d := yaml.NewDecoder(bytes.NewReader([]byte(item.data)))
+		for true {
```
ccoVeille: This could be used, it would be more classic

```suggestion
		for {
```

emitterc.go
```diff
@@ -361,6 +361,9 @@ func (emitter *yamlEmitter) emitStreamStart(event *yamlEvent) bool {
 // Expect DOCUMENT-START or STREAM-END.
 func (emitter *yamlEmitter) emitDocumentStart(event *yamlEvent, first bool) bool {
 	if event.typ == yaml_DOCUMENT_START_EVENT {
+		// The next element has been parsed because yaml_emitter_need_more_events() demands
+		// one more element for yaml_DOCUMENT_START_EVENT (accumulate == 1)
+		isEmpty := emitter.events[emitter.events_head + 1].typ == yaml_DOCUMENT_END_EVENT
```
ccoVeille: I saw the comment, but isn't safer to consider something like this ?

```suggestion
		isEmpty := len(emitter.events) < emitter.events_head + 1 || emitter.events[emitter.events_head + 1].typ == yaml_DOCUMENT_END_EVENT
```

I'm not sure about the +1 and <, maybe it's +2, or <=
But I think you will get the point

felixfontein: I think `&&` instead of `||` would be better here.

felixfontein:
Hmm, the failing YTS tests made me think more about this. It seems that the comments that belong to the previous (empty) document in case of
```
# comment
---
key: value
```
are stored in `HeadComment` of the `DocumentNode`, which is never used in other cases. So extracting that information from there is sufficient. So maybe the second change in decode.go should better be removed, to avoid empty documents suddenly appearing unexpectedly.

I've updated the PR accordingly.

node_test.go
```diff
@@ -46,8 +46,27 @@ var nodeTests = []struct {
 			}},
 		},
 	}, {
-		"[encode]null\n",
-		yaml.Node{},
```
ccoVeille, line 50: Is there a reason to remove this test ?

I'm unsure it was done on purpose

emitterc.go
ccoVeille: I assumed having no next, could mean it was empty. That's why I suggested that

emitterc.go
```diff
@@ -361,6 +361,7 @@ func (emitter *yamlEmitter) emitStreamStart(event *yamlEvent) bool {
 // Expect DOCUMENT-START or STREAM-END.
 func (emitter *yamlEmitter) emitDocumentStart(event *yamlEvent, first bool) bool {
 	if event.typ == yaml_DOCUMENT_START_EVENT {
+		isEmpty := emitter.events_head+1 < len(emitter.events) && emitter.events[emitter.events_head+1].typ == yaml_DOCUMENT_END_EVENT
```
ccoVeille: I woulije to suggest something Luke thus that could be clearer and Kore idiomatic maybe


```suggestion
		var isEmpty bool
		next, found :=  emitter.events[emitter.events_head+1]
		if found && next.typ == yaml_DOCUMENT_END_EVENT {
			isEmpty = true
		}
```

decode.go
```diff
@@ -152,6 +152,13 @@ func (p *parser) parse() *Node {
 		return p.document()
 	case yaml_STREAM_END_EVENT:
 		// Happens when attempting to decode an empty buffer.
+		if len(p.event.head_comment) > 0 {
+			// The buffer is empty, but there is a comment.
+			// Create a document node which contains this comment.
+			n := p.node(DocumentNode, "", "", "")
+			p.event.head_comment = nil
+			return n
+		}
```
felixfontein, line 171: This change fixes parsing of empty-except-comment documents like:
```
# comment
```
Before this change, `Node.Decode()` would simply return `io.EOF` as an error. Now it returns a `DocumentNode` with `HeadComment` set.

internal/libyaml/emitter.go
```diff
@@ -452,12 +453,15 @@ func (emitter *yamlEmitter) emitDocumentStart(event *yamlEvent, first bool) bool
...
-			if !emitter.putLineBreak() {
+			if !isEmpty && !emitter.putLineBreak() {
 				return false
 			}
 		}
 
 		emitter.state = yaml_EMIT_DOCUMENT_CONTENT_STATE
+		if isEmpty {
+			emitter.state = yaml_EMIT_DOCUMENT_END_STATE
+		}
```
felixfontein, line 468: The changes in this file ensure that serializating a `DocumentNode` with `HeadComment` set but no content results in
```
# comment
```
instead of a serialization error (`expected SCALAR, SEQUENCE-START, MAPPING-START, or ALIAS, but got document end`).

node_test.go
felixfontein: No, that was not intentional.

emitterc.go
felixfontein: I don't think it's possible to get `yaml_DOCUMENT_START_EVENT` not being followed by anything else (except by intentionally sending a broken event sequence to the emitter).

emitterc.go
felixfontein: I don't think this works (https://stackoverflow.com/a/64415778).

ccoVeille: Oh you are right

decode.go
ccoVeille: Thanks for the confirmation.

Maybe you could add a simple comment explaining the purpose of this if block

internal/libyaml/emitter.go
ccoVeille: Maybe you could add an inline comment about this. Your explanation seems valuable

emitterc.go
ccoVeille: Yes, but we don't know how people can use the lib. So I appreciate you added the check on the length.

parserc.go
```diff
@@ -308,6 +308,7 @@ func (parser *yamlParser) parseDocumentStart(event *yamlEvent, implicit bool) bo
 			typ:        yaml_DOCUMENT_START_EVENT,
 			start_mark: token.start_mark,
 			end_mark:   token.end_mark,
+			implicit:   true,
```
ccoVeille: I'm unsure what this change is about.

Could you elaborate a bit here?

Thanks

ccoVeille (approved):
LGTM 👍

I have small suggestions anyway.

Also, I'm curious about other maintainers' feedback

parserc.go
felixfontein: The `implicit` flag is used to mark document start and end events, to distinguish between explicit and implicit versions. In all other three cases (explicit document start, implicit document end, implicit document start) it is set, but for an implicit document start it is not set (and thus defaults to `false`). I don't think this is used anywhere (except when emitting the stream of events again), but this is IMO a clear bug that should be fixed.

decode.go
felixfontein: I added/extended the comments in all places (except for `implicit: true`).

SVilgelm:
Hi @felixfontein, could you please rebase the PR?

felixfontein:
@SVilgelm done.

felixfontein:
I rebased again to remove the conflict with 4645e5c423368846ca2598e1082786e5fbf7a0bd.

parserc.go
felixfontein: In case anyone wonders where this part went: #138.

felixfontein:
Is there any chance of getting this merged before there are again conflicts?

ingydotnet:
@felixfontein Are you mixing up YAML "documents" with "streams"?

An empty file or one with just comments is a yaml stream with 0 documents.

imho Unmarshal should fail on it, because Unmarshal makes a Node from the first document and there are none.

ingydotnet:
```
$ make go-yaml 
go build -o go-yaml ./cmd/go-yaml
$ ./go-yaml -n <<<$''
$ ./go-yaml -n <<<$'# foo'
$ ./go-yaml -n <<<$'# foo\n---\n'
kind: Document
foot: '# foo'
content:
  - kind: Scalar
    tag: '!!null'
$ ./go-yaml -E <<<$''
$ ./go-yaml -E <<<$'# foo'
$ ./go-yaml -E <<<$'# foo\n---\n'
- {Event: DOCUMENT-START, Pos: 2;1}
- {Event: SCALAR, Tag: '!!null', Pos: 4;1}
- {Event: DOCUMENT-END, Pos: 2;1}
$

**Comment by ingydotnet** [comment]
With your PR:
```
$ ./go-yaml -n <<<$''
$ ./go-yaml -n <<<$'# foo\n'
kind: Document
head: |
  # foo
$ ./go-yaml -n <<<$'---\n'
kind: Document
content:
  - kind: Scalar
    tag: '!!null'
$ ./go-yaml -n <<<$'# foo\n---\n'
kind: Document
head: '# foo'
content:
  - kind: Scalar
    tag: '!!null'
$ 
```

I think this is wrong:
```
$ ./go-yaml -n <<<$'# foo\n'
kind: Document
head: |
  # foo
```

There is no document in this stream. But I get your point. We need (or at least want) to attach the comments to something.

Something like this might be acceptable:
```
$ ./go-yaml -n <<<$'# foo\n'
kind: Stream  # or maybe Empty
head: |
  # foo
```

----

Apologies for chiming in so late. It's been a busy few months, but I'm keen to get the go-yaml PR queue cleared out before year end.

I'm not happy with the current comment handling in go-yaml. It has certain bugs including some comments disappearing. Some comments get misplaced, and blank lines and indentation is not preserved.

Comment processing is not part of the YAML Spec for good reasons. There's just no solid way to do it right for all concerns.

v4 is going to have a rich option system. I'm leaning towards v4 handling comments in an agnostic way, where you configure go-yaml with a comment handling library that does your bidding. go-yaml might provide 2 or 3 base strategies, but in the end, you can have as much control as you want.

**Comment by felixfontein** [comment]
> @felixfontein Are you mixing up YAML "documents" with "streams"?

Well, I meant both :) Like here:

```yaml
# foo
---
"bar"
```

I would assume that the comment before `---` belongs to of an "empty (bare) document". (So the stream has two documents.)

And in

```yaml
# comment
```

I would assume it's a stream with a single empty (bare) document.

From your comment, and reading the relevant parts of the spec, I guess this is wrong, though.

> There is no document in this stream. But I get your point. We need (or at least want) to attach the comments to something.
> 
> Something like this might be acceptable:
> 
> ```
> $ ./go-yaml -n <<<$'# foo\n'
> kind: Stream  # or maybe Empty
> head: |
>   # foo
> ```

That would be OK for me, but how would `# foo\n---` be represented?

```
kind: Stream  # or maybe Empty
head: |
  # foo

kind: Document
content:
  - kind: Scalar
    tag: '!!null'
```
?

I guess "Empty" would be more fitting here than "Stream".

> I'm not happy with the current comment handling in go-yaml. It has certain bugs including some comments disappearing. Some comments get misplaced, and blank lines and indentation is not preserved.

I fully agree. This is very painful for all software trying to use this library to handle YAML comments, like [SOPS](github.com/getsops/sops) (which uses go-yaml, and where this and other problems come from).

> v4 is going to have a rich option system. I'm leaning towards v4 handling comments in an agnostic way, where you configure go-yaml with a comment handling library that does your bidding. go-yaml might provide 2 or 3 base strategies, but in the end, you can have as much control as you want.

Is there a timeline when this will be available?

ingydotnet:
> Is there a timeline when this will be available?

See #212

ingydotnet:
I'm really thinking we need a node of  `kind: Stream` in situations where we need stream level info that isn't attached to a Document.

I'll make a PR for this soon.

ingydotnet:
We could also have a `Stream` field for `Document` nodes. That might make more sense.

ingydotnet:
I think that with the new API we can add WithStreamNodes, such the loader.Load() and LoadAll() can return an extra "Node" of "kind: Stream" that appear at the start, end and between every node.

These will contain info that isn't clearly a part of any document.

ingydotnet:
@felixfontein Take a look at #230. It adds optional "stream" nodes to be reported by load. It adds WithStreamNodes.

Also adds `-S` for the CLI:
```
$ ./go-yaml <<<'
> a: 1
> ---
> b: 2
> ' -n -S
---
kind: Stream
encoding: UTF-8
---
kind: Document
content:
- kind: Mapping
  content:
  - kind: Scalar
    text: a
  - kind: Scalar
    tag: '!!int'
    text: "1"
---
kind: Stream
encoding: UTF-8
---
kind: Document
content:
- kind: Mapping
  content:
  - kind: Scalar
    text: b
  - kind: Scalar
    tag: '!!int'
    text: "2"
---
kind: Stream
encoding: UTF-8
```

ingydotnet:
> Allows to correctly parse and serialize
> 
> ```
> # comment
> ```
> 
> without losing the comment (parsing), or getting some error (serialization).

Once #230 merges along with the comment plugin I'm working on, you will be able to use options that will give you a single stream node with that comment as part of it. 

Load would not return any document nodes because there are no documents in that yaml. 
```
$ ./go-yaml -n -S <<<''
---
kind: Stream
encoding: UTF-8
```

at the moment we get:
```
$ ./go-yaml -n -S <<<'# comment'
---
kind: Stream
encoding: UTF-8
```

Because comments only get added to nodes and we don't have a plug-in to change that behavior yet. And we can't change the behavior for V3.

ingydotnet:
@felixfontein #230 is merged. Give it a try. An empty stream will return 1 Stream Node. It can hold a comment but won't do that yet, because we'll need a comment plugin that handles that situation.

We still need the plugin API in place. I'm working hard on that. Should be very soon. Then we'll have have a system where you/we can develop better comment plugins.

felixfontein:
@ingydotnet thanks, I did try it, but without a comment plugin there's indeed not much to see :)

ingydotnet:
Working on that. 🙂

Issue #118, opened 2025-09-12
Apply best practices in terms of security for GitHub actions


ccoVeille:
#116 was somehow an attempt. Let's discuss here about the need, and then it could be implemented.

Here are the things I can think about:

- restrict `permissions` in GHA workflows. Such as using only `read` one

- pin version of GitHub Actions we are using

This would prevent tag hijacking

- add a GitHub action to automatically pin versions that could be added later.

In a way, what would be done for security reasons will be applied in the future.

I can think about [pinact](https://github.com/suzuki-shunsuke/pinact) (hi @suzuki-shunsuke 👋) and its [action](https://github.com/suzuki-shunsuke/pinact-action), but there are others out there.

- make sure it will work with GitHub Dependabot

- consider the implication of having longer yamllint with the hash in GHA version

Here I feel like we could disable the line length setting for `.github/workflows` folder. An ignore rule could be added for line length only via .yamllint.yaml

- take a look at [harden-runner](https://github.com/step-security/harden-runner) GHA. I'm unsure how it would interact or conflict with pinact or other solution we may set up.

There are other things to discuss I think. But there are good bases.

esacteksab:
In the spirit of GitHub Actions security, you may or may not find value in [zizmorcore/zizmor](https://github.com/zizmorcore/zizmor) which has an [action](https://github.com/zizmorcore/zizmor-action). 

With regards to line length, I had wondered about excluding `.github/workflows` in `.yamllint` rather than extending the length as I did in #116, but wasn't sure if excluding the contents of that entire directory was more or less desirable.

A bit off topic, but while iterating on #116 and #117, I found `uv tool run yamllint .` useful in verifying things locally, I thought about adding that to the contributing guide, but wasn't entirely sure where, or again, if that would be welcomed (acknowledging the irony of the contents of that PR).

Looking at `pinact`, I wonder if it would pass `yamllint` as I encountered issues last night, as it expected 2 spaces before a comment and one after, and the formatting appears to do one before, one after, so may need another exclusion (or another reason to just ignore `.github/workflows` directory entirely).

If you all decide what you'd like, I'd be happy to try and implement it.

ccoVeille:
> In the spirit of GitHub Actions security, you may or may not find value in [zizmorcore/zizmor](https://github.com/zizmorcore/zizmor) which has an [action](https://github.com/zizmorcore/zizmor-action). 

[...]

PR #124, opened 2025-09-19
ci: pin github actions versions

caarlos0:
this is recommended to prevent supply chain attacks, as tags can be changed (and the `@v[single digit]` constantly are.

dependabot will still update these - the downside is that there will be more dependabot PRs.

### DIFF
--- a/.github/workflows/check-commit-message.yaml
+++ b/.github/workflows/check-commit-message.yaml
@@ -14,7 +14,7 @@ jobs:
     if: github.event.pull_request.user.login != 'dependabot[bot]'
     steps:
     - name: Checkout code
-      uses: actions/checkout@v5
+      uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
       with:
         fetch-depth: 0  # Fetch all history to ensure all SHAs are available
     - name: Check PR Commits
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -27,7 +27,7 @@ jobs:
           build-mode: autobuild
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v5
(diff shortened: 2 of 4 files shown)


ccoVeille (approved):
You are somehow working on #118

- #118

Do you mind applying the other suggestions that were also listed in the issue?

.github/workflows/check-commit-message.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
     if: github.event.pull_request.user.login != 'dependabot[bot]'
     steps:
     - name: Checkout code
-      uses: actions/checkout@v5
+      uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
```
SVilgelm: please add a comment with the exact version:
```suggestion
      uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
```
Dependabot will update the comment as well, it improves readability

SVilgelm (changes requested):

ccoVeille:
#163 was opened and somehow replaces what you did here @caarlos0

I would suggest you to close it.

.github/workflows/check-commit-message.yaml
caarlos0: yep, done.

SVilgelm:
@caarlos0 I created PR: https://github.com/yaml/go-yaml/pull/163, because this was not moving since Sep 19. If you would like to continue work on this, I"m okay and can close my PR.
Could you please check it and copy the `yamllint` comments to disable `line-lenght` rule, otherwise CI will fail

SVilgelm:
The `yamllint` also wants to see two spaces before the comment: `__#`

caarlos0:
ah, let's leave yours then, no problem :)

Issue #109, opened 2025-09-09
Anchor map key with no space before colon works in v3; breaks in v4


mikefarah:
Not sure if this is intentional in v4 - but this used to work in v3:

```yaml
f:
  a: &a cat
  *a: b  # this specifically
```
Whereas v4 this no longer works - and you need a space between the anchor and colon

```yaml
f:
  a: &a cat
  *a : b  # this will now work 
```

From what I understand the yaml standard was a little vague here - and tools like yamllint also do not like `*a: blah` - I'm happy enough to update the test in yq because I think this is fairly niche if anyone is doing it - just wanted to double check :)

Also this is the only test that failed when I updated to v4! https://github.com/mikefarah/yq/pull/2468

ingydotnet:
https://play.yaml.io/main/parser?input=ZjoKICBhOiAmYSBjYXQKICAqYTogYiAgIyB0aGlzIHNwZWNpZmljYWxseQ==

That's invalid YAML because anchors can contain `:` chars.

libyaml and friends only allowed minimal `[-a-zA-Z0-9]` chars in anchors, but we expanded that.

the spec allows most printable unicode in anchors. 

we changed go-yaml to allow the ascii allowed chars.

If this is causing pain, I have no problem in remeving `:` from the allowed chars in go-yaml.

Let me know.

ingydotnet:
Honestly, I think we should remove `:` for now. 

Maybe even remove it from the spec in a future YAML spec 😄

ccoVeille:
> Maybe even remove it from the spec in a future YAML spec 😄 

Hey, that's cheating!!! You we changing the rule. 🤣😂

> Honestly, I think we should remove `:` for now. 

More seriously, I agree with you

ccoVeille:
What is the PR involved in this change?

I failed to find it back. I feel like it could be `#47`

ingydotnet:
> What is the PR involved in this change?
> 
> I failed to find it back. I feel like it could be `#47`

Yes. https://github.com/yaml/go-yaml/pull/47/files#diff-116f5076d34c636aefab7d3aa208d8fd1eb6beb514c2174720b605a4be572e9dR69

ingydotnet:
I feel wrong about allowing invalid yaml (though we have for a long time here).

Maybe we could warn on this?

It would involve a change in the scanner which is hairy.

I guess I could revert the `:`  for now and leave an open issue here to warn / deprecate / something else.

[...]

PR #125, opened 2025-09-19
Fix alias emitted in mapping

ccoVeille:
The idea is to make sure that a YAML alias in a mapping is followed by a space before the `:`

Related to:
- #47 
- #109 

The issue appeared with the refactoring made with #47 that made sure to respect the YAML specification that allows having a `:` in an anchor and alias name.

The issue was reported by @mikefarah when he tried to migrate on go.yaml.in/yaml/v4

Please note this PR, doesn't fix the way a YAML is considered valid or not.

It simply makes sure to emit valid YAML when dealing with alias as a key in a mapping.


ccoVeille:
I should have mentionned it. This is draft. As a proof of concept and for discussion purpose.

internal/libyaml/emitter.go
```diff
@@ -813,7 +813,16 @@ func (emitter *yamlEmitter) emitBlockMappingKey(event *yamlEvent, first bool) bo
...
 	if emitter.checkSimpleKey() {
 		emitter.states = append(emitter.states, yaml_EMIT_BLOCK_MAPPING_SIMPLE_VALUE_STATE)
-		return emitter.emitNode(event, false, false, true, true)
+		if !emitter.emitNode(event, false, false, true, true) {
+			return false
+		}
+
+		if event.typ == yaml_ALIAS_EVENT {
+			// make sure there's a space after the alias
+			return emitter.put(' ')
```
ccoVeille, line 822: I should add tests for this

ccoVeille:
cc @mikefarah @inteon

ingydotnet:
I would prefer that we revert allowing `:` in anchors.
The libyaml family has always been overly strict about chars allowed in anchors.
They only allow `[-\w]` chars.
We allowed more chars (all the 7bit ascii allowed chars) but we are currently far from spec compliant.
The spec allows all unicode chars minus the syntax chars it disallows.
I feel this is both unimportant and potentially dangerous.
Removing `:` will not bother anyone and will cause less problems (like the one yq reported).

ccoVeille:
pushed back

mikefarah:
Where are we at with this PR? @ccoVeille  / @ingydotnet ?

ingydotnet (approved):

ccoVeille:
@ingydotnet  I rebased and fixed conflicts.

stefanprodan (approved):
LGTM

ccoVeille:
> Got it only after reading all discussions

Yes, it's a fun and complicated issue for a niche 😅

PR #127, opened 2025-09-20
Include the content index in parsing errors

itchyny:
Adding the index to the parsing errors is useful for identifying the error position for package users.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -160,6 +160,7 @@ func (p *parser) node(kind Kind, defaultTag, tag, value string) *Node {
 	if !p.textless {
 		n.Line = p.event.StartMark.Line + 1
 		n.Column = p.event.StartMark.Column + 1
+		n.Index = p.event.StartMark.Index
 		n.HeadComment = string(p.event.HeadComment)
 		n.LineComment = string(p.event.LineComment)
 		n.FootComment = string(p.event.FootComment)
@@ -350,6 +351,7 @@ func (d *decoder) terror(n *Node, tag string, out reflect.Value) {
 		Err:    fmt.Errorf("cannot unmarshal %s%s into %s", shortTag(tag), value, out.Type()),
 		Line:   n.Line,
 		Column: n.Column,
+		Index:  n.Index,
 	})
 }
 
@@ -366,6 +368,7 @@ func (d *decoder) callUnmarshaler(n *Node, u Unmarshaler) (good bool) {
 			Err:    err,
 			Line:   n.Line,
 			Column: n.Column,
+			Index:  n.Index,
 		})
 		return false
 	}
@@ -394,6 +397,7 @@ func (d *decoder) callObsoleteUnmarshaler(n *Node, u obsoleteUnmarshaler) (good 
 			Err:    err,
 			Line:   n.Line,
 			Column: n.Column,
+			Index:  n.Index,
 		})
 		return false
 	}
@@ -601,6 +605,7 @@ func (d *decoder) scalar(n *Node, out reflect.Value) bool {
 					Err:    err,
 					Line:   n.Line,
 					Column: n.Column,
+					Index:  n.Index,
 				})
 				return false
 			}
@@ -788,6 +793,7 @@ func (d *decoder) mapping(n *Node, out reflect.Value) (good bool) {
 						Err:    fmt.Errorf("mapping key %#v already defined at line %d", nj.Value, ni.Line),
 						Line:   nj.Line,
 						Column: nj.Column,
+						Index:  nj.Index,
 					})
 				}
 			}
@@ -940,6 +946,7 @@ func (d *decoder) mappingStruct(n *Node, out reflect.Value) (good bool) {
 						Err:    fmt.Errorf("field %s already set in type %s", name.String(), out.Type()),
 						Line:   ni.Line,
 						Column: ni.Column,
+						Index:  ni.Index,
 					})
 					continue
 				}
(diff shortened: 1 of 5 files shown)


ccoVeille:
I resolved the conflicts that were added by recent PR being merged

itchyny:
I believe the `Line` and `Column` in errors should be 1-origin (and `Index` is of course 0-origin), but 68b682443b806aae3b15486b6a27b4893f66ea6c added test cases that look `Line` and `Column` are 0-origin. `parser.mark.line` (and column) looks to be 0-origin, `Node.Line` (and Column) looks to be 1-origin (that's why `UnmarshalError.Line` is 1-origin). We should use 1-origin for `ParserError.Line` as well.

itchyny:
I'm confused with the checking `if p.parser.context_mark.line != 0`. The value should be 0 on the first line.

itchyny:
Okay, I understand the issue a bit. `yamlParser.fetchStreamEnd` resets the `parser.mark position`, and which seems to be necessary for parsing correct YAML contents, but when the mark of the token `yaml_STREAM_END_TOKEN` is used for the error, the line is unnecessarily beyond the YAML contents. Basically we should use `line := p.parser.context_mark.line + 1`, but when the error happens at the end of the stream (which does not contain newline at the end), we should decrement.

ccoVeille (approved):
Nice catch here @itchyny 

Just to make sure, unless I'm wrong you are fixing a bug that was already present before your changes?

I would like to understand if the PR includes a simple feature request to add the index, or if it also contains a bug fix about line reporting

decode.go
```diff
@@ -107,33 +107,19 @@ func (p *parser) peek() yamlEventType {
...
-	}
-	var column int
-	if p.parser.context_mark.column != 0 {
-		column = p.parser.context_mark.column
-	} else if p.parser.problem_mark.column != 0 {
-		column = p.parser.problem_mark.column
+	line := p.parser.context_mark.line
+	if !p.parser.stream_end_produced {
+		line++
 	}
```
ccoVeille: Please add a comment here about the reason behind this.

decode_test.go
```diff
@@ -1156,7 +1156,8 @@ func TestParserErrorUnmarshal(t *testing.T) {
 	expectedErr := &yaml.ParserError{
 		Message: "could not find expected ':'",
 		Line:    2,
-		Column:  0,
+		Column:  1,
```
ccoVeille: Also I feel like you are fixing the column computing as you said.

I'm a bit afraid it brings confusion with the PR.

Do you think you could open two PRs?

One for fixing line and column that were already there. And one to add the index.

But maybe you and other maintainers would find it over kills

ccoVeille (changes requested):

itchyny:
This PR contains a bug fix for existing problems.

- `ParserError.Line` is 0-origin in the upstream, in some situations (depending on `p.parser.error`). This PR fixes this to be 1-origin consistently, and this changes some error messages as you can see `line 1:` added to the tests.
- `ParserError.Column` is 0-origin from when implemented (68b682443b806aae3b15486b6a27b4893f66ea6c), but `UnmarshalError.Column` is 1-origin. This PR fixes this situation.

There are remaining issues unfortunately.

- When the parsing error happens at the end of the stream (without newline at the end), the Line is correct but Column is wrong (is 1). We can see the bug with the input `v: [`, this yields Column: 1. But I can't fix the issue due to the difficulties of `yaml_STREAM_END_TOKEN`.

decode_test.go
ccoVeille: What are your thoughts who this ?

@itchyny @ingydotnet

itchyny: Separated into #135.

ingydotnet:
Can we get some team opinions on this?

ccoVeille:
I feel like #135 was opened to replace it, no?

EDIT: no #127 was split. So #135 should be rebased and merged first

ccoVeille:
I would like to get the feature the PR brings. The changes are limited. But with other PRs that were merged or the ones that are about to be, the code here might have diverted a lot, no? I mean conflict party

ingydotnet:
> I feel like #135 was opened to replace it, no?
> 
> EDIT: no #127 was split. So #135 should be rebased and merged first

@ccoVeille now that #135 is merge, what do you want to do with this? 

I can resolve the conflicts if you're interested.

ccoVeille:
Yes, please try to rebase and fix this way only the third commit would remain and we could review it again

ingydotnet:
> Yes, please try to rebase and fix this way only the third commit would remain and we could review it again

Working on that now.

ingydotnet:
> Yes, please try to rebase and fix this way only the third commit would remain and we could review it again

@ccoVeille @itchyny Okay, I've reworked this PR on top of the current main. Please have a look and tell me what you think.

node_test.go
```diff
@@ -431,6 +440,17 @@ func formatNodeInfo(n yaml.Node) *NodeInfo {
 		info.Foot = n.FootComment
 	}
 
+	// Include position fields if non-zero
+	if n.Line != 0 {
```
ccoVeille, line 444: You are using omitempty

So they won't be in result if zero, no ?

```go
	info.LineNum = n.Line
	info.Col = n.Column
	info.Index = n.Index
```

ccoVeille (approved):
One suggestion, otherwise 👍

ingydotnet:
#311 solves this for all load phase errors (including parser).

PR #129, opened 2025-09-22
Fix int64 minimum value handling in tests for cross-architecture compatibility

ccoVeille:
The bug arises from the use of generic integer types (int) in test cases involving large numerical values that exceed the range of a 32-bit integer.

In Go, the size of the int type depends on the platform, this behavior causes inconsistencies when handling large integers, such as -9223372036854775808 and 9007199254740993.

On 32-bit systems, these values cannot be represented correctly as int, leading to potential overflow or truncation errors. By explicitly using the int64 type in the affected test cases, the code ensures proper handling and representation of large integers across all platforms, avoiding platform-specific bugs and improving the reliability of the tests.
Reproduce

Run tests setting GOARCH to 386.

GOARCH=386 GOOS=linux go test -vet=off -v

Changes:

    Switched expected values from map[string]interface{} to map[string]int64
    to ensure type consistency and avoid type mismatch failures in DeepEquals.
    Ensured literal -9223372036854775808 is treated safely as an int64 to
    prevent compile-time constant overflow on 32-bit platforms.
    Updated Marshal and Unmarshal tests to explicitly use int64 when working
    with large integer constants that exceed int32 range.

These changes address potential issues with integer overflow and ensure compatibility across platforms with differing integer size representations.

This PR applies https://github.com/kubernetes-sigs/yaml/pull/120 to from https://github.com/kubernetes-sigs/yaml to yaml/go-yaml

This bug was identified and fixed by @arthurbdiniz


decode_test.go
```diff
@@ -439,7 +439,7 @@ var unmarshalTests = []struct {
 	},
 	{
 		"int_overflow: 9223372036854775808", // math.MaxInt64 + 1
-		map[string]int{},
```
liggitt: this bit looks reasonable to me ... switching to a sized int looks good

decode_test.go
```diff
@@ -244,7 +244,7 @@ var unmarshalTests = []struct {
 	},
 	{
 		"bin: -0b1000000000000000000000000000000000000000000000000000000000000000",
-		map[string]any{"bin": -9223372036854775808},
+		map[string]int64{"bin": -9223372036854775808},
```
liggitt: this bit looks questionable ... decoding into `map[string]any` should work properly and preserve as much data as possible

ccoVeille: I'm unsure what you report here. So I would rephrase and you will tell me if I got it.

I feel like you are reporting the fact the tests should make sure we are able to process an untyped int in a map[string]any

If it's what you reported I would say this is already covered by the tested line 211.

So the test line I'm updating here would be more an edge case about the math MaxInt thing when decoding from a binary notation. I'm not even sure there is a need for this test to be honest

What do you think about this?

liggitt: I think it's actually really important to make sure decoding into `map[string]any` does the right thing with integer inputs that are bigger than the platform's unsized int, but which fit within int64.

I think modifying this test case to something like this would be more reasonable:
```go
map[string]any{"bin": int64(-9223372036854775808)},
```

ccoVeille: There is a problem when using int64 then.
The DeepEqual fails on 64bits arch.

Your suggestion might work, but I face something else.

I will have to check, maybe it's a bug with the way go-yaml decodes the integer in a map any.

It could be normal. I just need to check further.

If I fails to find a solution, I will move to arch specific build Go files.

liggitt: Ah, ok... I think the following will work in that case:

```diff
diff --git a/decode_test.go b/decode_test.go
index 4ec7893..bb6fb82 100644
--- a/decode_test.go
+++ b/decode_test.go
@@ -244,7 +244,9 @@ var unmarshalTests = []struct {
 	},
 	{
 		"bin: -0b1000000000000000000000000000000000000000000000000000000000000000",
-		map[string]any{"bin": -9223372036854775808},
+		// 64-bit platforms decode into unsized ints.
+		// 32-bit platforms decode into sized int64.
+		map[string]any{"bin": unsizedIntIfPossibleWithoutOverflow(-9223372036854775808)},
 	},
 	{
 		"decimal: +685_230",
@@ -962,6 +964,16 @@ ba?r: a?bc
 	},
 }
 
+// 64-bit platforms decode into unsized ints.
+// 32-bit platforms decode into sized int64.
+func unsizedIntIfPossibleWithoutOverflow(i int64) any {
+	if i == int64(int(i)) {
+		return int(i)
+	} else {
+		return i
+	}
+}
+
 type M map[string]any
 
 type inlineB struct {
```

ccoVeille: Yes, I came to a solution like this by my own.

It was already suggested by @arthurbdiniz with https://github.com/kubernetes-sigs/yaml/pull/120#issuecomment-3349147793

ingydotnet:
@ccoVeille How are you feeling on this PR?

decode_test.go
```diff
@@ -38,6 +39,16 @@ var negativeZero = math.Copysign(0.0, -1.0)
 
 var unmarshalIntTest = 123
 
+// archSafeInt returns an int64 representation of v that is safe for the current architecture.
+func archSafeInt(v int64) any {
+	if strconv.IntSize == 64 || math.MinInt32 < v && v < math.MaxInt32 {
```
liggitt:
```suggestion
	if strconv.IntSize == 64 || math.MinInt32 <= v && v <= math.MaxInt32 {
```

ccoVeille: indeed thanks, I applied the fix.

ccoVeille:
> @ccoVeille How are you feeling on this PR?

For me this PR should be merged, code is fine now.

I wish I have a few approvals, but it could be merged as this, I think.

liggitt:
changes lgtm, I'd suggest squashing commits

ingydotnet:
@ccoVeille the current changes to `.github/workflows/go.yaml` feel like they don't belong here. Is the branch out of sync?

ccoVeille:
> @ccoVeille the current changes to `.github/workflows/go.yaml` feel like they don't belong here. Is the branch out of sync?

Could you please spot a line of the file that seems out of the PR scope ?

The changes I'm introducing are about testing all architectures (os, CPU and 32/64bits) in the unit tests. Something that was missing and caused the issues to be undetected.

So I'm fixing the unit tests. Make sure the issues cannot happen again.

Maybe I introduced a change that shouldn't be there. Please help to spot the issue if there are any

ingydotnet:
@liggitt @stefanprodan 

I got this older PR resolved and working again with the current code. Can you give it a quick review, please?

decode_test.go
```diff
@@ -40,6 +41,16 @@ var negativeZero = math.Copysign(0.0, -1.0)
 
 var unmarshalIntTest = 123
 
+// archSafeInt returns an int64 representation of v that is safe for the current architecture.
```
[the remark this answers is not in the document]
ingydotnet: 6d8ff7e

decode_test.go
```diff
@@ -235,6 +246,16 @@ var unmarshalTests = []struct {
 		"\t\n",
 	},
 
+	// Cross-architecture numeric tests
+	{
+		"bin: -0b1000000000000000000000000000000000000000000000000000000000000000",
+		map[string]any{"bin": archSafeInt(-9223372036854775808)},
+	},
+	{
```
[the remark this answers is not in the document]
ingydotnet: 6d8ff7e

decode_test.go
```diff
@@ -235,6 +246,16 @@ var unmarshalTests = []struct {
 		"\t\n",
 	},
 
+	// Cross-architecture numeric tests
+	{
+		"bin: -0b1000000000000000000000000000000000000000000000000000000000000000",
+		map[string]any{"bin": archSafeInt(-9223372036854775808)},
```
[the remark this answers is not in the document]
ingydotnet: 6d8ff7e

decode_test.go
```diff
@@ -40,6 +41,16 @@ var negativeZero = math.Copysign(0.0, -1.0)
 
 var unmarshalIntTest = 123
 
+// archSafeInt returns an int64 representation of v that is safe for the current architecture.
+func archSafeInt(v int64) any {
+	if strconv.IntSize == 64 || math.MinInt32 <= v && v <= math.MaxInt32 {
+		return int(v) // int is safe
+	}
+
+	// on 32-bit systems, and v is overflows int, we need to return an int64
```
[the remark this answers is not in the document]
ingydotnet: 6d8ff7e

ingydotnet:
@ccoVeille you might want to look it over as well.

.github/workflows/go.yaml
```diff
@@ -15,17 +15,60 @@ jobs:
...
+          GOARCH: arm64
+          test-args: -race
+        - os: windows-latest
+          GOOS: windows
+          GOARCH: amd64
+          test-args: -race
+        - os: ubuntu-latest
+          GOOS: linux
+          GOARCH: 386
+          test-args: ''  # race is not available on 32 bits architecture
```
ccoVeille, line 53: > [!NOTE]
> The PR is fine, and can be merged as-is.

But for records, about the part of this PR that is about setting race detection arg per OS/ARCH we could do it differently maybe.

I'm just dropping a note that -race availability could be detected by using this

```bash
# Check if current GOOS/GOARCH supports race AND if CGO is enabled
if go tool dist list -json | grep -q '"ImportPath": "runtime/race"' && [ "$(go env CGO_ENABLED)" == "1" ]; then
    echo "Running tests with -race..."
    go test -race ./...
else
    echo "Race detector not supported on this platform/config. Running standard tests..."
    go test ./...
fi
```

@ingydotnet We could use this either in our Makefile, but you could be interested to add such detection in your go.mk file you maintain in makeplus

ingydotnet: Decent idea. Let's take it up later...

ccoVeille: I created
- https://github.com/yaml/go-yaml/issues/215

ingydotnet:
> I'm good with the PR.

Thanks for checking. Given that I'm going to go ahead and merge it.

PR #130, opened 2025-09-27
Add `Column` field to `ParserError`

rhysd:
`yaml.UnmarshalError` has `Column` field but `yaml.ParserError` doesn't. This PR adds the missing field so that users can report more accurate error position on their error handling.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -121,13 +121,19 @@ func (p *parser) fail() {
 			line++
 		}
 	}
+	var column int
+	if p.parser.context_mark.column != 0 {
+		column = p.parser.context_mark.column
+	} else if p.parser.problem_mark.column != 0 {
+		column = p.parser.problem_mark.column
+	}
 	var msg string
 	if len(p.parser.problem) > 0 {
 		msg = p.parser.problem
 	} else {
 		msg = "unknown problem parsing YAML content"
 	}
-	fail(&ParserError{msg, line})
+	fail(&ParserError{msg, line, column})
 }
 
 func (p *parser) anchor(n *Node, anchor []byte) {
--- a/decode_test.go
+++ b/decode_test.go
@@ -1145,19 +1145,32 @@ func TestDecoderErrors(t *testing.T) {
 	}
 }
 
-func TestParserError(t *testing.T) {
+func TestParserErrorUnmarshal(t *testing.T) {
 	var v struct {
 		A, B int
 	}
 	data := "a: 1\n=\nb: 2"
 	err := yaml.Unmarshal([]byte(data), &v)
 	asErr := new(yaml.ParserError)
-	if !errors.As(err, &asErr) {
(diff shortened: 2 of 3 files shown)


decode.go
```diff
@@ -121,13 +121,19 @@ func (p *parser) fail() {
 			line++
 		}
 	}
+	var column int
+	if p.parser.context_mark.column != 0 {
+		column = p.parser.context_mark.column
+	} else if p.parser.problem_mark.column != 0 {
```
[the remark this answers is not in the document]
rhysd: The code style is aligned to the above `line` case. I don't think this suggestion is reasonable.

yaml.go
```diff
@@ -323,6 +323,7 @@ func failf(format string, args ...any) {
 type ParserError struct {
 	Message string
 	Line    int
+	Column  int
 }
 
 func (e *ParserError) Error() string {
```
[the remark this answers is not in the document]
rhysd: I'm not sure column number should be included in the canonical error message. I'd like to await maintainer's review.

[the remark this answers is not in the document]
ccoVeille: I feel like there is no need, when there is a parsing error the issue is often on the line and focusing on a column might be imprecise.

ccoVeille (changes requested):
Please add unit tests to cover this change

rhysd:
@ccoVeille Thanks for the point. I added the new test case where the column number is not zero. I noticed that `ParserError` returned from `Decode` method was not tested. So I added it as a separate test case testing it.

decode_test.go
```diff
@@ -1,4 +1,5 @@
 //
+
```
ccoVeille: Please revert this change, or explain why you did it

decode_test.go
```diff
@@ -1158,6 +1159,23 @@ func TestParserError(t *testing.T) {
...
+	}
+	assert.DeepEqual(t, expectedErr, asErr)
+}
+
+func TestParserErrorDecoder(t *testing.T) {
+	var v any
+	data := "value: -"
+	err := yaml.NewDecoder(strings.NewReader(data)).Decode(&v)
+	asErr := new(yaml.ParserError)
+	if !errors.As(err, &asErr) {
```
ccoVeille: You could use assert.ErrorAs here

That's its purpose

decode_test.go
rhysd: Ah, this is my bad. I'll force push the next change reverting this.

decode_test.go
rhysd: Thanks. This is the first time to use `assert` library for me. I'll try it.

ccoVeille: It's our in-house assertion library that mimicks popular ones such as testify and co. But we wanted to avoid depending on anything. Because often these lib depends on YAML lib such as the one in this project.

Doing like this avoids possible cyclic dependencies.

ingydotnet:
Squashed 3 commits to 1 and pushed back to @rhysd

PR #131, opened 2025-10-01
Update the GNUmakefile

### LINKED ISSUE #120: gofmt or gofumpt?

ingydotnet:
* Use latest Makes commit.
* Auto-install golangci-lint for fmt and lint targets
* Update CONTRIBUTING.md
* Add a `lint` target
* Allow YTS-TAG and YTS-URL overrides
* General Makefile hygiene

Fixes #120 not installing golangci-lint

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -18,10 +18,6 @@ GO-VERSION ?= 1.24.0
 endif
 GO-VERSION-NEEDED := $(GO-VERSION)
 
-GOLANGCI-LINT-VERSION ?= v2.5.0
-GOLANGCI-LINT-INSTALL := \
-  https://github.com/golangci/golangci-lint/raw/main/install.sh
-
 # yaml-test-suite info:
 YTS-URL ?= https://github.com/yaml/yaml-test-suite
 YTS-TAG ?= data-2022-01-17
@@ -53,7 +49,12 @@ SHELL-NAME := makes go-yaml
 include $(MAKES)/clean.mk
 include $(MAKES)/shell.mk
 
-MAKES-CLEAN := $(dir $(YTS-DIR))
+GOLANGCI-LINT-VERSION ?= v2.5.0
+GOLANGCI-LINT-INSTALL := \
+  https://github.com/golangci/golangci-lint/raw/main/install.sh
+GOLANGCI-LINT := $(LOCAL-BIN)/golangci-lint-$(GOLANGCI-LINT-VERSION)
+
+MAKES-CLEAN := $(dir $(YTS-DIR)) $(GOLANGCI-LINT)
 
 v ?=
 count ?= 1
@@ -78,10 +79,10 @@ test-yts-fail: $(GO-DEPS) $(YTS-DIR)
 	@echo 'Testing yaml-test-suite failures'
 	@RUNFAILING=1 bash -c "$$yts_pass_fail"
 
-fmt: golangci-lint
+fmt: $(GOLANGCI-LINT)
 	$< fmt ./...
 
-lint: golangci-lint
+lint: $(GOLANGCI-LINT)
 	$< run
 
 fumpt: $(GO)
@@ -123,7 +124,7 @@ define yts_pass_fail
 endef
 export yts_pass_fail
 
-golangci-lint: $(GO-DEPS)
-	@[[ -f $$(go env GOPATH)/bin/$@ ]] || \
-	  curl -sSfL $(GOLANGCI-LINT-INSTALL) | \
-	    bash -s -- -b $$(go env GOPATH)/bin $(GOLANGCI-LINT-VERSION)
+$(GOLANGCI-LINT): $(GO-DEPS)
+	curl -sSfL $(GOLANGCI-LINT-INSTALL) | \
+	  bash -s -- -b $(LOCAL-BIN) $(GOLANGCI-LINT-VERSION)
+	mv $(LOCAL-BIN)/golangci-lint $@


yaml.go
```diff
@@ -221,7 +221,7 @@ func Marshal(in any) (out []byte, err error) {
 	e.marshalDoc("", reflect.ValueOf(in))
 	e.finish()
 	out = e.out
-	return
+	return out, err
```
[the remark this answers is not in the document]
ingydotnet: This change was made by `make fmt`.
Is it what we want, or do we need to adjust fmt?

GNUmakefile
```diff
@@ -12,88 +12,97 @@ endif
...
 endef
+export yts_pass_fail
+
+GOLANGCI-LINT-INSTALL := \
+  https://github.com/golangci/golangci-lint/raw/main/install.sh
+
+golangci-lint: $(GO-DEPS)
+	@[[ -f $$(go env GOPATH)/bin/$@ ]] || \
+	  curl -sSfL $(GOLANGCI-LINT-INSTALL) | \
+	    sh -s -- -b $$(go env GOPATH)/bin
```
ccoVeille: This is great as it install the latest golangci-lint version.

But I feel like we should somehow have a variable to check the current golangci-lint version. 

Somehow having golangci-lint v2.3.0 won't have the same error reporting.

Usually, a local golangci-lint is installed with Makefile in a local build dir in the repository.

GNUmakefile
```diff
@@ -12,88 +12,97 @@ endif
...
+  fi
 )
 endef
+export yts_pass_fail
+
+GOLANGCI-LINT-INSTALL := \
+  https://github.com/golangci/golangci-lint/raw/main/install.sh
+
+golangci-lint: $(GO-DEPS)
+	@[[ -f $$(go env GOPATH)/bin/$@ ]] || \
```
ccoVeille: It should only check about the fact golangci-lint is available or not via the PATH, so with command -v, no ?

For example, golangci-lint could be available locally via nix, homebrew, rpm, deb …

Here I feel there is a problem because we could have the following case:

- golangci-lint is available locally
- the Makefile installs it in GOPATH/bin
- make file calls golangci-lint but not the fresh installed one

Another use case would be this

- golangci-lint is not installed
- the Makefile installs it in GOPATH/bin
- GOPATH/bin is not in the PATH
- make file calls golangci-lint, but it fails to find it

GNUmakefile
ccoVeille: Go is not fun for this.

You should check `go env GOBIN`, and fallback on `$(go env GOPATH)/bin` if GOBIN is empty

ingydotnet: Good point. In 1b1feb2 I've added `GOLANGCI-LINT-VERSION ?= v2.5.0`
You can override with: `make fmt GOLANGCI-LINT-VERSION=v2.3.0`

GNUmakefile
ingydotnet: I don't think it makes sense to use anything but the Makefile installed specific version.
However I forgot that we don't always install a local Go (which I would have preferred), so we might be picking up the wrong one. I'll fix this.

ingydotnet: Fixed in 1305e0b
We install to `$(LOCAL-BIN)` which only the Makefile knows about. No depending on GO* vars which are not deterministic here.
I suffixed the binary with the version so version overrides would trigger an install if needed.

GNUmakefile
ingydotnet: 1305e0b gets rid of looking at `GO*` env vars.

ingydotnet:
I do think it's a tiny bit risky to keep allowing use of a system installed Go. But only a tiny bit.
We want the DX to be fully deterministic, imho.

yaml.go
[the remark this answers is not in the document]
ingydotnet: @ccoVeille what do you think about this? Should we ignore Copilot here. Or is golangci-lint wrong?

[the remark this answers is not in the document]
ccoVeille: gofumpt was updated to limit the usage of naked return that is possible but often bring confusion.

So Copilot is wrong here.

gofumpt is right.

CONTRIBUTING.md
```diff
@@ -70,32 +70,38 @@ Before submitting an issue, please:
...
 - Make your changes
-- Run tests and linters
-  - `make test` to run tests locally
-  - `golangci-lint run` to lint locally (requires [golangci-lint](https://golangci-lint.run/) to be installed.)
+- Run tests, linters and formatters
+  - `make test-all`
+  - `make lint`
+  - `make fmt`
+  - `make tidy`
+  - `make vet`
```
SVilgelm, line 112: `make vet` is not needed, `golangci-lint` also runs `govet` and will catch all errors

CONTRIBUTING.md
```diff
@@ -70,32 +70,38 @@ Before submitting an issue, please:
...
+- Installing Go is not necessary. See "The Makefile" below.
 - Fork and clone the repository
 - Make your changes
-- Run tests and linters
-  - `make test` to run tests locally
-  - `golangci-lint run` to lint locally (requires [golangci-lint](https://golangci-lint.run/) to be installed.)
+- Run tests, linters and formatters
+  - `make test-all`
+  - `make lint`
+  - `make fmt`
```
SVilgelm, line 110: `make fmt` is not needed, `golangci-lint` will catch all formatting errors

ingydotnet: Of course we expect CI to catch things. But I think these are useful steps for new contributors (the people reading this).
I'd certainly run these myself before pushing.

SVilgelm: Hmm, I was wrong with this case. `make fmt` is needed, but it should run `golangci-lint fmt'

CONTRIBUTING.md
SVilgelm: will be removed in a separate PR: https://github.com/yaml/go-yaml/pull/158

SVilgelm:
I would be glad to see this PR merged as soon as possible :)

SVilgelm (approved):

ingydotnet:
@SVilgelm Can you doublecheck my merge to resolve conflicts?
Normally we need 2 team reviews, but I'll merge if you :+1:

GNUmakefile
```diff
@@ -1,7 +1,7 @@
 # Auto-install https://github.com/makeplus/makes at specific commit:
 MAKES := .cache/makes
 MAKES-LOCAL := .cache/local
-MAKES-COMMIT ?= 840c40310d5b93b977e7d516d37ea9a2137c2d96
+MAKES-COMMIT ?= 654f7c57ca30a2b08cb4aab8bb0c0d509510ad81
```
SVilgelm, line 4: For learning purposes, what is this line for?

ingydotnet:
> @SVilgelm Can you doublecheck my merge to resolve conflicts? Normally we need 2 team reviews, but I'll merge if you 👍

Sorry, I phrased that ambiguously. I don't know if your :+1: means you'll look or all-good.

Please reapprove if the latter.

PS: I'm mostly unsure if I picked the preferred `fmt` makefile rule...

SVilgelm:
No worries, I understand how GitHub flow works 😀

SVilgelm:
I approved . And that 👍 meant that I'm going to look at the PR

GNUmakefile
dolmen: While I understand what it does (see the makeplus/make project), I don't like that a Go project depends on such an external dependency with versioning maintained obscurely via Git hashes.

This is a supply chain issue because, as a reviewer I can't just verify it by looking at the source code.

ingydotnet: > While I understand what it does (see the makeplus/make project), I don't like that a Go project depends on such an external dependency with versioning maintained obscurely via Git hashes.
> 
> This is a supply chain issue because, as a reviewer I can't just verify it by looking at the source code.

@dolmen,

I only added this because the project is important enough to make it completely deterministic.
Using a version tag is less so, since tags can be changed.

You can verify it's source code here: https://github.com/makeplus/makes/tree/654f7c57ca30a2b08cb4aab8bb0c0d509510ad81

How would you suggest changing things?

ingydotnet: > For learning purposes, what is this line for?

We are using https://github.com/makeplus/makes for reusable Makefile parts.
Any `make` command will clone/cache that repo first time.

https://github.com/yaml/go-yaml/blob/c0c5dccf8a16db9fbcfefc64664c216fbb985063/GNUmakefile#L40
adds go support. It auto installs Go when a rule using `go` is invoked.

No prerequisites are needed to interact with the go-yaml repo except git, bash, gnu make and curl.

The SHA is used to pin the `makes` repo to make everything 100% deterministic.

The only thing I don't like is that some of the team wanted the makefile to use their local go if present.
This could make things not deterministic, but I won't worry about it for now.

SVilgelm: > The only thing I don't like is that some of the team wanted the makefile to use their local go if present.
> This could make things not deterministic, but I won't worry about it for now.

I never install go on System or User level. I prefer to use Goland and use its interface to install the version of Go I need for a particular project.
This helps me to switch the version of go on a fly.

With our setup you can:
```
make test GO-VERSION=1.23.4
```

You can also start a subshell with that version with

```
make shell GO-VERSION=1.23.4
```

And then do any go thing or make thing, with just `make test` etc...

I like this better because it works the same for everyone with zero friction.

Btw when I say it auto installs things, it's all temporal under `./.cache/.local/`. Won't affect your system at all.

PR #132, opened 2025-10-01
Remove unreachable code for binary, octal integers

itchyny:
This code is unnecessary because `strconv.ParseInt` and `strconv.ParseUint`
with base 0 already parses the binary, octal and hexadecimal integers.

### DIFF
--- a/resolve.go
+++ b/resolve.go
@@ -217,56 +217,6 @@ func resolve(tag string, in string) (rtag string, out any) {
 					return floatTag, floatv
 				}
 			}
-			if strings.HasPrefix(plain, "0b") {
-				intv, err := strconv.ParseInt(plain[2:], 2, 64)
-				if err == nil {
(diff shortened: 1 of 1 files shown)


ingydotnet:
@itchyny Thanks. How did you find this code? I assume there's some standard tool for that?

itchyny:
@ingydotnet I was just reading this code and noticed it does not check `0x` which smells like a bug. But tests contains hexadecimal number cases, and noticed that `strconv.ParseInt(plain, 0, 64)` parses the binary and octal numbers (including the `0777` format), not just in hexadecimal.

dolmen:
Related: #144, #145

itchyny:
I don't think they are related. Even after merging this PR, binary and octal integers are still supported (the behavior does not change).

ingydotnet (approved):

PR #133, opened 2025-10-02
Fix decoding !!float tags with uint64 values

itchyny:
`!!float 18446744073709551615` cannot be decoded into `float64`.

### DIFF
--- a/resolve.go
+++ b/resolve.go
@@ -150,6 +150,10 @@ func resolve(tag string, in string) (rtag string, out any) {
 					rtag = floatTag
 					out = float64(v)
 					return
+				case uint64:
+					rtag = floatTag
+					out = float64(v)
+					return
 				case int:
 					rtag = floatTag
 					out = float64(v)
--- a/decode_test.go
+++ b/decode_test.go
@@ -570,6 +570,14 @@ var unmarshalTests = []struct {
 		map[string]any{"v": float64(-1)},
 	},
 	{
+		"v: !!float 9223372036854775807",
+		map[string]any{"v": float64(math.MaxInt64)},
+	},
+	{
+		"v: !!float 18446744073709551615",
+		map[string]any{"v": float64(math.MaxUint64)},
+	},
+	{
 		"v: !!null ''",
 		map[string]any{"v": nil},
 	},


resolve.go
```diff
@@ -150,6 +150,10 @@ func resolve(tag string, in string) (rtag string, out any) {
 					rtag = floatTag
 					out = float64(v)
 					return
+				case uint64:
+					rtag = floatTag
```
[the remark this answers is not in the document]
ccoVeille: @itchyny what do you think about the example provided?

ccoVeille (approved):
LGTM 👍 I checked it on GOARCH=386 to make sure

ingydotnet:
I'll need to get this working again.

ingydotnet:
@itchyny @ccoVeille I got the conflicts resolved.

ingydotnet:
Dammit. I hit the wrong button! Sorry. I'll make a new PR with @itchyny as the commit author.

ingydotnet:
It looks like I can reopen it.

ingydotnet:
Phew!!!

resolve.go
```diff
@@ -150,6 +150,10 @@ func resolve(tag string, in string) (rtag string, out any) {
 					rtag = floatTag
 					out = float64(v)
 					return
+				case uint64:
```
ccoVeille, line 142: I'm unsure why the type assertion in the switch catch int64 and int

But here I see only uint64 being added and not uint.

I'm unsure why this is different, or simply why we had int and int64 previously

testdata/decode.yaml
```diff
@@ -766,6 +766,20 @@
     want:
       v: -1.0
 
+- decode:
+    name: explicit float tag on MaxInt64
+    yaml: 'v: !!float 9223372036854775807'
+    type: map[string]any
+    want:
+      v: 9.223372036854776e+18
```
ccoVeille, line 809: Not nessarily needed here, but can we add sometheing to the test suite about bring able to say we `want v:` but being able to say and so check the type of v ?

I mean when it's an int, uint64, uint, int64, float64

We might only need in with tests such as loading things into `any`, but I would love this to be possible to be able to validate what we get.

I feel like the way to transform yaml to Go via Load, then Go to YAML, leads to lose information. The test suite is about testing Go, so we should be able to see what we got in Go. Its YAML representation when dumped doesn't have this level of information

ingydotnet:
I feel like #134 and #136 work could be added to here. They are all related.
I'll add a commit to bring in the #134 and #136 work to here and close those PRs.

testdata/decode.yaml
```diff
@@ -674,12 +674,46 @@
...
+    want:
+      v: MaxUint64
+
 - decode:
     name: uint64 underflow
     yaml: 'uint64_underflow: -1'
     type: map[string]uint64
     want: {}
 
+# Float to int boundary tests (PR #134)
```
ingydotnet, line 699: Is this test showing that we just ignore pairs with values that overflow?

Shouldn't we error?

ingydotnet: I asked Claude to go over this PR with me. Here's the discussion: https://gist.github.com/ingydotnet/6269a23820bf8572048769c2f65aa797

@itchyny @ccoVeille @liggitt thoughts?

ccoVeille: This PR showed things that will either overflow or have precision loss.

This PR is limited to float to uint somehow.

I feel like there might be other issues about precision loss hidden in thr darkness.

I didn't check, but the precision could also be considered when someone is doing large enough int64 to float64. here the issue would be in Go, not on go-yaml, but everything is in Go.

I might be wrong about my assumptions, but I wanted to raise the fact the problem might be larger than what we think right now.

I would be OK to merge this PR as-is, and then open an issue to try to address precision loss.

Also, I think that whatever we choose to do, we should provide a way for user to bypass any error. So the "accept precision loss" should be a plugin, and not accepting it the default of the v4.

resolve.go
ccoVeille: @ingydotnet @itchyny any thoughts here ?

itchyny: Because these are the resolved types. See https://github.com/yaml/go-yaml/blob/8a35024c7b6f20515706ac075b932ee258af39a8/resolve.go#L201-L213.

ccoVeille: OK, indeed. Thanks

PR #134, opened 2025-10-02
Fix underflow check for float to int conversion

itchyny:
Currently, decoding `v: -1.0` into `map[string]uint64` results into `map[string]uint64{"v": 0}` (should be an empty map like overflow case) because it does not check underflow.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -649,12 +649,12 @@ func (d *decoder) scalar(n *Node, out reflect.Value) bool {
 				return true
 			}
 		case float64:
-			if !isDuration && resolved <= math.MaxInt64 && !out.OverflowInt(int64(resolved)) {
+			if !isDuration && resolved >= math.MinInt64 && resolved <= math.MaxInt64 && !out.OverflowInt(int64(resolved)) {
 				out.SetInt(int64(resolved))
 				return true
 			}
 		case string:
-			if out.Type() == durationType {
+			if isDuration {
 				d, err := time.ParseDuration(resolved)
 				if err == nil {
 					out.SetInt(int64(d))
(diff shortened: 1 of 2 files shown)


decode.go
```diff
@@ -649,12 +649,12 @@ func (d *decoder) scalar(n *Node, out reflect.Value) bool {
...
 			}
 		case float64:
-			if !isDuration && resolved <= math.MaxInt64 && !out.OverflowInt(int64(resolved)) {
+			if !isDuration && resolved >= math.MinInt64 && resolved <= math.MaxInt64 && !out.OverflowInt(int64(resolved)) {
 				out.SetInt(int64(resolved))
 				return true
 			}
 		case string:
-			if out.Type() == durationType {
+			if isDuration {
```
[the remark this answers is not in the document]
itchyny: Same value.

decode.go
```diff
@@ -649,12 +649,12 @@ func (d *decoder) scalar(n *Node, out reflect.Value) bool {
 				return true
 			}
 		case float64:
-			if !isDuration && resolved <= math.MaxInt64 && !out.OverflowInt(int64(resolved)) {
+			if !isDuration && resolved >= math.MinInt64 && resolved <= math.MaxInt64 && !out.OverflowInt(int64(resolved)) {
 				out.SetInt(int64(resolved))
```
ccoVeille, line 655: You are fixing float being lower than MinInt64 could you add a test for it ?

Thanks

itchyny: I think `int_underflow_float` case covers it.

ccoVeille: Indeed. I missed it 🤦‍♂️

ccoVeille (approved):

ccoVeille (approved):

ingydotnet:
Merged into #133 (which has conflicts resolved)

PR #135, opened 2025-10-02
Fix Line and Column in ParserError to be 1-origin

itchyny:
This is a cherry-pick from #132.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -107,26 +107,11 @@ func (p *parser) peek() yamlEventType {
 }
 
 func (p *parser) fail() {
-	var line int
-	if p.parser.context_mark.line != 0 {
-		line = p.parser.context_mark.line
-		// Scanner errors don't iterate line before returning error
-		if p.parser.error == yaml_SCANNER_ERROR {
-			line++
-		}
-	} else if p.parser.problem_mark.line != 0 {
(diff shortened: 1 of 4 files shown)


ingydotnet:
I was able to get Claude to resolve these conflicts. I kept @itchyny as the author and then force-pushed. 
We'll need to review again.

internal/libyaml/parser.go
```diff
@@ -688,6 +688,7 @@ func (parser *Parser) parseNode(event *Event, block, indentless_sequence bool) b
 	if block {
 		context = "while parsing a block node"
 	}
+	parser.skipToken()
```
ccoVeille: While the other changes are clear about column being displayed with +1 as it's logic for the column in a line.

I don't get what is the change on this line. I feel like it might be the "line" part of "Fix Line and Column in ParserError to be 1-origin"

I would like to see a test for this. Or simply remove this change

ingydotnet: I'll work on that.

ingydotnet: @ccoVeille fixed in e34903c

ccoVeille (approved):

ingydotnet (approved):

PR #137, opened 2025-10-04
Add fuzzing test for `yaml.Marshal` and `yaml.Unmarshal`

rhysd:
This PR adds [a fuzzing test](https://en.wikipedia.org/wiki/Fuzzing) for go-yaml package using [Go's standard fuzzing framework](https://go.dev/doc/security/fuzz/).

- Add a minimal fuzz test case to validate `yaml.Unmarshal` and `yaml.Marshal` work without panics
- Add `make` task to run fuzzing test in one command (`make fuzz`)
- Add CI workflow which runs the fuzzing test for 10 minutes weekly (on every Saturday 6:00 AM). Since running fuzzing test takes a bit longer time, it is not suitable for running on every commit


GNUmakefile
```diff
@@ -39,6 +39,7 @@ include $(MAKES)/shell.mk
 
 v ?=
 count ?= 1
+time ?= 60s
```
ccoVeille: I would fuzztime as a variable name then

fuzz_test.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"io/fs"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"go.yaml.in/yaml/v4"
+)
+
+func setupSeedCorpus(f *testing.F) {
```
ccoVeille, line 12:
```suggestion
func setupSeedCorpus(f *testing.F) {
	f.Helper()

```

ccoVeille (approved):

.github/workflows/fuzz.yaml
```diff
@@ -0,0 +1,25 @@
+name: Fuzzing test
+
+on:
+  schedule:
+  - cron: 0 6 * * 6
+
+permissions:
+  contents: read  # for actions/checkout to fetch code
```
SVilgelm: please change it to `read-all`, because it is unclear what permissions are in use for other types

```suggestion
permissions: read-all
```

.github/workflows/fuzz.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  fuzz:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v5
```
SVilgelm: we changed all other workflows to pin the actions to sha, could you please check the changes in main branch and update this one?

.github/workflows/fuzz.yaml
```diff
@@ -0,0 +1,22 @@
...
+      uses: actions/setup-go@v6
+      with:
+        go-version: stable
+    - name: Run fuzzing test for 10 minutes
+      run: make fuzz time=600s
+    - name: Upload failing inputs as artifact
+      uses: actions/upload-artifact@v4
+      with:
+        path: testdata/fuzz
+        if-no-files-found: ignore
```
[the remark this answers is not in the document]
SVilgelm: agree, you probably need to set write permission for artifacts

SVilgelm (changes requested):

rhysd:
I apologize for the lack of response because I was working on other projects. I rebased this branch onto the latest main branch and addressed all comments.

ingydotnet:
See #210

ccoVeille:
We do have an interest in considering fuzz again now #210 is merged

ingydotnet:
> We do have an interest in considering fuzz again now #210 is merged

@ccoVeille are you arguing for or against this?

Please elaborate.

ccoVeille:
I was saying that now #210 is merged, the work on fuzz that was initiated in this PR can start again. Fuzz is great thing to detect issues, we do have interested in it

ccoVeille:
Maybe this could be considered

https://github.com/srlabs/golibafl


Found it via @nikolaydubina and its wonderful https://github.com/nikolaydubina/go-recipes

Especially with this post

- https://github.com/nikolaydubina/go-recipes/issues/273

ingydotnet:
I'd like to see our fuzzing story and code worked on, certainly.

@rhysd would you care to look at the various fuzzing things we have in place now, and steer this (or an alternate PR) in the direction we need to go?

nikolaydubina:
interesting. probably effective fuzzing can be done even without LibAFL. you just need effective corpus. probably can run AI loop: read whole repo -> run fuzz tests -> get coverage -> feed that back to LLM -> ask LLM to write new test input file -> add it to `/testdata` in repo -> repeat.

probably this can be even done in some simple bash script or something and you will really rapidly accumulate a very good test corpus. no need to waste 30s of CPU cycles either on every push haha.

Happy New year guys! 🎄

PR #148, opened 2025-10-15
Add comment into to -t token output in CLI

ingydotnet:
Comment into was missing from token output in the go-yaml CLI.

```
$ <<<'a: b # c' ./go-yaml -t
- {Token: STREAM-START}
- {Token: BLOCK-MAPPING-START}
- {Token: KEY}
- {Token: SCALAR, Value: a}
- {Token: VALUE}
- {Token: SCALAR, Value: b, Line: '# c'}
- {Token: BLOCK-END}
- {Token: STREAM-END}
```

### DIFF
--- a/cmd/go-yaml/parser.go
+++ b/cmd/go-yaml/parser.go
@@ -45,6 +45,28 @@ func (p *Parser) Next() (*Token, error) {
 		EndColumn:   int(yamlToken.EndMark.Column),
 	}
 
+	// Call unfoldComments to process comment information from the parser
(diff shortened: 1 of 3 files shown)


ccoVeille:
@ingydotnet do you think you could add unit tests to check the rendering of a basic example like the one you provided in the PR description?

ingydotnet:
> @ingydotnet do you think you could add unit tests to check the rendering of a basic example like the one you provided in the PR description?

Yes, but I would rather do that in a subsequent PR.

See #149

SVilgelm (approved):

PR #151, opened 2025-10-22
Use go errors for emitter and writer

inteon:
instead of returning booleans and storing the error message in the emitter, we now return the error value directly.

### DIFF
--- a/internal/libyaml/emitter.go
+++ b/internal/libyaml/emitter.go
@@ -142,11 +142,6 @@ func (emitter *Emitter) writeLineBreak(s []byte, i *int) error {
 	return nil
 }
 
-// Construct emitter error.
-func emitterError(problem string) error {
-	return errors.New(problem)
-}
-
 // Emit an event.
 func (emitter *Emitter) Emit(event *Event) error {
 	emitter.events = append(emitter.events, *event)
@@ -210,7 +205,7 @@ func (emitter *Emitter) appendTagDirective(value *TagDirective, allow_duplicates
 			if allow_duplicates {
 				return nil
 			}
-			return emitterError("duplicate %TAG directive")
+			return errors.New("duplicate %TAG directive")
 		}
 	}
 
@@ -317,7 +312,7 @@ func (emitter *Emitter) stateMachine(event *Event) error {
 		return emitter.emitBlockMappingValue(event, false)
 
 	case EMIT_END_STATE:
-		return emitterError("expected nothing after STREAM-END")
+		return errors.New("expected nothing after STREAM-END")
 	}
 	panic("invalid emitter state")
 }
@@ -325,7 +320,7 @@ func (emitter *Emitter) stateMachine(event *Event) error {
 // Expect STREAM-START.
 func (emitter *Emitter) emitStreamStart(event *Event) error {
 	if event.Type != STREAM_START_EVENT {
-		return emitterError("expected STREAM-START")
+		return errors.New("expected STREAM-START")
 	}
 	if emitter.encoding == ANY_ENCODING {
 		emitter.encoding = event.encoding
@@ -482,7 +477,7 @@ func (emitter *Emitter) emitDocumentStart(event *Event, first bool) error {
 		return nil
 	}
 
-	return emitterError("expected DOCUMENT-START or STREAM-END")
+	return errors.New("expected DOCUMENT-START or STREAM-END")
 }
 
 // emitter preserves the original signature and delegates to
@@ -519,7 +514,7 @@ func (emitter *Emitter) emitDocumentContent(event *Event) error {
 // Expect DOCUMENT-END.
 func (emitter *Emitter) emitDocumentEnd(event *Event) error {
 	if event.Type != DOCUMENT_END_EVENT {
-		return emitterError("expected DOCUMENT-END")
+		return errors.New("expected DOCUMENT-END")
 	}
 	// [Go] Force document foot separation.
 	emitter.foot_indent = 0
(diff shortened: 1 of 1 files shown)


internal/libyaml/writer.go
```diff
@@ -22,27 +22,27 @@
...
 
-// Set the writer error and return false.
-func (emitter *Emitter) setWriterError(problem string) bool {
-	emitter.ErrorType = WRITER_ERROR
-	emitter.Problem = problem
-	return false
+import "errors"
+
+// Construct writer error.
+func writerError(problem string) error {
```
SVilgelm: What is a purpose of this helper? It would be easier to use `fmt.Errorf` or `errors.New` directly

internal/libyaml/writer.go
```diff
@@ -22,27 +22,27 @@
...
 
 	// Check if the buffer is empty.
 	if emitter.buffer_pos == 0 {
-		return true
+		return nil
 	}
 
 	if err := emitter.write_handler(emitter, emitter.buffer[:emitter.buffer_pos]); err != nil {
-		return emitter.setWriterError("write error: " + err.Error())
+		return writerError("write error: " + err.Error())
```
SVilgelm: Wouldn't it be better to use `fmt.Errorf("write error: %w", err)` instead?

internal/libyaml/writer.go
ccoVeille: @SVilgelm is right

internal/libyaml/emitter.go
```diff
@@ -132,31 +139,29 @@ func (emitter *Emitter) writeLineBreak(s []byte, i *int) bool {
...
 
-// Set an emitter error and return false.
-func (emitter *Emitter) setEmitterError(problem string) bool {
-	emitter.ErrorType = EMITTER_ERROR
-	emitter.Problem = problem
-	return false
+// Construct emitter error.
+func emitterError(problem string) error {
+	return errors.New(problem)
 }
```
ccoVeille: I would expect this method to be removed.

And all calls to this method to be replaced by direct call to errors.New

internal/libyaml/writer.go
```diff
@@ -22,27 +22,22 @@
...
-	emitter.Problem = problem
-	return false
-}
+import "fmt"
 
 // Flush the output buffer.
-func (emitter *Emitter) flush() bool {
+func (emitter *Emitter) flush() error {
 	if emitter.write_handler == nil {
 		panic("write handler not set")
```
SVilgelm, line 30: Just a comment. I think we need to get rid of the panic where it is possible and return error instead. But this is going to be a huge work, and should be done after all active PRs are merged.

SVilgelm: https://github.com/yaml/go-yaml/issues/167

internal/libyaml/emitter.go
```diff
@@ -902,77 +902,76 @@ func (emitter *Emitter) emitNode(event *Event,
 	case MAPPING_START_EVENT:
 		return emitter.emitMappingStart(event)
 	default:
-		return emitter.setEmitterError(
-			fmt.Sprintf("expected SCALAR, SEQUENCE-START, MAPPING-START, or ALIAS, but got %v", event.Type))
+		return fmt.Errorf("expected SCALAR, SEQUENCE-START, MAPPING-START, or ALIAS, but got %v", event.Type)
```
SVilgelm, line 905: nit: `EventType` has `func String()` function, so `%s` can be used in an error message instead `#v`, which uses reflect to get a type.
```suggestion
		return fmt.Errorf("expected SCALAR, SEQUENCE-START, MAPPING-START, or ALIAS, but got %s", event.Type)
```

internal/libyaml/emitter.go
```diff
@@ -1162,135 +1161,137 @@ func (emitter *Emitter) processHeadComment() bool {
...
-	return true
+	return nil
 }
 
 // Check if a %YAML directive is valid.
-func (emitter *Emitter) analyzeVersionDirective(version_directive *VersionDirective) bool {
+func (emitter *Emitter) analyzeVersionDirective(version_directive *VersionDirective) error {
 	if version_directive.major != 1 || version_directive.minor != 1 {
-		return emitter.setEmitterError("incompatible %YAML directive")
+		return errors.New("incompatible %YAML directive")
```
SVilgelm, line 1224: nit: Should it be `YAML` instead of `%YAML`?

```suggestion
		return errors.New("incompatible YAML directive")
```

SVilgelm: I was wrong

internal/libyaml/emitter.go
```diff
@@ -902,77 +902,76 @@ func (emitter *Emitter) emitNode(event *Event,
...
-			fmt.Sprintf("expected SCALAR, SEQUENCE-START, MAPPING-START, or ALIAS, but got %v", event.Type))
+		return fmt.Errorf("expected SCALAR, SEQUENCE-START, MAPPING-START, or ALIAS, but got %v", event.Type)
 	}
 }
 
 // Expect ALIAS.
-func (emitter *Emitter) emitAlias(event *Event) bool {
-	if !emitter.processAnchor() {
-		return false
+func (emitter *Emitter) emitAlias(event *Event) error {
```
ccoVeille, line 910: Just noticed (with linter) that the `event` provided as argument is not used. Is that normal @inteon ?

ingydotnet: @ccoVeille can you make an issue of this?

ccoVeille: I wanted to wait for Tim feedback first, and then create it.

But I can so I did

- #220

inteon: Indeed, it seems like the event is unused. I'm not sure whether this is an issue, the fields in the Event object might just not be required?

PR #152, opened 2025-10-22
Use go errors for parser

inteon:
instead of returning booleans and storing the error message in the parser, we now return the error value directly.

### DIFF
--- a/internal/libyaml/parser.go
+++ b/internal/libyaml/parser.go
@@ -718,7 +718,9 @@ func (parser *Parser) parseBlockSequenceEntry(event *Event, first bool) error {
 		mark := token.EndMark
 		prior_head_len := len(parser.HeadComment)
 		parser.skipToken()
-		parser.splitStemComment(prior_head_len)
+		if err := parser.splitStemComment(prior_head_len); err != nil {
+			return err
+		}
 		if err := parser.peekToken(&token); err != nil {
 			return err
 		}
@@ -766,7 +768,9 @@ func (parser *Parser) parseIndentlessSequenceEntry(event *Event) error {
 		mark := token.EndMark
 		prior_head_len := len(parser.HeadComment)
 		parser.skipToken()
-		parser.splitStemComment(prior_head_len)
+		if err := parser.splitStemComment(prior_head_len); err != nil {
+			return err
+		}
 		if err := parser.peekToken(&token); err != nil {
 			return err
 		}


internal/libyaml/scanner.go
```diff
@@ -2820,36 +2915,42 @@ func (parser *Parser) scanPlainScalar(token *Token) bool {
...
 
 	var start_mark Mark
 	var text []byte
 
 	for peek := 0; peek < 512; peek++ {
-		if parser.unread < peek+1 && !parser.updateBuffer(peek+1) {
-			break
+		if parser.unread < peek+1 {
+			if parser.updateBuffer(peek+1) != nil {
+				break
```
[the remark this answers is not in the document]
inteon: The break matches the original behavior.

internal/libyaml/scanner.go
```diff
@@ -2910,8 +3011,10 @@ func (parser *Parser) scanComments(scan_mark Mark) bool {
 
 	peek := 0
 	for ; peek < 512; peek++ {
-		if parser.unread < peek+1 && !parser.updateBuffer(peek+1) {
-			break
+		if parser.unread < peek+1 {
+			if parser.updateBuffer(peek+1) != nil {
+				break
```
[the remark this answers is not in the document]
inteon: The break matches the original behavior.

cmd/go-yaml/parser.go
```diff
@@ -29,10 +29,9 @@ func (p *Parser) Next() (*Token, error) {
 	}
 
 	var yamlToken libyaml.Token
-	if !p.parser.Scan(&yamlToken) {
-		if p.parser.ErrorType != libyaml.NO_ERROR {
-			return nil,
-				fmt.Errorf("parser error: %v", p.parser.Problem)
+	if err := p.parser.Scan(&yamlToken); err != nil {
+		if err != io.EOF {
```
ccoVeille: Please use

```suggestion
		if !errors.Is(err, io.EOF) {
```
It would be safer if something wraps the error later

cmd/go-yaml/parser.go
```diff
@@ -29,10 +29,9 @@ func (p *Parser) Next() (*Token, error) {
...
+			return nil, fmt.Errorf("parser error: %v", err.Error())
```
ccoVeille: Always prefer error wrapping

```suggestion
			return nil, fmt.Errorf("parser error: %w", err)
```

decode.go
```diff
@@ -102,40 +101,22 @@ func (p *parser) peek() libyaml.EventType {
...
 
-func (p *parser) fail() {
-	var line int
-	if p.parser.ContextMark.Line != 0 {
-		line = p.parser.ContextMark.Line
-		// Scanner errors don't iterate line before returning error
-		if p.parser.ErrorType == libyaml.SCANNER_ERROR {
-			line++
+func (p *parser) fail(err error) {
+	parserErr, ok := err.(*libyaml.ParserError)
```
ccoVeille: Please use errors.As here, the current code is not safe against wrapped error

internal/libyaml/reader.go
```diff
@@ -27,12 +27,8 @@ import (
...
 
 // Set the reader error and return 0.
-func (parser *Parser) setReaderError(problem string, offset int, value int) bool {
-	parser.ErrorType = READER_ERROR
-	parser.Problem = problem
-	parser.ProblemOffset = offset
-	parser.ProblemValue = value
-	return false
+func (parser *Parser) setReaderError(problem string, offset int, value int) error {
+	return &ParserError{problem, 0, 0}
```
ccoVeille: I would prefer

```suggestion
	return &ParserError{Message:
problem}
```

internal/libyaml/scanner.go
```diff
@@ -599,19 +593,20 @@ func (parser *Parser) readLine(s []byte) []byte {
...
+	if parser.stream_end_produced || parser.errored {
+		return io.EOF
 	}
 
 	// Ensure that the tokens queue contains enough tokens.
 	if !parser.token_available {
-		if !parser.fetchMoreTokens() {
-			return false
+		if err := parser.fetchMoreTokens(); err != nil {
+			parser.errored = true
```
ccoVeille: Is there a need for this `errored` variable?

If we do, I would prefer an `hadError`, `errorOccured`, or variants

internal/libyaml/scanner.go
```diff
@@ -1661,20 +1677,23 @@ func (parser *Parser) scanDirective(token *Token) bool {
 
 		// Unknown directive.
 	} else {
-		parser.setScannerError("while scanning a directive",
+		return parser.setScannerError("while scanning a directive",
```
ccoVeille: I feel like this method should be renamed now.

Something like formatScannerError

Also,maybe there is no need for it to be a method maybe, a function could be enough

Previously it was setting something in the parser struct, now it doesn't any more, right?

inteon: This has to remain a method, since we are reading `parser.mark`.
I think we can simplify and refactor this function, but that should probably be done in a separate PR.

ccoVeille: You are right.

internal/libyaml/parser.go
```diff
@@ -116,38 +123,72 @@ func (parser *Parser) skipToken() {
...
+	if parser.stream_end_produced || parser.hadError || parser.state == PARSE_END_STATE {
+		return io.EOF
 	}
 
 	// Generate the next event.
-	return parser.stateMachine(event)
+	if err := parser.stateMachine(event); err != nil {
+		parser.hadError = true
+		return err
+	}
```
SVilgelm: Hi, could you please explain what is a purpose of `hadError` variable?

Correct me if I'm wrong, but `hadError` is a flag to stop the future calls of `Parser` function if the `stateMachine` returned any error?

I think that `Parse` function should return `io.EOF` only if the stream ended, if not, then it should return the error from previous call.

internal/libyaml/reader.go
```diff
@@ -27,12 +27,10 @@ import (
...
 // Set the reader error and return 0.
-func (parser *Parser) setReaderError(problem string, offset int, value int) bool {
-	parser.ErrorType = READER_ERROR
-	parser.Problem = problem
-	parser.ProblemOffset = offset
-	parser.ProblemValue = value
-	return false
+func (parser *Parser) setReaderError(problem string, offset int, value int) error {
+	return &ParserError{
+		Message: problem,
```
SVilgelm: I would suggest to add offset and value to the problem, otherwise those parameters should be deleted

internal/libyaml/scanner.go
```diff
@@ -599,19 +593,20 @@ func (parser *Parser) readLine(s []byte) []byte {
...
 // Scan gets the next token.
-func (parser *Parser) Scan(token *Token) bool {
+func (parser *Parser) Scan(token *Token) error {
 	// Erase the token object.
 	*token = Token{} // [Go] Is this necessary?
 
 	// No tokens after STREAM-END or error.
-	if parser.stream_end_produced || parser.ErrorType != NO_ERROR {
-		return true
+	if parser.stream_end_produced || parser.hadError {
```
SVilgelm: same question here

internal/libyaml/reader.go
```diff
@@ -27,12 +27,10 @@ import (
...
+	}
 }
```
ccoVeille: I would remove this method and return &ParserError directly wherever setReaderError is called

internal/libyaml/parser.go
```diff
@@ -116,38 +123,72 @@ func (parser *Parser) skipToken() {
...
-	parser.ErrorType = PARSER_ERROR
-	parser.Problem = problem
-	parser.ProblemMark = problem_mark
-	return false
+func (parser *Parser) setParserError(problem string, problem_mark Mark) error {
+	if len(problem) == 0 {
+		problem = "unknown problem parsing YAML content"
+	}
+	return &ParserError{problem, problem_mark.Line, problem_mark.Column}
 }
```
ccoVeille: I would replace this method by a constructor.

```suggestion
func newParserError(problem string, problem_mark Mark) error {
	if len(problem) == 0 {
		problem = "unknown problem parsing YAML content"
	}
	return &ParserError{problem, problem_mark.Line, problem_mark.Column}
}
```

But then it means that there is another way to consider this

Remove the whole method, do not add a function, replace all calls to setParserError to use &ParserError{} directly, and move the len(problem) to ParserError.Error method.

internal/libyaml/parser.go
```diff
@@ -116,38 +123,72 @@ func (parser *Parser) skipToken() {
...
+	var column int
+	if context_mark.Column != 0 {
+		column = context_mark.Column
+	} else if problem_mark.Column != 0 {
+		column = problem_mark.Column
+	}
+	if len(problem) == 0 {
+		problem = "unknown problem parsing YAML content"
+	}
+	return &ParserError{problem, line, column}
```
ccoVeille:
```suggestion
	return &ParserError{problem, line, column}
```

See my suggestion here

https://github.com/yaml/go-yaml/pull/152#discussion_r2482163587

Also, please use field names


```suggestion
	return &ParserError{Message: problem, Line: line, Column: column}
```

internal/libyaml/parser.go
```diff
@@ -1001,17 +1031,17 @@ func (parser *Parser) parseFlowSequenceEntry(event *Event, first bool) bool {
...
 // flow_sequence_entry  ::= flow_node | KEY flow_node? (VALUE flow_node?)?
 //
 //	*** *
-func (parser *Parser) parseFlowSequenceEntryMappingKey(event *Event) bool {
-	token := parser.peekToken()
-	if token == nil {
-		return false
+func (parser *Parser) parseFlowSequenceEntryMappingKey(event *Event) error {
+	var token *Token
+	if err := parser.peekToken(&token); err != nil {
```
ccoVeille, line 1016: I'm annoyed by this pattern

I would use this

```suggestion
	token, err := parser.peekToken()
	if err != nil {
```

See https://github.com/yaml/go-yaml/pull/152#discussion_r2482192482

internal/libyaml/scanner.go
```diff
@@ -874,14 +882,14 @@ func (parser *Parser) fetchNextToken() (ok bool) {
...
 
 	// If we don't determine the token type so far, it is an error.
-	return parser.setScannerError(
+	return parser.formatScannerError(
 		"while scanning for the next token", parser.mark,
 		"found character that cannot start any token")
 }
 
-func (parser *Parser) simpleKeyIsValid(simple_key *SimpleKey) (valid, ok bool) {
+func (parser *Parser) simpleKeyIsValid(simple_key *SimpleKey) (valid bool, err error) {
```
ccoVeille, line 902: What about this

```suggestion
func (parser *Parser) validateSimpleKey(simple_key *SimpleKey) (err error) {
```

And return an unexported sentinel when the simple key is not valid.

The catch the sentinel with errors.Is

That would be the idiomatic Go way

internal/libyaml/parser.go
```diff
@@ -65,12 +68,16 @@ import (
...
+		if err := parser.fetchMoreTokens(); err != nil {
+			return err
+		}
 	}
+
+	token := &parser.tokens[parser.tokens_head]
+	parser.unfoldComments(token)
+	*out = token
 	return nil
 }
```
ccoVeille, line 80: Please consider something simpler

```suggestion
func (parser *Parser) peekToken() (*Token, error {
	if !parser.token_available {
		if err := parser.fetchMoreTokens(); err != nil {
			return nil, err
		}
 	}

	token := &parser.tokens[parser.tokens_head]
	parser.unfoldComments(token)
 	return token, nil
 }
```

See https://github.com/yaml/go-yaml/pull/152#discussion_r2482172216

decode.go
```diff
@@ -102,40 +102,22 @@ func (p *parser) peek() libyaml.EventType {
...
-		if p.parser.ErrorType == libyaml.SCANNER_ERROR {
-			line++
+func (p *parser) fail(err error) {
+	var parserErr *libyaml.ParserError
+	if errors.As(err, &parserErr) {
+		err = &ParserError{
+			Message: parserErr.Message,
+			Line:    parserErr.Line,
+			Column:  parserErr.Column,
 		}
```
SVilgelm: Why do you convert from internal ParserError to an external one? Anyway it will be converted to yamlError and then panicked.

I think you can get rid of `fail` method completely and use `fail` function, because you already removed all logic

SVilgelm (commented):
Really huge work has been done here, have few questions.

SVilgelm:
One more, could you please enable `nilnesserr` linter, seems they PR is the best use case for that linter :)

internal/libyaml/parser.go
ccoVeille: Yes, I'm also worried here. I wasn't sure, so I didn't report it. But I'm not the only one apparently. So it means my first impression might be right and there is a problem here

decode.go
ccoVeille: Same question

internal/libyaml/parser.go
SVilgelm: I had same idea, but I think it would be better to do in a separate PR

internal/libyaml/parser.go
inteon: I updated the code and now save the last error and return that error if it is set.

internal/libyaml/scanner.go
inteon: I think this is ok, the function returns true/ false depending on the key being valid and returns an error when it encounters are parsing failure.

internal/libyaml/parser.go
ccoVeille: Not really, because this PR changes the cods to do this

```diff
-func (parser *Parser) peekToken() *Token { 
+func (parser *Parser) peekToken(out **Token) error {
```

While this could be used

```diff
-func (parser *Parser) peekToken() *Token { 
+func (parser *Parser) peekToken() (*Token, error) {
```

internal/libyaml/scanner.go
ccoVeille: OK then 👍

SVilgelm (changes requested):
see the comments above

internal/libyaml/parser.go
inteon: The issue with the second example is that it is very easy to create this bug: https://go.dev/play/p/t9VU5sdMGHV
That is why I opted to go with the first function signature instead.

ccoVeille: It's a matter of scope, I understand why you want that. I'm unsure there is a need for that. It's a matter of linting I feel.

I don't find it idiomatic also.

But I won't disagree if we go in that direction.

I would prefer to benefit from the changes you are introducing 🎉

internal/libyaml/parser.go
ccoVeille: What do you think here ?

internal/libyaml/parser.go
ccoVeille: This one also needs to be addressed, or at least replied.

ccoVeille:
Now #177 is merged, would you restart working on this one @inteon ?

internal/libyaml/reader.go
ccoVeille: This comment is still relevant @inteon

internal/libyaml/parser.go
inteon: has been solved in https://github.com/yaml/go-yaml/pull/177

internal/libyaml/parser.go
inteon: has been solved in https://github.com/yaml/go-yaml/pull/177 + renamed the function to `formatParserError` in this PR.

internal/libyaml/reader.go
inteon: fixed

decode.go
inteon: removed the conversion

internal/libyaml/reader.go
inteon: fixed in https://github.com/yaml/go-yaml/pull/177

internal/libyaml/reader.go
inteon: fixed in https://github.com/yaml/go-yaml/pull/177

ingydotnet:
Where is this at with the new codebase?

I can try to get Claude to resolve things.

Thoughts?

inteon:
> Where is this at with the new codebase?
> 
> I can try to get Claude to resolve things.
> 
> Thoughts?

I rebased the code on master.

ingydotnet:
@ccoVeille @SVilgelm I think this is gtg. Needs a team review.

internal/libyaml/parser.go
```diff
@@ -116,41 +121,47 @@ func (parser *Parser) skipToken() {
...
 
 // Parse gets the next event.
-func (parser *Parser) Parse(event *Event) bool {
+func (parser *Parser) Parse(event *Event) error {
 	// Erase the event object.
 	*event = Event{}
 
+	if parser.lastError != nil {
+		return parser.lastError
```
ccoVeille, line 129: The previous behavior was returning true on error.

So here, it means that we are somehow changing the behavior slightly. It's OK as it's the internal parser, but it can have consequences.

I feel like we have multiple options here when last error is non nil:
- return the lastError (code currently in the PR)
- return nil when last error (somehow equivalent to the old code retuning true)
- return a sentinel error such as "cannot parse as an error was previously reported"

I would like to avoid returning the lastError it could be confusing, also it's strange to store the last error in a struct. Especially, now the error is returned all along the stack.

I would prefer a boolean such as "parserFailed = true"

But I feel like you had this and @SVilgelm asked you to change (and I approved it 😅)

https://github.com/yaml/go-yaml/pull/152#discussion_r2482102936

I'm sorry to bring this again on top

internal/libyaml/reader.go
```diff
@@ -104,21 +102,20 @@ func (parser *Parser) updateRawBuffer() bool {
 	if err == io.EOF {
```
ccoVeille, line 102: Adding an errors.Is won't hurt

internal/libyaml/reader.go
```diff
@@ -28,14 +28,12 @@ import (
 	"io"
 )
 
-// Set the reader error and return 0.
-func (parser *Parser) setReaderError(problem string, offset int, value int) bool {
-	parser.Err = ReaderError{
+func formatReaderError(problem string, offset int, value int) error {
```
ccoVeille, line 31: As I previously stated in a comment a few weeks ago, I would remove this method and replace calls to this method to return the error that is currently in this method

Here I'm nitpicking, but I feel this method is no longer needed

internal/libyaml/parser.go
```diff
@@ -1169,7 +1174,7 @@ func (parser *Parser) parseFlowMappingValue(event *Event, empty bool) bool {
 }
 
 // Generate an empty scalar event.
-func (parser *Parser) processEmptyScalar(event *Event, mark Mark) bool {
+func (parser *Parser) processEmptyScalar(event *Event, mark Mark) error {
```
ccoVeille, line 1177: It's a bit strange to have such signature when method cannot return an error

internal/libyaml/parser.go
```diff
@@ -116,41 +121,47 @@ func (parser *Parser) skipToken() {
...
+		return err
+	}
+
+	return nil
 }
 
-// Set parser error.
-func (parser *Parser) setParserError(problem string, problem_mark Mark) bool {
-	parser.Err = ParserError{
+func formatParserError(problem string, problem_mark Mark) error {
```
ccoVeille, line 146: See https://github.com/yaml/go-yaml/pull/152#discussion_r2643782726

I would remove that method

internal/libyaml/yaml.go
```diff
@@ -575,11 +575,9 @@ type AliasData struct {
 // Parser structure holds all information about the current
 // state of the parser.
 type Parser struct {
-	// Err handling
-	Err error
+	lastError error
```
ccoVeille, line 578: See https://github.com/yaml/go-yaml/pull/152/changes#r2643754224

I find strange to store an error here

Especially if we consider it's just for storing a state

internal/libyaml/parser.go
ccoVeille: I have reconsidered my position, and I dislike the idea of storing an error in the struct for that. I suggest to set a state (like hadError) and return a sentinel error

See here https://github.com/yaml/go-yaml/pull/152/changes#r2643754224

ingydotnet:
@ccoVeille Do you want me to add a commit to address these things? Do you want to? Should we wait on @inteon?
Should we merge this and address your concerns in a follow up PR?

I feel like this is close to being done.

ccoVeille:
I'm with family, and not allowed to use computer. I manage to use my phone for now 🤪

I won't be able to work on this.

The PR can be merged like this, we can address this by opening a backlog issue.

The PR solves many historical issues.

If we compare what the PR and what are the things that I would like to change in the PR, we can compare respectively to a boulder and sand dust 😁

ingydotnet:
> I'm with family, and not allowed to use computer. I manage to use my phone for now 🤪
> 
> I won't be able to work on this.
> 
> The PR can be merged like this, we can address this by opening a backlog issue.
> 
> The PR solves many historical issues.
> 
> If we compare what the PR and what are the things that I would like to change in the PR, we can compare respectively to a boulder and sand dust 😁

I would attempt to work on your issues but I don't think I have the permissions to push back to this pull request. 
So I'll merge and open a backlog issue.

ingydotnet:
@ccoVeille Would you mind approving this so I can do it legit?

ccoVeille (approved):

PR #154, opened 2025-10-26
Optional support of Json struct tags and (Un)Marshaler interfaces as fallback

### LINKED ISSUE #153: Support json tags as fallback
It would be great if go-yaml supports the json tags as the fallback if the yaml tags are missing.

There are multiple libraries which unmarshal yaml using go-yaml into `any` type, then marshal to json and unmarshal to a structure. Those libraries were created, because the original author of go-yaml rejected to support go tags.

SVilgelm:
This PR makes Go struct tag resolution more flexible when types use json struct tags instead of yaml struct tags and ensures anonymous struct fields are inlined consistently under the fallback mode.
It also fallbacks to `json.(Un)Marshaller` interfaces if they are implemented. 
**Note**: The fallback to the interface implemented by using double conversion: `Obj -> JSON -> any -> YAML` for marshaling or `Node -> any -> JSON -> Obj` for unmarshaling. There is no more direct way to invoke interfaces, so this trade-off is required for compatibility.

Example:
```go
type Test struct {
  F string `json:foo`
}

func main() {
  text := []byte(`foo: "bar"`)
  var v Test
  d := NewDecoder(bytes.NewReader(text))
  d.FallbackToJSON(true)
  d.Decode(&v)
  fmt.Println(v.F)
  // Output:
  // bar
}
```

Fixes #153 

1. Update `getStructInfo` function and add tests
  Update struct tag parsing accept an optional fallback to the `json` struct tag and propagate that behavior through recursive calls.
    - Add a fallbackToJSON parameter to getStructInfo so callers can request that empty yaml tags be replaced by the json struct tag when present.
    - When fallbackToJSON is enabled, treat anonymous fields whose json struct tag exists (or which have no yaml struct tag but fallback is on) as inline if their type is a struct or pointer-to-struct.
    - Propagate fallbackToJSON in recursive getStructInfo calls for nested types so tag resolution remains consistent.
    - Add comprehensive tests for getStructInfo covering normal and fallback-to-json behavior.
2. Add FallbackToJSON option to encode and decoder
    - Add fallbackToJSON bool fields to internal decoder and encoder
      structs.
    - Pass fallbackToJSON into getStructInfo calls so struct tag lookup
      considers json struct tags when enabled.
    - Expose Decoder.FallbackToJSON and Encoder.FallbackToJSON methods that
      set the option on the public types.
    - Wire Decoder.Decode to transfer the public Decoder setting into the
      internal decoder instance; likewise for Encoder.
    - Add examples (tests) demonstrating Decoder.FallbackToJSON and
      Encoder.FallbackToJSON behavior.
3. Support json.(Un)Marshaler interfaces
    - implements encoder path for json.Marshaler to convert JSON output to
      YAML via an intermediate interface{} and then marshal that.
    - adds decoder.callJSONUnmarshaler to route a decoded node through
      json.Unmarshal to satisfy json.Unmarshaler implementations and report
      errors with proper node position.
    - extends decoder selection logic to try json.Unmarshaler as a fallback
      when FallbackToJSON(true) is set.
    - updates tests (yaml_test.go) to cover both encoder and decoder JSON
      fallback behavior via a TestFallbackToJSON type that tracks when
      MarshalJSON/UnmarshalJSON are invoked and verifies round-trip 
      behavior.
4. Normalize JSON
    - change callJSONUnmarshaler to return bool and collect TypeError and
      UnmarshalError details from the initial node.Decode call.
    - extract JSON marshaling/unmarshaling logic into unmarshalJSON helper.
    - add normalizeJSON to convert YAML-parsed structures into JSON-safe
      structures: convert map[any]any to map[string]any (stringifying
      non-string keys via JSON), recursively normalize arrays and maps, and
      detect duplicate keys that would conflict after conversion.
    - add tests for json.Unmarshaler support including maps with string and
      non-string keys to ensure correct normalization and error handling.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -19,6 +19,7 @@ import (
 	"encoding"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"math"
@@ -386,11 +387,11 @@ func (d *decoder) callUnmarshaler(n *Node, u Unmarshaler) (good bool) {
 }
 
 func (d *decoder) callJSONUnmarshaler(n *Node, u json.Unmarshaler) bool {
-	var v any
-
 	// First decode the node into an interface{} using the normal decoding rules.
+	var v any
 	if err := n.Decode(&v); err != nil {
-		if te, ok := err.(*TypeError); ok {
+		var te *TypeError
+		if errors.As(err, &te) {
 			d.terrors = append(d.terrors, te.Errors...)
 		} else {
 			d.terrors = append(d.terrors, &UnmarshalError{
@@ -487,9 +488,10 @@ func normalizeJSON(input any) (any, error) {
 			a[i] = jv
 		}
 		return a, nil
-	}
 
-	return input, nil
+	default:
+		return input, nil
+	}
 }
 
 func (d *decoder) callObsoleteUnmarshaler(n *Node, u obsoleteUnmarshaler) (good bool) {


yaml_test.go
```diff
@@ -0,0 +1,544 @@
...
+func (t *TestFallbackToJSON) MarshalJSON() ([]byte, error) {
+	type Alias TestFallbackToJSON
+	data, err := json.Marshal(Alias(*t))
+	t.marshaled = true
+	return data, err
+}
+
+func ExampleDecoder_FallbackToJSON() {
+	text := []byte(`---
+json_a: 42
```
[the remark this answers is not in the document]
SVilgelm: prefer to keep it, so, the actual keys will be next to each other

SVilgelm:
@ccoVeille Could you please review this PR? It is huge, but all changes are hidden behind `FallbackToJSON` option, so should be safe to merge.
The only concern I have that the PR is huge and maybe it is better to split it for better readability.

SVilgelm:
`go vet` step fails:
```
# go.yaml.in/yaml/v4
Error: ./yaml_test.go:28:5: struct field tag `just_b` not compatible with reflect.StructTag.Get: bad syntax for struct tag pair
Error: ./yaml_test.go:127:5: struct field tag `just_b` not compatible with reflect.StructTag.Get: bad syntax for struct tag pair
Error: Process completed with exit code 2.
```

What should I do in this case? I can remove a test case for "Foo \`foo\`" scenario or `govet` CI step can be removed because it is covered by `golangci-lint`

Also, I'm going to fix the commit messages to comply with the `commit check`

ccoVeille:
I was busy with personal project today.
I only reviewed your PR quickly.

There are a few things that I would have done differently, and one I'm worried about even if I didn't took the time to check because it's late here. And I don't have a computer with me.

So please bare with me, I might be wrong and/or tired 😅

I wouldn't have added a FallbackToJSON method to the decoder, but an Option pattern to a new method named DecodeWithOptions

But it would require #155 first.


The other things I'm worried about is the behavior of code when YAML and JSON tags are present. Especially when their logic differ.

```go
type Whatever struct {
 Password string `yaml:"password" json:"pwd"`
 Secret string `yaml:"-" json:"secret"`
 Bar int `json:"bar"`
 Foo string `json:"foooo"`
}
```

With the following yaml

```yaml
password: qwerty
pwd: ~
secret: danger
bar: 42
foo: 123
foooo: 456
```

I would expect to get

```go
Whatever{
 Password: "qwerty" // and not "~"
 Secret: "" // because of yaml:"-"
 Bar: 42
 Foo: "123" // not "456" because of foooo
}
```

Here I'm trying to find extreme awkward examples, but I feel like they should be considered.

I wrote this on a phone. I didn't check. But I would expect this to be discussed with many people, because once added it would be difficult, if not impossible to change later.

Please note, while I do understand the need. I don't see an urge to add this feature quickly.

But I get the idea, and I like it. It requires some thoughts, discussions, and iterations.

SVilgelm:
Thank you for reviewing.

> Here I'm trying to find extreme awkward examples, but I feel like they should be considered.

I have covered this case and created the unit tests, the `yaml` tag always takes a precedent:
- https://github.com/yaml/go-yaml/pull/154/files#diff-5afd41e7e0a70c3a6f0e1572f6938d9aacf5bc590033f7e798a0fb6f741e51ddR29
- https://github.com/yaml/go-yaml/pull/154/files#diff-5afd41e7e0a70c3a6f0e1572f6938d9aacf5bc590033f7e798a0fb6f741e51ddR53

> I wouldn't have added a FallbackToJSON method to the decoder, but an Option pattern to a new method named DecodeWithOptions

We can talk about this in #155. I am fully open for any ideas.

> Please note, while I do understand the need. I don't see an urge to add this feature quickly.

Sure, no rush at all

I still have a question about `govet` step in CI. Should I drop a test case for a scenario when a structure tag is not compatible:
```go
type Whatever struct {
 Password string `yaml:"password" json:"pwd"` // It is okay
 Secret string `secret` // <- It is not compatible, but supported by current code
}
```

Or we can drop the `govet` step in CI :) Any way it is covered by `golangci-lint`, which allows to add `//nolint:govet` comments to ignore those cases.

ccoVeille:
> Thank you for reviewing.
> 
> > Here I'm trying to find extreme awkward examples, but I feel like they should be considered.
> 
> I have covered this case and created the unit tests, the `yaml` tag always takes a precedent:
> - https://github.com/yaml/go-yaml/pull/154/files#diff-5afd41e7e0a70c3a6f0e1572f6938d9aacf5bc590033f7e798a0fb6f741e51ddR29
> - https://github.com/yaml/go-yaml/pull/154/files#diff-5afd41e7e0a70c3a6f0e1572f6938d9aacf5bc590033f7e798a0fb6f741e51ddR53

I will take a look, thanks

> > I wouldn't have added a FallbackToJSON method to the decoder, but an Option pattern to a new method named DecodeWithOptions
> 
> We can talk about this in #155. I am fully open for any ideas.

Let's continue discussion in #155

> > Please note, while I do understand the need. I don't see an urge to add this feature quickly.
> 
> Sure, no rush at all
> 
> I still have a question about `govet` step in CI. Should I drop a test case for a scenario when a structure tag is not compatible:
> ```go
> type Whatever struct {
>  Password string `yaml:"password" json:"pwd"` // It is okay
>  Secret string `secret` // <- It is not compatible, but supported by current code
> }
> ```

Could you open an issue for this?

I mean the fact `secret` is supported even if the tag is invalid.

May I challenge you here?

For me, it's normal and expected that it works.

Consider this:

```go
type Whatever struct {
  Password string `yaml:"password" json:"pwd"` // It is okay
  Secret string `secret2` // <- what does current code?
}
 ```

With 

```yaml
password: qwerty
secret2: danger
```

I would expect a "bug" only if it would be decoded with 

```go
var _ = Whatever{Password: "qwerty", Secret: "danger"}
```

I mean I would expect the lib to fall back on default behavior of trying to match the name of field in the struct, as if no struct tag were provided.

What does json.Unmarshal with this on the same struct ?

```json
{
  "pwd": "qwerty",
  "secret": "danger"
}
```

Does it fail because tags are invalid or does it decode Secret=danger ?

Here again, I didn't check what you reported, I based my reply on your comment and my thoughts.

**Unless you agree with me, please reply to this in the issue you will create about it.**


> Or we can drop the `govet` step in CI :) Any way it is covered by `golangci-lint`, which allows to add `//nolint:govet` comments to ignore those cases.

The govet thing is a good thing. The govet is automatic with go test. I get the point of being able to write invalid code to test an invalid behavior. I faced it and removed govet by using golangci-lint too.

Please note to disable govet in go test, `go test -vet=off` can be used

https://pkg.go.dev/cmd/go#hdr-Testing_flags

SVilgelm:
> The govet is automatic with go test

not really:
```shell
25-10-28 7:54 % go test ./...                                                                                                                                               
ok      go.yaml.in/yaml/v4      2.453s                                                                                                                                      
?       go.yaml.in/yaml/v4/cmd/go-yaml  [no test files]                                                                                                                     
?       go.yaml.in/yaml/v4/internal/libyaml     [no test files]                                                                                                             
ok      go.yaml.in/yaml/v4/internal/testutil/assert     (cached)                                                                                                            
ok      go.yaml.in/yaml/v4/yts  0.619s

% go vet ./...                                                                                                                                                
# go.yaml.in/yaml/v4                                                                                                                                                        
# [go.yaml.in/yaml/v4]                                                                                                                                                      
./yaml_test.go:28:5: struct field tag `just_b` not compatible with reflect.StructTag.Get: bad syntax for struct tag pair                                                    
./yaml_test.go:127:5: struct field tag `just_b` not compatible with reflect.StructTag.Get: bad syntax for struct tag pair                                                   
```

As you can see, I ran `go test` locally and it does not fail, but `go vet` fails. Okay, let me create a separate issue for this case to have a discussion in #157

decode.go
```diff
@@ -383,6 +385,113 @@ func (d *decoder) callUnmarshaler(n *Node, u Unmarshaler) (good bool) {
 	}
 }
 
+func (d *decoder) callJSONUnmarshaler(n *Node, u json.Unmarshaler) bool {
+	var v any
+
+	// First decode the node into an interface{} using the normal decoding rules.
+	if err := n.Decode(&v); err != nil {
+		if te, ok := err.(*TypeError); ok {
```
ccoVeille: Please use errors.As here

Using type assertion on error is anti-pattern since Go 1.13

adding errcheck linter could be considered in golangci-lint

But I'm worried it might need to be addressed in another PR as it might imply to fix more than this.

So here using errors.As would be preferred

yaml.go
```diff
@@ -113,6 +114,13 @@ func (dec *Decoder) KnownFields(enable bool) {
 	dec.knownFields = enable
 }
 
+// FallbackToJSON enables using the "json" struct tags as a fallback
+// when no "yaml" struct tag is present.
+// If enabled, the anonymous struct fields are also considered for inlining.
+func (dec *Decoder) FallbackToJSON(enable bool) {
```
ccoVeille, line 120: Except this that I would prefer not to be added here, but via an Option passed as a variadic, I'm fine with the changes

SVilgelm: The options are not implemented yet. I'll change this logic to Option when they are ready for sure

decode.go
SVilgelm: good catch, thank you!

ingydotnet:
See #212

ingydotnet:
@SVilgelm will you have time to adapt this now that Options are a merged thing?

yaml.go
ingydotnet: Options are ready for a while now.

PR #158, opened 2025-10-28
Add cases for non-compliant struct tag and remove `go vet` check from CI and Makefile

### LINKED ISSUE #157: go vet fails on a test case for an invalid struct tag
```go
type Whatever struct {
  F string `yaml:"foo" // the correct tag
  B string `bar`       // incorrect tag
}
```

The logic of `go-yaml` is: https://github.com/yaml/go-yaml/blob/main/yaml.go#L645-L648
```
           * Get a `yaml` tag
           |                |
     * If empty           * Not empty
     |                    |
* Check if the tag line   * Use the tag
is not empty and does 
not contain `:` 
character, then use the
whole string                   
```

The issue, that it is impossible to create a tests case, since `go vet` check fails with an error:
```go
struct field tag `just_b` not compatible with reflect.StructTag.Get: bad syntax for struct tag pair
```

What should we do here?
1. Ignore this scenario and don't create the unit tests
2. Remove the logic to use the whole tag string: `tag = string(field.Tag)`
3. Remove `go vet` in CI, because it does not support `//nolint:govet` directives and its functionality fully covered by `golangci-lint`
4. Any other suggestions?

I created unit tests in my repo to show how it works:

```go
	t.Run("invalid tag yaml", func(t *testing.T) {
		type Whatever struct {
			F string `yaml:"foo"` // the correct tag
			B string `bar`        // incorrect tag
		}
		v := Whatever{
			F: "value1",
			B: "value2",
		}
		data, err := yamlV4.Marshal(v)
		t.Logf("yaml: %s", string(data))
		// Output:
		// yaml: foo: value1
		//       bar: value2
		require.NoError(t, err)
		var v2 Whatever
		data = []byte(`foo: value1
bar: value2
`)
		err = yamlV4.Unmarshal(data, &v2)
		require.NoError(t, err)
		require.Equal(t, v, v2)
		// no errors
	})
	t.Run("invalid tag json marshal", func(t *testing.T) {
		type Whatever struct {
			F string `json:"foo"` // the correct tag
			B string `bar`        // incorrect tag
		}
		v := Whatever{
			F: "value1",
			B: "value2",
		}
		data, err := json.Marshal(v)
		t.Logf("json: %s", string(data))
		// Output:
		// json: {"foo":"value1","B":"value2"}

		require.NoError(t, err)
		var v2 Whatever
		data = []byte(`{"foo":"value1","bar":"value2"}`)
		err = json.Unmarshal(data, &v2)
		require.NoError(t, err)
		require.Equal(t, v, v2)
		// Error: Not equal:
		//        expected: yaml_benchmarks.Whatever{F:"value1", B:"value2"}
		//        actual  : yaml_benchmarks.Whatever{F:"value1", B:""}
	})
```

SVilgelm:
Add test cases to decode_test.go and encode_test.go that cover an edge case described in issue #157 where a struct field uses an incorrect tag syntax (a bare string like `bar` without the `yaml:` prefix) alongside a correct `yaml:"foo"` tag. The new tests assert that the library continues to accept and round-trip values when one field uses the supported-but-nonstandard tag form.

This clarifies expected behavior and prevents regressions for legacy struct tags that some users rely on.

Remove the redundant go vet invocation from both the GitHub Actions
workflow and the GNUmakefile. The CI workflow no longer runs
"go vet ./..." and the "vet" target and its invocation in the makefile
are deleted.

Disable the Go vet tool when running test and helper commands
to avoid vet-related failures blocking test execution. Update the main
test target so its result parsing is not affected by vet output.

Fixes #157

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -90,11 +90,7 @@ fmt: $(GOLANGCI-LINT-VERSIONED)
 	$< fmt ./...
 
 lint: $(GOLANGCI-LINT-VERSIONED)
-	$< run
-
-fumpt: $(GO)
-	@go install mvdan.cc/gofumpt@latest
-	gofumpt -l -w $(GO-FILES)
+	$< run ./...
 
 cli: $(CLI-BINARY)
 


decode_test.go
```diff
@@ -960,6 +960,19 @@ ba?r: a?bc
...
+	// issue #157
+	{
+		`foo: foo
+bar: bar`,
+		struct {
+			F string `yaml:"foo"` // the correct tag, because it has `yaml` prefix
+			B string `bar`        //nolint:govet // the incorrect tag, but supported
+		}{
+			F: "foo",
+			B: "bar",
```
ccoVeille: Please add something like


```suggestion
			B: "bar", // TODO: fix this see #157
```

decode_test.go
```diff
@@ -960,6 +960,19 @@ ba?r: a?bc
 			"ba?r": "a?bc",
 		},
 	},
+
+	// issue #157
+	{
+		`foo: foo
+bar: bar`,
```
ccoVeille: I'm nitpicking here.

I wouldn't use this as it creates confusion.

Maybe this


```suggestion
		`foo: abc
bar: def`,
```

Because otherwise it's unclear if the foo is the key or value.

It's clearly the key, but a refactoring could lead to change this without being detected

encode_test.go
```diff
@@ -582,6 +582,18 @@ var marshalTests = []struct {
...
+	// issue #157
+	{
+		struct {
+			F string `yaml:"foo"` // the correct tag, because it has `yaml` prefix
+			B string `bar`        //nolint:govet // the incorrect tag, but supported
+		}{
+			F: "foo",
+			B: "bar",
+		},
+		"foo: foo\nbar: bar\n",
```
ccoVeille: Please add a comment here too

ccoVeille (commented):
Some comments.

Please remove the go vet jobs from GHA and the govet in Makefile entry

You might also have to update go test calls use -vet=off as flags with GHA and Makefile, the govet linters in golangci-lint is enough and supports `//nolint:govet` as you said.

decode_test.go
SVilgelm: sure, thank you for suggestion

encode_test.go
SVilgelm: sure

ccoVeille:
> Some comments.
> 
> Please remove the go vet jobs from GHA and the govet in Makefile entry
> 
> You might also have to update go test calls use -vet=off as flags with GHA and Makefile, the govet linters in golangci-lint is enough and supports `//nolint:govet` as you said.

About this ☝️

It reminds me your comment in the other issue 👇

> > The govet is automatic with go test
>
> not really:
>
> (…)
> 
> As you can see, I ran `go test` locally and it does not fail, but `go vet` fails. 
>
> _Originally posted by @SVilgelm in https://github.com/yaml/go-yaml/pull/154#issuecomment-3457026903_

I'm unsure how you got no errors as using go test uses go vet by default with Go 1.24+

https://go.dev/doc/go1.24#vet

SVilgelm:
I don't know, 
```shell
25-10-28 14:25 % go test ./...                                                                                                                                              
ok      go.yaml.in/yaml/v4      2.087s
?       go.yaml.in/yaml/v4/cmd/go-yaml  [no test files]
?       go.yaml.in/yaml/v4/internal/libyaml     [no test files]
ok      go.yaml.in/yaml/v4/internal/testutil/assert     0.615s
ok      go.yaml.in/yaml/v4/yts  1.099s

25-10-28 14:25 % go version                                                                                                                                                 
go version go1.24.9 darwin/arm64
```


Okay, I checked the documentation for `test` cmd, it says:
> go test runs go vet on the package                                                                                                       
and its test source files to identify significant problems.
...
To disable the running of go vet, use the -vet=off flag. To run all                                                                                                         
checks, use the -vet=all flag.

Seems this particular check is disabled for test cmd. The `vet` tool is like `golangci-lint`, it also runs multiple analyzers.
To prove this idea, I run `go test -vet=all` and it failed.

SVilgelm:
I got an idea, maybe instead of removing `vet` target from Makefile, I should change it to invoke `golangci-lint`?

ccoVeille:
You could indeed this could do the job

`golangci-lint run --enable-only govet`

But, I'm unsure here. At least you would only have it in makefile and not the GHA.

I would simply remove it. But it doesn't mean I'm right about it. I get your point, I'm unsure what is the best.

As `make lint` is also available, I'm unsure there is a need for a `make vet`. I mean why having a `vet` target that doesn't call `go vet`, and call something that is also available and checked by `lint` target

If it is removed, there might be a need to update contributions guide.

Also, please consider adding `-vet=off` to `go test` in `test` target

SVilgelm:
Looks like the the `lint` target is not merged yet: `make: *** No rule to make target `lint'`

SVilgelm:
> Also, please consider adding -vet=off to go test in test target

sure, will add

GNUmakefile
```diff
@@ -51,14 +51,14 @@ count ?= 1
...
-	go test$(if $v, -v)
+	go test$(if $v, -v) -vet=off
 
 test-data: $(YTS-DIR)
 
 test-all: test test-yts-all
 
 test-yts: $(TEST-DEPS)
-	go test$(if $v, -v) ./yts -count=$(count)
+	go test$(if $v, -v) -vet=off ./yts -count=$(count)
```
ccoVeille: I don't think there is a need to use -vet=off for yts tests suite.

GNUmakefile
```diff
@@ -92,7 +89,7 @@ $(YTS-DIR):
 define yts-pass-fail
 ( \
   result=.cache/local/tmp/yts-test-results; \
-  go test ./yts -count=1 -v | \
+  go test -vet=off ./yts -count=1 -v | \
```
ccoVeille: See 

https://github.com/yaml/go-yaml/pull/158#discussion_r2471280954

GNUmakefile
SVilgelm: done

GNUmakefile
SVilgelm: done

ccoVeille:
> Looks like the the `lint` target is not merged yet: `make: *** No rule to make target `lint'`

Indeed #131 is not merged yet.

SVilgelm:
@ccoVeille @ingydotnet I rebased this PR and I think it is ready

ingydotnet:
> @ccoVeille @ingydotnet I rebased this PR and I think it is ready

@SVilgelm thanks. I'm out of tuits for this until at least Friday...

GNUmakefile
```diff
@@ -31,7 +31,7 @@ MAKES-REALCLEAN := $(dir $(YTS-DIR))
 
 # Setup and include go.mk and shell.mk:
 GO-FILES := $(shell find -not \( -path ./.cache -prune \) -name '*.go' | sort)
-GO-CMDS-SKIP := test fmt vet
```
ccoVeille, line 34: Please check the behavior of this change.

I feel like it would enable go vet because the Makefile would make it available

https://github.com/makeplus/makes/blob/main/go.mk

I wouldn't change this.

This variable is here to allow to overwrite make targets that are available by default via makeplus.

SVilgelm: Oh I see, let me revert it

SVilgelm: Could you run the CI please?

ccoVeille: Done

.github/workflows/go.yaml
```diff
@@ -44,9 +44,7 @@ jobs:
       uses: actions/setup-go@v6
       with:
         go-version: ${{ matrix.go-versions }}
-    - name: Run go vet
-      run: go vet ./...
     - name: Run yaml-test-suit tests
       run: make test-yts
     - name: Run go test
-      run: GO111MODULE=on go test -v -race .
+      run: GO111MODULE=on go test -vet=off -v -race .
```
ccoVeille: The GO111MODULE=on is no longer needed 😅

```suggestion
      run: go test -vet=off -v -race .
```

ccoVeille: Also, I feel I might have forgot to update this like here when I worked on moving libyaml to internal

It should be this here

```suggestion
      run: go test -vet=off -v -race ./...
```

GNUmakefile
```diff
@@ -62,7 +62,7 @@ count ?= 1
 
 # Test rules:
 test: $(GO-DEPS)
-	go test$(if $v, -v)
+	go test$(if $v, -v) -vet=off
```
ccoVeille: See

https://github.com/yaml/go-yaml/pull/158#discussion_r2483171983

I feel like it should be


```suggestion
	go test$(if $v, -v) -vet=off ./...
```

GNUmakefile
ccoVeille, line 1: The same way we add ./... to `go test`, I would add the ./... to `make lint`, because `make fmt ` has it when calling golangci-lint

SVilgelm: done, also remove `fumpt` target

ccoVeille:
Let's merge and continue with other PR.

It's a test we add and some addition to makefile

PR #159, opened 2025-10-29
Refactor yts tests to make them more reliable

SVilgelm:
- Change loadKnownFailingTests to accept testing.TB, return a  map[string]int, and fail fast on file open errors. Track run counts  for each known failing test to detect tests that did not run.
- Introduce knownFailingButPassedTests to record known-failing tests that unexpectedly pass.
- Update shouldSkipTest to accept testing.TB, mark helper, record run counts, register a cleanup to detect known-failing-but-passed cases, and simplify skip/log messages. Honor RUNALL and RUNFAILING env flags.
- Initialize knownFailingTests at test start and add two verification subtests: CheckKnownFailingTests (ensures listed known failures actually ran) and CheckKnownFailingButPassedTests (fails if a known-failing test passed).
- Improve runTestsInDir: skip dotfiles, compute relative test names for subtests, and add helper marker. Make error messages use %q for paths.
- Simplify and clarify runTest logging and error messages.
- Change JSONComparisonTest to compare the values as JSON text to avoid the test failures, because Json unmarshals all numbers as float64, while YAML preserves integer types when possible.

These changes make the test suite more reliable and informative by failing early on configuration issues, tracking which known-failing tests run, and flagging regressions where failures unexpectedly stop occurring.

### DIFF
--- a/yts/test_suite_test.go
+++ b/yts/test_suite_test.go
@@ -176,7 +176,7 @@ func normalizeLineEndings(s string) string {
 
 func runTest(t *testing.T, testPath string) {
 	t.Helper()
-	
+
 	// Read test description
 	testDescription := string(readFile(t, false, testPath, "==="))
 	if testDescription == "" {
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -7,6 +7,7 @@ linters:
   - misspell
   - nolintlint
   - staticcheck
+  - thelper
   disable:
   - errcheck
   - ineffassign


SVilgelm:
Maybe it is landed in the spam folder? Email should be from `sergey@vilgelm.com` address with `CNCF slack` subject
or just send an invite to me

ingydotnet:
> Maybe it is landed in the spam folder? Email should be from `sergey@vilgelm.com` address with `CNCF slack` subject or just send an invite to me

Yep. Was in spam...

ingydotnet:
@carloslima would like to see your review here.

yts/test_suite_test.go
```diff
@@ -11,64 +12,108 @@ import (
 	"go.yaml.in/yaml/v4"
 )
 
-var knownFailingTests = loadKnownFailingTests()
+const (
+	testDir        = "testdata/data-2022-01-17"
+	knownTestsFile = "known-failing-tests"
```
ccoVeille, line 19: I would use go:embed then

yts/test_suite_test.go
```diff
@@ -11,64 +12,108 @@ import (
...
+	t.Helper()
+
 	entries, err := os.ReadDir(dirPath)
 	if err != nil {
-		t.Fatalf("Failed to read directory %s: %v", dirPath, err)
+		t.Fatalf("Failed to read directory %q: %v", dirPath, err)
 	}
 
 	for _, entry := range entries {
+		if strings.HasPrefix(entry.Name(), ".") {
```
ccoVeille: Please add a comment here to explain why you skip this entry

Something like "skip dotfiles"

yts/test_suite_test.go
```diff
@@ -91,8 +136,6 @@ func getEvents(in []byte) (string, error) {
 }
 
 func runTest(t *testing.T, testPath string) {
-	t.Helper()
```
ccoVeille, line 93: I didn't launch the tests, but the removal of this line is strange to me

This is a test helper, no?

Error would be reported where they fail.

But maybe it's the purpose of the changes.

Could you add t.Helper linter via golangci-lint and add a `//nolint:thelper // explanation `

yts/test_suite_test.go
ccoVeille: Please note, I don't expect a go:embed.

I'm suggesting it and I'm looking for your feedback

I'm fine with current implementation

SVilgelm: I used `embed` as a first iteration, but ran into some issues, probably due to a lack of understanding of the code.

Now I understand it better and actually got it working :)

It seems to make traversing the testdata directory easier because it doesn't include hidden directories and symbolic links, so it essentially automatically filters out unnecessary folders.

SVilgelm: Thank for suggestion

yts/test_suite_test.go
SVilgelm: actually I removed this logic, `embed` excluded all hidden files and directories

yts/test_suite_test.go
SVilgelm: good catch, looks like I removed it by mistake.
reverted back and enabled `thelper` linter

ccoVeille (approved):

yts/test_suite_test.go
```diff
@@ -11,70 +13,108 @@ import (
...
-var knownFailingTests = loadKnownFailingTests()
+const (
+	testDir        = "testdata/data-2022-01-17"
+	knownTestsFile = "known-failing-tests"
+)
+
+// run `make test-data` to download the test data
+//
+//go:embed known-failing-tests
+//go:embed testdata/data-2022-01-17
```
ccoVeille: Does this work ?

ccoVeille (changes requested):

yts/test_suite_test.go
SVilgelm: yep, I tested it locally

SVilgelm:
Okay, here is the catch. The CI wants testdata to be present :)

SVilgelm:
I can add `make test-data` to `lint` step, but we need to merge this PR https://github.com/yaml/go-yaml/pull/158 first, to remove `go vet`

yts/test_suite_test.go
SVilgelm: `embed` can embed multiple items
```go
//go:embed known-failing-tests
//go:embed testdata/data-2022-01-17
```
is the same as
```go
//go:embed known-failing-tests testdata/data-2022-01-17
```

ccoVeille:
I would remove the go:embed then. It doesn't bring much, but causes issues

SVilgelm:
let's put on hold this until other PRs are merged, then we can come back to this discussion

carloslima:
> @carloslima would like to see your review here.

Will look at it today. Sorry been a bit busy and didn't get a chance to follow recent changes

⏳

SVilgelm:
> I have changes pending merge on #64 that I'd prefer to get merged first and then apply any relevant changes here on top of that.
> 
> Can this be rebased on top of #64?

sure

SVilgelm:
@ccoVeille 
> I would remove the go:embed then. It doesn't bring much, but causes issues

Actually it brings enough to keep it:
1. It filters out unneeded folders automatically, which lets to have a better structured tests, nested tests that repeats the folders structure:
Screenshot 2025-11-02 at 9 48 40 AM
2. It fails earlier, on compilation stage, if the tests are missing.

I tried different approaches to make the nested structure and only this just works, without any magic

ccoVeille:
the `go test` are failing because the go vet are failing because of the use of go:embed when yts data was not made available in the GHA.

Which means the go vet fails for anything than when testing yts data.

I feel like there is no need to fight and fix them as we are about to remove the go vet, but I feel a bit strange for anyone trying to launch our tests with go test ./... will get an error because they cannot know they have to use go test -vet=off

The contribution guide could mention it, but I feel like we might get false positive issue being opened about that

SVilgelm:
In this case `go vet` fails because `embed` checks for the data to be present on compilation stage. It is unrelated to `vet` at all, the project is not compilable without testdata.
The `lint` job fails as well for same reason.
I want to explorer another approach to avoid this situation

SVilgelm:
@ccoVeille I dropped `embed` logic, so now it just walks through the testdata and creates nested test suits.

@carloslima I reviewed your PR and I would prefer to merge this one first and then rebase yours, it has too many issues.

yts/test_suite_test.go
```diff
@@ -1,232 +1,279 @@
...
 	case !runFailing && isKnownFailing:
-		t.Skipf("Skipping known failing test: %s", name)
+		tb.Skip("Skipping known failing test")
 	}
 }
 
 func TestYAMLSuite(t *testing.T) {
-	testDir := "./testdata/data-2022-01-17"
-	if _, err := os.Stat(testDir + "/229Q"); os.IsNotExist(err) {
+	if !fileExists(testDir + "/229Q") {
```
ccoVeille: Nitpicking

229Q is a magical number here. I would move it to a constant with a comment about we check the presence of one test folder

yts/test_suite_test.go
```diff
@@ -1,232 +1,279 @@
...
+				}
+			})
 		}
 	}
 }
 
+func fileExists(elem ...string) bool {
+	_, err := os.Stat(path.Join(elem...))
+	return err == nil
+}
```
ccoVeille, line 159: Nitpicking

It's a test, so it could fail here

```suggestion
func fileExists(t *testing.T, elem ...string) {
	fPath = path.Join(elem...)
	_, err := os.Stat(fpath)
	if err != nil {
		t.Fatalf("file %q is missing", fPath)
  }
}
```

yts/test_suite_test.go
```diff
@@ -1,232 +1,279 @@
...
+
+func readFile(tb testing.TB, must bool, elem ...string) []byte {
+	tb.Helper()
+	filename := path.Join(elem...)
+	data, err := os.ReadFile(filename)
+	if err != nil && must {
+		tb.Fatalf("Failed to read file %q: %v", filename, err)
+		tb.FailNow()
```
ccoVeille: Fatalf calls FailNow, no?

Or you wanted an Errorf + FailNow?

ccoVeille (approved):

yts/test_suite_test.go
SVilgelm: good catch, forgot to remove it

yts/test_suite_test.go
SVilgelm: in most use cases it should not fail, it ii should be as it is now

yts/test_suite_test.go
SVilgelm: changed the logic to fail if no tests were found

SVilgelm: no more magic numbers

ccoVeille (approved):

ccoVeille:
@SVilgelm please take a look at the following PRs:

- #171
- #172 
- #173 

You may find things you want to reuse.

ingydotnet:
@SVilgelm We miss you! But this is too old now. Closing.

PR #160, opened 2025-10-30
Convert limits tests to subtests and refactor benchmarks

SVilgelm:
Refactor limit_test.go to use t.Run/b.Run subtests for each case, ensuring proper test isolation by capturing tc in the loop. Replace the previous ad-hoc benchmark helpers and multiple benchmark functions with a single BenchmarkLimits that iterates limitTests and runs each case under b.Run. Update TestLimits to run in parallel, which improves test suite performance and eliminates the need for skipping the tests in short mode.

These changes make tests and benchmarks clearer, avoid shared-loop closure bugs, and unify benchmarking into one runner for easier maintenance.

### DIFF
--- a/limit_test.go
+++ b/limit_test.go
@@ -47,12 +47,11 @@ var limitTests = []struct {
 }
 
 func TestLimits(t *testing.T) {
-	if testing.Short() {
-		t.Skip("skipping test in short mode.")
-	}
 	for _, tc := range limitTests {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
 			var v any
 			err := yaml.Unmarshal(tc.data, &v)
 			if tc.error != "" {


limit_test.go
```diff
@@ -47,88 +47,35 @@ var limitTests = []struct {
...
+		tc := tc
+		b.Run(tc.name, func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				var v any
+				err := yaml.Unmarshal(tc.data, &v)
+				if tc.error != "" {
+					assert.ErrorMatches(b, tc.error, err)
+				} else {
+					assert.NoError(b, err)
+				}
```
ccoVeille, line 76: Nitpicking here

I would have written


```suggestion
				if tc.error != "" {
					assert.NoError(b, err)
					continue
				}
				assert.ErrorMatches(b, tc.error, err)
```

SVilgelm: sure, less complexity is better

ccoVeille (approved):

PR #161, opened 2025-10-30
Convert node tests to table-driven subtests

SVilgelm:
Refactor node-related tests to use t.Run subtests and t.Setenv for isolation. Replace index-based loops with range loops that capture the loop variable (item := item) to avoid shared variable issues. Wrap decode/encode phases into named subtests ("decode", "encode") where applicable. Also add checks for comment logging inside the subtests and ensure node mutation checks run inside the encode subtest.

Remove unused os import and tidy test logging messages for clarity. These changes improve test isolation, parallel friendliness, and make failures clearer by scoping each case to its own subtest.

### DIFF
--- a/node_test.go
+++ b/node_test.go
@@ -19,7 +19,6 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"os"
 	"reflect"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


node_test.go
```diff
@@ -2753,55 +2752,61 @@ var nodeTests = []struct {
...
 
 func TestNodeRoundtrip(t *testing.T) {
-	defer os.Setenv("TZ", os.Getenv("TZ"))
-	os.Setenv("TZ", "UTC")
-	for i, item := range nodeTests {
-		t.Logf("test %d: %q", i, item.yaml)
+	t.Setenv("TZ", "UTC")
+	for _, item := range nodeTests {
+		item := item
+		t.Run("", func(t *testing.T) {
```
ccoVeille, line 2803: I'm worried about the empty name here

SVilgelm: It is converted to a counter:
Screenshot 2025-10-30 at 10 30 58 AM

SVilgelm:
```
=== RUN   TestNodeRoundtrip
=== RUN   TestNodeRoundtrip/#00
    node_test.go:2759: yaml: "null\n"
=== RUN   TestNodeRoundtrip/#00/decode
=== RUN   TestNodeRoundtrip/#00/encode
=== RUN   TestNodeRoundtrip/#01
    node_test.go:2759: yaml: "[encode]null\n"
=== RUN   TestNodeRoundtrip/#01/encode
=== RUN   TestNodeRoundtrip/#02
    node_test.go:2759: yaml: "foo\n"
```

SVilgelm: It would be better to give the meaningful names, but I'm not sure I can do it :)

SVilgelm: I can do an explicit way: `t.Run(fmt.Sprintf("%d", i), ...`

ccoVeille: No it's fine the way you did

ccoVeille (approved):

SVilgelm:
@ingydotnet this PR is ready for review as well

PR #162, opened 2025-10-30
Provide an Option pattern for the YAML package

### LINKED ISSUE #155: Add options to Marshal and Unmarshal functions
The current way to change the default behavior is a creation of an Encoder or a Decoder, then change them by calling the functions like `KnownFields()` or `SetIndent`:
```go
d := NewDecoder(stream)
d.KnownFields(true)
d.Decode(&out)
```

The Marshal and Unmarshal functions use the default behavior only.

It would be great if we can pass the options to Marshal and Unmarshal functions using three-dots notation:

```go
func Marshal(in any, ...Option) (out []byte, err error) {
...
}
```

Using Option pattern is pretty safe and backward compatible.

ccoVeille:
This allows for more flexible configuration of encoding and decoding
operations by passing functional options to the relevant functions.

We will add more options in the future to customize behavior as needed.

Replaces #110
Closes #155

### DIFF
--- a/options.go
+++ b/options.go
@@ -2,47 +2,23 @@ package yaml
 
 import (
 	"errors"
-	"io"
 )
 
-// config holds configuration options for YAML processing.
-//
-// It allows customization of various aspects of YAML parsing and serialization.
-type config struct {
-	// indent controls the number of spaces to use for indentation
-	indent *int
-	// knownFields enables strict field checking during unmarshalling
-	knownFields *bool
-
-	writer *io.Writer
-
-	// compactSequenceIndent determines if sequence indicators are part of indentation
-	compactSequenceIndent *bool
-}
-
-func newConfig(opts ...Option) (*config, error) {
-	c := &config{}
-	for _, opt := range opts {
-		if err := opt(c); err != nil {
-			return nil, err
-		}
-	}
-	return c, nil
-}
-
-type Option func(*config) error
+// EncoderOption allows configuring the YAML encoder and marshaler.
+type EncoderOption func(*encoder) error
 
 // WithIndent sets the number of spaces to use for indentation when
 // marshaling YAML content.
 //
-// If not set, the default indentation level will be used.
-func WithIndent(indent int) Option {
-	return func(c *config) error {
+// A negative indent value will result in an error.
+// 0 can be used to reset the default indentation level.
+func WithIndent(indent int) EncoderOption {
+	return func(e *encoder) error {
 		if indent < 0 {
 			return errors.New("yaml: cannot indent to a negative number of spaces")
 		}
 
-		c.indent = &indent
+		e.indent = indent
 		return nil
 	}
 }
@@ -52,28 +28,23 @@ func WithIndent(indent int) Option {
 //
 // If compact is true, '- ' is treated as part of the indentation.
 // If compact is false, '- ' is not treated as part of the indentation.
-func WithCompactSequenceIndent(compact bool) Option {
-	return func(c *config) error {
-		c.compactSequenceIndent = &compact
+func WithCompactSequenceIndent(compact bool) EncoderOption {
+	return func(e *encoder) error {
+		e.emitter.CompactSequenceIndent = compact
 		return nil
 	}
 }
+
+// DecoderOption allows configuring the YAML decoder and unmarshaler.
+type DecoderOption func(*decoder) error
 
 // WithKnownFields enables or disables strict field checking during YAML unmarshalling.
 //
 // When enabled, unmarshalling will return an error if the YAML input contains fields
 // that do not correspond to any fields in the target struct.
-func WithKnownFields(knownFields bool) Option {
-	return func(c *config) error {
-		c.knownFields = &knownFields
-		return nil
-	}
-}
-
-// withWriter sets the output writer for YAML serialization.
-func withWriter(w io.Writer) Option {
-	return func(c *config) error {
-		c.writer = &w
+func WithKnownFields(knownFields bool) DecoderOption {
+	return func(d *decoder) error {
+		d.knownFields = knownFields
 		return nil
 	}
 }
--- a/yaml.go
+++ b/yaml.go
@@ -98,7 +98,7 @@ func Unmarshal(in []byte, out any) (err error) {
 //
 // See the documentation of [Unmarshal] for details about the
 // conversion of YAML into a Go value.
-func UnmarshalWithOptions(in []byte, out any, opts ...Option) error {
+func UnmarshalWithOptions(in []byte, out any, opts ...DecoderOption) error {
 	return unmarshal(in, out, opts...)
 }
 
@@ -127,8 +127,8 @@ func NewDecoder(r io.Reader) *Decoder {
 // The decoder introduces its own buffering and may read
 // data from r beyond the YAML values requested.
 //
-// The [Option]s can be used to customize the behavior of the decoder.
-func NewDecoderWithOptions(r io.Reader, opts ...Option) (*Decoder, error) {
+// The [DecoderOption]s can be used to customize the behavior of the decoder.
+func NewDecoderWithOptions(r io.Reader, opts ...DecoderOption) (*Decoder, error) {
 	decoder, err := newDecoder(opts...)
 	if err != nil {
 		return nil, err
@@ -184,7 +184,7 @@ func (n *Node) Decode(v any) (err error) {
 //
 // See the documentation for [UnmarshalWithOptions] for details about the
 // conversion of YAML into a Go value.
-func (n *Node) DecodeWithOptions(v any, opts ...Option) (err error) {
+func (n *Node) DecodeWithOptions(v any, opts ...DecoderOption) (err error) {
 	defer handleErr(&err)
 	var d *decoder
 	d, err = newDecoder(opts...)
@@ -202,7 +202,7 @@ func (n *Node) DecodeWithOptions(v any, opts ...Option) (err error) {
 	return nil
 }
 
-func unmarshal(in []byte, out any, opts ...Option) (err error) {
+func unmarshal(in []byte, out any, opts ...DecoderOption) (err error) {
 	defer handleErr(&err)
 	var d *decoder
 	d, err = newDecoder(opts...)
@@ -276,13 +276,14 @@ func Marshal(in any) (out []byte, err error) {
 // MarshalWithOptions serializes the value provided into a YAML document using the provided [Option]s.
 //
 // See the documentation of [Marshal] for details about the conversion of Go values into YAML.
-func MarshalWithOptions(in any, opts ...Option) (out []byte, err error) {
+func MarshalWithOptions(in any, opts ...EncoderOption) (out []byte, err error) {
 	defer handleErr(&err)
-	e, err := newEncoder(opts...)
+	e, err := newEncoder(nil, opts...)
 	if err != nil {
 		return nil, err
 	}
 	defer e.destroy()
+
 	e.marshalDoc("", reflect.ValueOf(in))
 	e.finish()
 	out = e.out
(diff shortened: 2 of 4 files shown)


ccoVeille:
@inteon @HadrienPatte @SVilgelm @ingydotnet please find here the implementation I made for the Option pattern.

yaml.go
```diff
@@ -268,10 +367,12 @@ func (n *Node) Encode(v any) (err error) {
 	defer p.destroy()
 	doc := p.parse()
 	*n = *doc.Content[0]
-	return nil
+	return err
```
SVilgelm: Why did you change it to return `err`? This is success branch and it is better to explicitly return `nil`, to avoid any confusions. And you didn't change it in other similar functions.

```suggestion
	return nil
```

SVilgelm (changes requested):
Thank you, looks really good, just one thing that needs to be changed

yaml.go
SVilgelm: Could you please revert?

ccoVeille: > Why did you change it to return `err`?

I have no idea 😅🤪

I will revert, then I will add a linter about this.

ccoVeille: I found no linter about it

I thought [nilnesserr](https://github.com/alingse/nilnesserr) or [nilerr](https://github.com/gostaticanalysis/nilerr) would report it, but it doesn't

what do you think @alingse ?

yaml.go
```diff
@@ -87,63 +87,110 @@ type Marshaler interface {
...
+	}, nil
 }
 
 // KnownFields ensures that the keys in decoded mappings to
 // exist as fields in the struct being decoded into.
-func (dec *Decoder) KnownFields(enable bool) {
-	dec.knownFields = enable
+//
+// You can also use the [WithKnownFields] option when using [NewDecoderWithOptions].
+func (d *Decoder) KnownFields(enable bool) {
```
HadrienPatte, line 147: Thoughts on labeling `KnownFields()` as [deprecated](https://go.dev/wiki/Deprecated)?
The only usecase I see where it could still be useful would be if people want to change the `knownFields` property of an existing `Decoder`, I don't know if that's a common usage pattern.

yaml.go
```diff
@@ -271,7 +370,9 @@ func (n *Node) Encode(v any) (err error) {
 	return nil
 }
 
-// SetIndent changes the used indentation used when encoding.
+// SetIndent changes the indentation used when encoding.
+//
+// You can also use the [WithIndent] option when using [NewEncoderWithOptions].
 func (e *Encoder) SetIndent(spaces int) {
```
HadrienPatte, line 375: Same question as https://github.com/yaml/go-yaml/pull/162/files#r2481344225

yaml.go
ccoVeille: I hesitated to suggest it in this PR.

I feel like it's open debate to deprecate it now or later.

It depends on what we want to do.

I didn't include it as I wanted to reduce the changes.

It can be merged like this, then another PR could mark is as deprecated

options.go
```diff
@@ -0,0 +1,79 @@
...
+	c := &config{}
+	for _, opt := range opts {
+		if err := opt(c); err != nil {
+			return nil, err
+		}
+	}
+	return c, nil
+}
+
+type Option func(*config) error
```
inteon: @ccoVeille could we separate Encoder and Decoder options?

SVilgelm: I don't think it is a good idea, because we can have an option that can apply for both, like FallbackToJSON.

inteon: Maybe a system like this: https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/options.go could be used (EncoderOption and DecoderOption interfaces implemented by each of the options, and some options implement both).

yaml.go
SVilgelm: I agree, with @HadrienPatte 
Let's mark those methods as deprecated. So, it will open an ability to remove it in the future v5 release.
And also will prevent from adding new methods. Now, if I like to add new option, I will have a dilemma - add only option or also add a method. But, if the methods are marked as deprecated, then no dilemma.

yaml.go
SVilgelm: because they both check the cases when an `err` must be returned instead of `nil`, not backward. Nothing wrong in returning `err` variable with `nil` value, it just may confuse a reader.

options.go
ccoVeille: I'm glad you raised this point.

I hesitated. I wanted to make the PR simple, so I went with only one option interface.

But I'm glad I'm not the one who thought about it. I can add it for sure.

The way kubernetes might be a bit too complex, but why not.

I'll wait for more feedbacks.

Note: we could also consider to merge like this, then to change before the v4 is published

SVilgelm: I just had a call with @ingydotnet and he has an idea about loading the config from a yaml or json file.
Could you make the fields Public, but keep the config private :) This will allow to marshal and unmarshal.

Also another idea is to have nested structures for Encoding and Decoding options.

ccoVeille: Someone want to add things to go-yaml CLI, haha !

I'm fine with the idea.

Having sub structs is indeed an option.

So we can pass only the minimal struct to encode, decode, node.decode, Marshall, Unmarshal.

ccoVeille: I will have to think about it. There are a few things that I need to figure out for options that apply to more than one, but not all of them considering:
- how it could imply to the YAML configuration idea that popped out, that is nice

- the multiple sub structs

- the multiple EncoderOption, DecoderOption, ... interfaces.

- the fact someone might one to set OptionA to foo to encoder but to bar for unmarshaller.

Maybe it can be bitmask setting to pass to the Option

SetIndent(4, ENCODER | NODE_ENCODER) // so here setting the SetIdent to all except Marshal

While the default way to use it could be

SetIndent(4, EVERYWHERE)

I cannot explain easily yet, but I feel there is something to think about in details.

I repeat I feel like this could be done after this PR is merged.

yaml.go
ccoVeille: I feel like this checker in gocritic is a good candidate. (Not yet activated in go-yaml)

https://go-critic.com/overview.html#nilvalreturn

If it doesn't support the code I had to update (I cannot check now) I'll open an issue as the linter seems the right one.

I unresolved the thread discussion to do not forget this.

options.go
SVilgelm: Let's make it simple for now. If an option can be applied to both, then we can figure out the better interface, maybe that option will use bit mask

ccoVeille:
I moved it as draft as I have a few changes to apply from the code review

options.go
```diff
@@ -0,0 +1,79 @@
...
+		return nil
+	}
+}
+
+// WithCompactSequenceIndent configures whether the sequence indicator '- ' is
+// considered part of the indentation when marshaling YAML content.
+//
+// If compact is true, '- ' is treated as part of the indentation.
+// If compact is false, '- ' is not treated as part of the indentation.
+func WithCompactSequenceIndent(compact bool) Option {
```
RaduBerinde: Does this and `WithKnownFields` need to take a bool? The default is false, so an option that just turns it on would be sufficient.

options.go
```diff
@@ -0,0 +1,79 @@
...
+
+type Option func(*config) error
+
+// WithIndent sets the number of spaces to use for indentation when
+// marshaling YAML content.
+//
+// If not set, the default indentation level will be used.
+func WithIndent(indent int) Option {
+	return func(c *config) error {
+		if indent < 0 {
```
RaduBerinde, line 17: What about 0? I think *some* indentation is required for correct marshaling?

options.go
RaduBerinde: If we keep a single Option, could we add some validation to error out when an option that doesn't make sense is used? (e.g. passing KnownFields to marshaling)

options.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+// config holds configuration options for YAML processing.
+//
+// It allows customization of various aspects of YAML parsing and serialization.
+type config struct {
+	// indent controls the number of spaces to use for indentation
+	indent *int
+	// knownFields enables strict field checking during unmarshalling
+	knownFields *bool
```
RaduBerinde: Can this and `compactSequenceIndent` just be bools? (fewer allocations is always better)
indent could also be an `int` with 0 meaning "default"

options.go
ccoVeille: You are right. I'll address this

options.go
ccoVeille: It's a good question. I'll think about it

options.go
ccoVeille: It's true these booleans are assigned to false by default. So yes, we could. I'll wait for other maintainers feedbacks, but why not.

About the 0 indent, we could but it's a bit strange, but why not with the correct documentation/comment.

options.go
RaduBerinde: IMO the slight complexity added by interfaces is well worth it to disallow using options incorrectly at compile time. On the flip side, the methods can operate directly on the Encoder/Decoder so we don't need a separate config. For example: https://gist.github.com/RaduBerinde/5af78b7632997d251edcc50996070e24

ccoVeille: You are right.

It's close to what @inteon suggested here by taking a look at what was done in kubernetes options

- https://github.com/yaml/go-yaml/pull/162#discussion_r2481719207

Maybe not, at least the gist you worked is clear and simple.

I will try to work on this.

I feel like this should be addressed before merging this PR

options.go
ccoVeille: Based on the fact Encoder had the following methods, I think so
- DefaultSeqIndent
- CompactSeqIndent

options.go
ccoVeille: I have removed the config struct, so I feel this one can be closed.

options.go
ccoVeille: I have simply updated the go doc of the method, if 0 is passed, the code will use 4 spaces as indentation.

options.go
RaduBerinde: We can have two options (DecaultSeqIndent and CompactSeqIndent) to match the existing methods.

options.go
ccoVeille: I have worked on providing a simple implementation, something simpler than what you suggested @RaduBerinde 

I decided to provide something simpler.

options.go
ccoVeille: I prefer having less methods and having people passing `true` as a value, even if most people won't pass a `false`, but some may need it.

Here is like wording it's open for debate.

options.go
```diff
@@ -0,0 +1,50 @@
...
+// If compact is true, '- ' is treated as part of the indentation.
+// If compact is false, '- ' is not treated as part of the indentation.
+func WithCompactSequenceIndent(compact bool) EncoderOption {
+	return func(e *encoder) error {
+		e.emitter.CompactSequenceIndent = compact
+		return nil
+	}
+}
+
+// DecoderOption defines a function type for configuring the YAML decoder or unmarshaler.
```
dolmen: "function type" is an implementation detail that should not be mentioned.

options.go
```diff
@@ -0,0 +1,50 @@
+package yaml
+
+import (
+	"errors"
+)
+
+// EncoderOption defines a function type for configuring the YAML encoder or marshaler.
```
dolmen: "function type" is an implementation detail that should not be mentioned.

dolmen (changes requested):

RaduBerinde:
Would love if this got merged - I am trying to update CockroachDB from v2 to v4 but we are blocked on not being able to pass the "known fields" option to `Node.Decode()`. CC @ingydotnet

ingydotnet:
> Would love if this got merged - I am trying to update CockroachDB from v2 to v4 but we are blocked on not being able to pass the "known fields" option to `Node.Decode()`. CC @ingydotnet

I agree this is the number one most important thing to get merged, but:
* I don't fully understand it yet
* We don't have any team reviews on this
* I feel we need to have a team meeting to go over this
* I don't have a time slot for that until next week

ingydotnet:
See #212 for an evolution of this PR.

ingydotnet:
Closing in favor of #212

PR #116, opened 2025-09-12
Fixes #70 with go-version-action

esacteksab:
I saw #70 and thought I'd take a swing at it. Admittedly, this is a very opinionated solution, that's simply more than "adopting the `go-version-action`, in addition to, the PR does:

1. Updates existing actions across all the workflows and pins the actions to SHAs for greater security
2. Formats workflow files so `yamllint` passes (spaces before/after comments)
  2a. It does so at the expense of line length (92 vs 80)
3. Sets `permissions` to `contents: read` on the `go.yaml` workflow file restricting access to the `GITHUB_TOKEN`.

I'll understand if this PR tries to do too much and won't be upset or offended if you choose to close it.

Thanks for your time.

PR #163, opened 2025-10-31
Update CI: pin actions and set permissions

SVilgelm:
Pin specific action revisions and replace floating tags across multiple
GitHub Actions workflows (CodeQL, Go, commit-message checks, and file
linters) to use exact commit SHAs. This improves reproducibility and
prevents unexpected behavior from upstream tag changes.

Set permissions: change per-workflow minimal permissions to
permissions: read-all where appropriate and add top-level permissions
read-all for several jobs to ensure required access for actions.

Simplify runner selection in CodeQL: remove conditional macOS vs
Ubuntu selection and use ubuntu-latest to unify execution environment.

Other small fixes: update action versions for setup-go, golangci-lint
integration, and CodeQL init/analyze steps, and ensure persist-credentials
is set consistently for checkouts. These changes increase CI stability
and make workflow behavior more deterministic.

Related:

- #118
- #116

Closes #124

- #124


.github/workflows/go.yaml
```diff
@@ -6,18 +6,20 @@ on:
...
         persist-credentials: false
     - name: Set up Go
-      uses: actions/setup-go@v6
+      uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00  # v6.0.0
       with:
         go-version: stable
-    - uses: golangci/golangci-lint-action@v8
+    - uses: golangci/golangci-lint-action@4afd733a84b1f43292c63897423277bb7f4313a9  # v8.0.0
       with:
         version: v2.4
```
ccoVeille: You could bump to 2.6.0 here maybe, then

.github/workflows/codeql.yaml
```diff
@@ -27,18 +28,18 @@ jobs:
...
     - name: Checkout repository
-      uses: actions/checkout@v5
+      uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8  # v5.0.0
       with:
         persist-credentials: false
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v4
+      uses: github/codeql-action/init@0499de31b99561a6d14a36a5f662c2a54f91beee  # v4.31.2
```
ccoVeille, line 38: About [yamllint issue](https://github.com/yaml/go-yaml/actions/runs/18986464360/job/54231143737#step:3:1) reported by the CI 

You can read this

> Using a yaml directive to disable next line should as be considered as it's an easy move, that imply nothing else on the yamllint configuration.
>
> `# yamllint disable-line rule:line-length` looks like it does, so. Then no changes need to be made to `.yamllint`.
_Originally posted by @esacteksab in https://github.com/yaml/go-yaml/issues/118#issuecomment-3288653091_

SVilgelm: yeah, I read that

SVilgelm: fixed by adding ignore directive for line-length

.github/workflows/go.yaml
SVilgelm: updated

.yamllint.yaml
```diff
@@ -17,3 +17,5 @@ rules:
     check-keys: true  # by default, only values are checked
     quote-type: single
     required: only-when-needed
+  line-length:
+    ignore: .github/workflows/*.yaml
```
ccoVeille: Please add a comment above this line.

Something like: "because of version pinning by hash for GHA actions, we remove the limit here.

ccoVeille: This is a follow up of https://github.com/yaml/go-yaml/pull/163#discussion_r2482809948

While I understand it's easy to proceed this way, I'm a bit annoyed it disable the line length on everything, and not only the one containing hashes.

I would prefer to use the per line yamllint directive

.github/workflows/codeql.yaml
ccoVeille: Follow up: https://github.com/yaml/go-yaml/pull/163#discussion_r2482828664

.github/workflows/check-commit-message.yaml
ccoVeille, line 1: > [!NOTE]
> I'm quoting a random file to be able to use a review comment instead of a comment. My comment is not related to this file

Shouldn't you mention the issue `#118` and PR `#163` in the PR description to benefits of automatic closes if yours get merged ?

.yamllint.yaml
SVilgelm: changed to per line

.github/workflows/check-commit-message.yaml
SVilgelm: done

.github/workflows/codeql.yaml
SVilgelm: changed to per-line comment

.github/workflows/go.yaml
```diff
@@ -6,42 +6,46 @@ on:
...
-      uses: actions/setup-go@v6
+      uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00  # v6.0.0
       with:
         go-version: stable
-    - uses: golangci/golangci-lint-action@v8
+    # yamllint disable-line rule:line-length
+    - uses: golangci/golangci-lint-action@4afd733a84b1f43292c63897423277bb7f4313a9  # v8.0.0
       with:
-        version: v2.4
+        version: v2.6.0  # must be in sync with the version in GNUmakefile
```
ccoVeille: About this, we should try to find a way to set the version in a file that GHA and Makefile could read.

What do you think about it?

It's just a suggestion to be moved to a separate issue and addressed later of course

ccoVeille (approved):

stefanprodan (approved):
LGTM

.github/workflows/go.yaml
SVilgelm: We can just use `make lint` instead of using this action, or we can do
```yaml
with:
  install-mode: none
```
in this case the action will not install the linter, but it will try to find `golangci-lint` in `PATH`: https://github.com/golangci/golangci-lint-action/blob/main/src/install.ts#L41-L47
so we can do something like:
```yaml
    - run: make lint-install
    - uses: golangci/golangci-lint-action@4afd733a84b1f43292c63897423277bb7f4313a9  # v8.0.0
      with:
        version: v2.6.0  # must be in sync with the version in GNUmakefile
        install-mode: none
      env:
        PATH: ${PATH}:.cache/local/bin
```

and
```make
lint-install: $(GOLANGCI-LINT)
  mv $(GOLANGCI-LINT) $(LOCAL-BIN)/golangci-lint
```

SVilgelm: done

ccoVeille: This change is great. But I would prefer to move this change out of the current PR.

Some reasons for that:
- have a clearer history
- allow a discussion about it
- consider other replacements in GHA workflows

I think about go test calling, make test

SVilgelm: I updated CI to install gloangci-lint via make and made the action to use pre-installed linter:
```
run golangci-lint
  Running [/home/runner/work/go-yaml/go-yaml/.cache/local/bin/golangci-lint config path] in [/home/runner/work/go-yaml/go-yaml] ...
  Running [/home/runner/work/go-yaml/go-yaml/.cache/local/bin/golangci-lint config verify] in [/home/runner/work/go-yaml/go-yaml] ...
  Running [/home/runner/work/go-yaml/go-yaml/.cache/local/bin/golangci-lint run] in [/home/runner/work/go-yaml/go-yaml] ...
  0 issues.
 ```

**Comment by SVilgelm** [comment]
@ingydotnet, Could you please review this PR when you have a time. After merging this one, we can enable setting to required actions to be pinned to sha:
Screenshot 2025-11-02 at 9 34 53 AM

**Review by ccoVeille — APPROVED** [review]

File: GNUmakefile [hunk 5]
```diff
@@ -124,7 +124,12 @@ define yts_pass_fail
 endef
 export yts_pass_fail
 
-$(GOLANGCI-LINT): $(GO-DEPS)
+# Downloads and installs golangci-lint to LOCAL-BIN (.cache/local/bin).
+golangci-lint-install: $(GO-DEPS) $(YTS-DIR)
```
ingydotnet: I would prefer it like:
```
$(GOLANGCI-LINT): $(GO-DEPS) $(YTS-DIR)
        curl -sSfL $(GOLANGCI-LINT-INSTALL) | \
          bash -s -- -b $(LOCAL-BIN) $(GOLANGCI-LINT-VERSION)
        mv $(LOCAL-BIN)/golangci-lint $@

golangci-lint-install: $(GOLANGCI-LINT)
```
if that works.

ingydotnet (changes requested):

GNUmakefile
SVilgelm: not really, because the Action is looking for a binary with name `golangci-lint`, when `$(GOLANGCI-LINT)` targets renames it to `golangci-lint-<version>`, that's why I split the target to two

ingydotnet: What do you mean by "the Action" in that sentence?

The `fmt` and `lint` Makefile rules depend on the value of `$(GOLANGCI-LINT)` (which contains a version, yes) and then they use that value directly via `$<`.

I don't get what needs `golangci-lint` (without the version in the name).

ingydotnet: I also wouldn't have the `$(GO-DEPS) $(YTS-DIR)` deps there as they appear to be unused in that rule.

SVilgelm: I'm talking about `golangci-lint-action`, that is used in CI to run `golangci-lint`.
The action expects `golangci-lint` binary in `PATH`. So, we need to be able to install it without renaming. Nothing should be changed for developers, all the targets must use `$(GOLANGCI-LINT)`, and only the CI will use `golangci-lint-install` target to install it as it is.

SVilgelm: The `$(GO-DEPS)` and `$(YTS-DIR)` are deleted

ingydotnet: > I'm talking about `golangci-lint-action`, that is used in CI to run `golangci-lint`. The action expects `golangci-lint` binary in `PATH`. So, we need to be able to install it without renaming. Nothing should be changed for developers, all the targets must use `$(GOLANGCI-LINT)`, and only the CI will use `golangci-lint-install` target to install it as it is.

I see. Thanks for explaining. I'd still do it a little different. I'll push a commit back to you.

PR #171, opened 2025-11-03
GNUmakefile fixes for testing

ingydotnet:
This PR is needed to fix a currently broken GNUmakefile

* Fixed 'make test' after 'make distclean'
* make test - now runs all tests
* make check - runs fmt tidy lint and test
* Test targets now clear the go testcache
* Fix broken 'make clean' to remove the CLI again

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -53,6 +53,9 @@ GO-DEPS += $(GO)
 else
 SHELL-DEPS := $(filter-out $(GO),$(SHELL-DEPS))
 endif
+GO-DEPS += clean-cache
+
+YTS-DEPS := $(GO-DEPS) $(YTS-DIR)
 
 SHELL-NAME := makes go-yaml
 include $(MAKES)/clean.mk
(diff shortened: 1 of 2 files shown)


GNUmakefile
```diff
@@ -58,19 +58,23 @@ SHELL-NAME := makes go-yaml
...
+	$(MAKE) tidy
+	$(MAKE) lint
+	$(MAKE) test
 
-test-data: $(YTS-DIR)
+test: test-unit test-yts-all
 
-test-all: test test-yts-all
+test-unit: $(GO-DEPS)
+	go test$(if $v, -v) -vet=off
```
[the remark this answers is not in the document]
ingydotnet: > The test-unit target is missing the package specifier `./...` that was present in the original test target. This will cause `go test` to fail with an error 'no Go files in [directory]' when run from the project root. Add `./...` to test all packages as intended.

Things actually fail with ./...
I want `make test-unit` to only run the go-yaml tests.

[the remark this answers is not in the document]
ingydotnet: Disagree. The purpose of `make test-unit` is to test the `yaml` package only.

ccoVeille: if you don't want to use ./...

Then we will have to consider a dedicated module for the CLI and yts, this way go test ./... will work as you expect, but also as any Go developper would expect to launch tests on the project

ccoVeille: I tried for the last 30 minutes, and it's not easy. I manage to do something for yts, but not for having the CLI in a separate go test

So here the best you could do is to add this


```suggestion
	go test$(if $v, -v) -vet=off . internal/...
```

ccoVeille: I opened 

- https://github.com/yaml/go-yaml/pull/172
- https://github.com/yaml/go-yaml/pull/173

What do you think to be the best @SVilgelm ?

I prefer #172 as it doesn't hide the test suite locally.

ingydotnet: @SVilgelm and I paired on this for a minute and modified this PR minimally. 

This PR is really just about getting `make test` to work from scratch again.

ingydotnet: > go test$(if $v, -v) -vet=off . internal/...

I'm also fine to do this.

ingydotnet: `go test internal/...` fails for me.
There's no testing of the internals in internals, just the testing framework tests.
We can address those later.
I want to get the makefile working again asap.

ccoVeille (approved):

ingydotnet:
This PR is a replaced by #198

PR #172, opened 2025-11-03
Fix test suite via env variable

ccoVeille:
This solution works should be fine for most cases


ccoVeille:
See https://github.com/yaml/go-yaml/pull/174#issuecomment-3488134933

Closing

EDIT: 👇 locked issue by error instead of closing 😅🤦‍♂️

PR #173, opened 2025-11-03
Fix Go test suite

ccoVeille:
testdata files are moved to a separate directory 'testdata/yts' to be excluded from go test ./... by default.


ccoVeille:
See https://github.com/yaml/go-yaml/pull/174#issuecomment-3488134933

Closing

PR #174, opened 2025-11-04
Download and run YAML Test Suite automatically in tests

SVilgelm:
Add TestMain to yts/test_suite_test.go to auto-download the
yaml-test-suite repository at a fixed tag into testdata/ when the
test directory is missing, and to checkout the required tag before
running the tests. Remove the previous runtime fatal check in
TestYAMLSuite that required manual data setup. Also add small
constants for the YTS URL, tag, and directory and import os/exec
and log to perform and report the clone/checkout actions.

Simplify GNUmakefile by removing explicit yts test-data rules and
YTS variables. Make test-yts, test-yts-all and test-yts-fail targets
no longer depend on a pre-fetched testdata directory so tests can
rely on the new automatic download behavior. These changes let CI and
local runs fetch the test suite on demand and reduce Makefile
maintenance for the test data.

### DIFF
--- a/yts/test_suite_test.go
+++ b/yts/test_suite_test.go
@@ -2,7 +2,9 @@ package yts
 
 import (
 	"encoding/json"
+	"log"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"reflect"
 	"strings"
@@ -10,6 +12,29 @@ import (
 
 	"go.yaml.in/yaml/v4"
 )
+
+const (
+	ytsURL = "https://github.com/yaml/yaml-test-suite"
+	ytsTAG = "data-2022-01-17"
+	ytsDir = "testdata/" + ytsTAG
+)
+
+func TestMain(m *testing.M) {
+	if _, err := os.Stat(ytsDir); os.IsNotExist(err) {
+		log.Printf("Downloading YTS test suite into %s...", ytsDir)
+		err = exec.Command("git", "clone", "-q", ytsURL, ytsDir).Run()
+		if err != nil {
+			log.Fatalf("git clone failed: %v", err)
+		}
+		log.Printf("Checking out YTS tag %s...", ytsTAG)
+		err = exec.Command("git", "-C", ytsDir, "checkout", "-q", ytsTAG).Run()
+		if err != nil {
+			log.Fatalf("git checkout failed: %v", err)
+		}
+	}
+
+	os.Exit(m.Run())
+}
 
 var knownFailingTests = loadKnownFailingTests()
 
@@ -48,13 +73,7 @@ func shouldSkipTest(t *testing.T) {
 }
 
 func TestYAMLSuite(t *testing.T) {
-	testDir := "./testdata/data-2022-01-17"
-	if _, err := os.Stat(testDir + "/229Q"); os.IsNotExist(err) {
-		t.Fatalf(`YTS tests require data files to be present at '%s'.
-Run 'make test-data' to download them first,
-or just run the tests with 'make test-all'.`, testDir)
-	}
-	runTestsInDir(t, testDir)
+	runTestsInDir(t, ytsDir)
 }
 
 func runTestsInDir(t *testing.T, dirPath string) {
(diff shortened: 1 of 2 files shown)


yts/test_suite_test.go
```diff
@@ -11,6 +13,29 @@ import (
...
+			log.Fatalf("git clone failed: %v", err)
+		}
+		log.Printf("Checking out YTS tag %s...", ytsTAG)
+		err = exec.Command("git", "-C", ytsDir, "checkout", "-q", ytsTAG).Run()
+		if err != nil {
+			log.Fatalf("git checkout failed: %v", err)
+		}
+	}
+
+	os.Exit(m.Run())
```
ccoVeille, line 36: os.Exit is no longer expected in TestMain


```suggestion
	m.Run()
```

yts/test_suite_test.go
```diff
@@ -11,6 +13,29 @@ import (
...
+		err = exec.Command("git", "clone", "-q", ytsURL, ytsDir).Run()
+		if err != nil {
+			log.Fatalf("git clone failed: %v", err)
+		}
+		log.Printf("Checking out YTS tag %s...", ytsTAG)
+		err = exec.Command("git", "-C", ytsDir, "checkout", "-q", ytsTAG).Run()
+		if err != nil {
+			log.Fatalf("git checkout failed: %v", err)
+		}
+	}
```
ccoVeille, line 34: Please consider moving this in a method and only call this method in TestMain

ccoVeille: But in fact, you don't need a test main

You only have one test. So you can use this function directly in TestYAMLSuite

This way you can use t.Logf without having a need to use log package

yts/test_suite_test.go
```diff
@@ -11,6 +13,29 @@ import (
 	"go.yaml.in/yaml/v4"
 )
 
+const (
+	ytsURL = "https://github.com/yaml/yaml-test-suite"
+	ytsTAG = "data-2022-01-17"
+	ytsDir = "testdata/" + ytsTAG
```
ccoVeille, line 19: Also [copilot is right](https://github.com/yaml/go-yaml/pull/174#discussion_r2488429281), depending on a tag has some security issues.

We could use the commit has here.

ccoVeille, line 19: filepath.Join should be used instead of strings concatenation

ingydotnet:
I would rather keep the test automations in the makefile.

yts/test_suite_test.go
SVilgelm: impossible for a constant, just need to make it as a variable

carloslima:
I second @ingydotnet. I'd prefer to keep the makefile.

ccoVeille:
I would have preferred to have what is implemented in this PR, but then it's a matter of language preference I feel.

As a Go Dev, I expect things to work with a `go test ./...`,  but the code would report and error informing users what to do, so I don't mind.

We will see in the future if we gets remarks or suggestions from contributors.

ccoVeille:
@carloslima any preference between 

- #172
- #173
- none of them ?

carloslima:
> @carloslima any preference between

Personally, neither seems ideal 😅

I feel like they're both trying to solve something that I personally don't see as a problem.

1. I don't like behavior that might be surprising. Automatically skipping a step but then failing in CI feels wrong to me.
2. Having tests inside `testdata` can be confusing or misleading.

In my view, both of these are worse than the original issue: you run `go test ./...` for the first time, see a clear error message telling you to run `make test-data`, and from then on `go test ./...` just works. Or you read the `CONTRIBUTING.md` file and run `make test-all`, which always takes care of everything.

ccoVeille:
Thanks for your feedback.

I get your points.

I closed #172 and #173

ccoVeille:
Based on the discussion, I feel like this PR could be closed.

@SVilgelm

ingydotnet:
> Based on the discussion, I feel like this PR could be closed.
> 
> @SVilgelm

I agree. Closing.

PR #175, opened 2025-11-04
Minimal fix for 2 Makefile bugs

ingydotnet:
'make test' was not working from a clean state.
'make clean' was not removing the go-yaml CLI.

I want the repo to not be broken asap. 
We have #171, #172, #173 and #174 but no agreement.
This fixes the current bugs. Then we can move on from there.


stefanprodan (approved):
LGTM

PR #177, opened 2025-11-06
Migrate to typed errors

inteon:
While working on https://github.com/yaml/go-yaml/pull/152, I got some feedback asking to improve the returned errors.

However, the errors can be improved separately from that PR. In this PR, I try to do that. This PR introduces typed errors and tries to make the error messages more useful by adding all the context that was previously missing.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -1144,13 +1144,13 @@ var unmarshalErrorTests = []struct {
 	{"v: [A,", "yaml: while parsing a flow node at line 1: did not find expected node content"},
 	{"v:\n- [A,", "yaml: while parsing a flow node at line 2: did not find expected node content"},
 	{"a:\n- b: *,", "yaml: while scanning an alias at line 2, column 5: line 2, column 6: did not find expected alphabetic or numeric character"},
-	{"a: *b\n", "yaml: unknown anchor 'b' referenced"},
+	{"a: *b\n", "yaml: line 1, column 4: unknown anchor 'b' referenced"},
 	{"a: &a\n  b: *a\n", "yaml: anchor 'a' value contains itself"},
 	{"value: -", "yaml: line 1, column 7: block sequence entries are not allowed in this context"},
 	{"a: !!binary ==", "yaml: !!binary value contains invalid base64 data"},
 	{"{[.]}", `yaml: cannot use '[]interface {}{"."}' as a map key; try decoding into yaml.Node`},
 	{"{{.}}", `yaml: cannot use 'map[string]interface {}{".":interface {}(nil)}' as a map key; try decoding into yaml.Node`},
-	{"b: *a\na: &a {c: 1}", `yaml: unknown anchor 'a' referenced`},
+	{"b: *a\na: &a {c: 1}", `yaml: line 1, column 4: unknown anchor 'a' referenced`},
 	{"%TAG !%79! tag:yaml.org,2002:\n---\nv: !%79!int '1'", "yaml: while scanning a %TAG directive at line 1: line 1, column 6: did not find expected whitespace"},
 	{"a:\n  1:\nb\n  2:", "yaml: while scanning a simple key at line 3: line 4, column 3: could not find expected ':'"},
 	{"a: 1\nb: 2\nc 2\nd: 3\n", "yaml: while scanning a simple key at line 3: line 4: could not find expected ':'"},
@@ -1205,9 +1205,9 @@ func TestParserErrorUnmarshal(t *testing.T) {
 	}
 	data := "a: 1\n=\nb: 2"
 	err := yaml.Unmarshal([]byte(data), &v)
-	asErr := new(yaml.ScannerError)
+	var asErr libyaml.ScannerError
 	assert.ErrorAs(t, err, &asErr)
-	expectedErr := &yaml.ScannerError{
+	expectedErr := libyaml.ScannerError{
 		ContextMark: libyaml.Mark{
 			Index:  5,
 			Line:   2,
@@ -1229,9 +1229,9 @@ func TestParserErrorDecoder(t *testing.T) {
 	var v any
 	data := "value: -"
 	err := yaml.NewDecoder(strings.NewReader(data)).Decode(&v)
-	asErr := new(yaml.ScannerError)
+	var asErr libyaml.ScannerError
 	assert.ErrorAs(t, err, &asErr)
-	expectedErr := &yaml.ScannerError{
+	expectedErr := libyaml.ScannerError{
 		Mark: libyaml.Mark{
 			Index:  7,
 			Line:   1,
@@ -2175,12 +2175,14 @@ func TestParserErrorUnknownAnchorPosition(t *testing.T) {
 	for _, test := range tests {
 		var n yaml.Node
 		err := yaml.Unmarshal([]byte(test.data), &n)
-		asErr := new(yaml.ParserError)
+		asErr := new(libyaml.ParserError)
 		assert.ErrorAs(t, err, &asErr)
-		expected := &yaml.ParserError{
+		expected := &libyaml.ParserError{
 			Message: "unknown anchor 'x' referenced",
-			Line:    test.line,
-			Column:  test.column,
+			Mark: libyaml.Mark{
+				Line:   test.line,
+				Column: test.column,
+			},
 		}
 		assert.DeepEqual(t, expected, asErr)
 	}
--- a/internal/libyaml/emitter.go
+++ b/internal/libyaml/emitter.go
@@ -141,13 +141,6 @@ func (emitter *Emitter) writeLineBreak(s []byte, i *int) error {
 	return nil
 }
 
-// Construct emitter error.
-func emitterError(problem string) error {
-	return EmitterError{
-		Message: problem,
-	}
-}
-
 // Emit an event.
 func (emitter *Emitter) Emit(event *Event) error {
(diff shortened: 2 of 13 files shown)


internal/libyaml/yaml.go
```diff
@@ -561,19 +576,7 @@ type AliasData struct {
...
-
-	// The byte about which the problem occurred.
-	ProblemOffset int
-	ProblemValue  int
-	ProblemMark   Mark
-
-	// The error Context.
-	Context     string
-	ContextMark Mark
+	Error error
```
ccoVeille: I would use Err here. The same way you use it for other errors struct


```suggestion
	Err error
```

internal/libyaml/emitter.go
```diff
@@ -142,6 +141,13 @@ func (emitter *Emitter) writeLineBreak(s []byte, i *int) error {
 	return nil
 }
 
+// Construct emitter error.
+func emitterError(problem string) error {
+	return EmitterError{
+		Message: problem,
+	}
+}
```
ccoVeille: For consistency with other part of the code where you are using another pattern. I would remove this 

```suggestion
```

See https://github.com/yaml/go-yaml/pull/177#discussion_r2500837639

internal/libyaml/reader.go
```diff
@@ -23,15 +23,25 @@
...
+}
+
+func (parser *Parser) directlySetReaderError(err error, offset int, value int) bool {
+	parser.Error = &ReaderError{
+		Offset: offset,
+		Value:  value,
+		Err:    err,
+	}
 	return false
 }
```
ccoVeille: I feel like you are using this only twice. And all if them in this file

I wouldn't add this method and directly return ReaderError as you do in this method.

Two reasons for this:
- we can live with the duplication cost
- the arguments of this method are unclear from the caller

See https://github.com/yaml/go-yaml/pull/177#discussion_r2500822650 and https://github.com/yaml/go-yaml/pull/177#discussion_r2500828841

internal/libyaml/reader.go
```diff
@@ -23,15 +23,25 @@
...
 // Set the reader error and return 0.
 func (parser *Parser) setReaderError(problem string, offset int, value int) bool {
-	parser.ErrorType = READER_ERROR
-	parser.Problem = problem
-	parser.ProblemOffset = offset
-	parser.ProblemValue = value
+	return parser.directlySetReaderError(
+		errors.New(problem),
+		offset, value,
+	)
```
ccoVeille:
```suggestion
	parser.Error = v&ReaderError{ 
                 Offset: offset, 
                 Value:  value, 
                 Err:    errors.New(problem),
         }
	return false
```

See https://github.com/yaml/go-yaml/pull/177#discussion_r2500815112

internal/libyaml/reader.go
```diff
@@ -101,7 +111,10 @@ func (parser *Parser) updateRawBuffer() bool {
 	if err == io.EOF {
 		parser.eof = true
 	} else if err != nil {
-		return parser.setReaderError("input error: "+err.Error(), parser.offset, -1)
+		return parser.directlySetReaderError(
+			fmt.Errorf("input error: %w", err),
+			parser.offset, -1,
+		)
```
ccoVeille:
```suggestion
		parser.Error = &ReaderError(
			Err: fmt.Errorf("input error: %w", err),
			Offset: parser.offset, Value: -1,
		)
		return false
```

see 
https://github.com/yaml/go-yaml/pull/177#discussion_r2500815112

internal/libyaml/emitter.go
```diff
@@ -205,7 +211,7 @@ func (emitter *Emitter) appendTagDirective(value *TagDirective, allow_duplicates
 			if allow_duplicates {
 				return nil
 			}
-			return errors.New("duplicate %TAG directive")
+			return emitterError("duplicate %TAG directive")
```
ccoVeille: See https://github.com/yaml/go-yaml/pull/177#discussion_r2500808204

```suggestion
			return EmitterError{ 
                 Message: "duplicate %TAG directive", 
         }
```

ccoVeille: But I feel the EmitterError should have an Err error, not a Message string

And so I would use


```suggestion
			return EmitterError{ 
                 Err: errors.New("duplicate %TAG directive"), 
         }
```

Also I would use a pointer as with other error strict for consistency

```suggestion
			return &EmitterError{ 
                 Err: errors.New("duplicate %TAG directive"), 
         }
```

yaml.go
```diff
@@ -319,24 +318,10 @@ func failf(format string, args ...any) {
...
-	if e.Line != 0 {
-		b.WriteString("line " + strconv.Itoa(e.Line) + ": ")
-	}
-	b.WriteString(e.Message)
-	return b.String()
-}
+type ParserError = libyaml.ParserError
+type ScannerError = libyaml.ScannerError
+type ReaderError = libyaml.ReaderError
+type WriterError = libyaml.WriterError
```
ccoVeille: I'm a bit annoyed because you are exporting all out internal type of error.

It could limit us in refactoring later.

I don't think user need to have access to every field of our error.

I would have defined sentinel errors, and return them by adding an ErrorIs on each type.

It's better for us and for users. They can simply catch 4 sentinel errors, and maybe we could wrap another common error witg them.

internal/libyaml/emitter.go
inteon: Currently only WriterError and ReaderError have an `Err` field (instead of `Message`) since they might wrap errors from the Writer/ Reader.

internal/libyaml/yaml.go
```diff
@@ -82,8 +83,22 @@ const (
...
+	var builder strings.Builder
+	if m.Line == 0 {
+		return "<unknown position>"
+	}
+
+	fmt.Fprintf(&builder, "line %d", m.Line)
+	if m.Column != 0 {
+		fmt.Fprintf(&builder, ", column %d", m.Column)
+	}
+
```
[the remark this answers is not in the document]
inteon: This logic/ issue was already in the source code: https://github.com/yaml/go-yaml/blob/e99d6a6e928a797feb752a3f839e0a7e45a46ddc/decode.go#L127-L131

I believe https://github.com/yaml/go-yaml/pull/135 might fix it.

yaml.go
inteon: This was the easiest way I found to keep the `TestParserErrorDecoder` test working. Feel free to make any changes to my PR (I don't think I fully understand what you are suggesting).

inteon: @ccoVeille I updated the source code, the internal errors are no longer being exported.

internal/libyaml/scanner_test.go
```diff
@@ -88,8 +88,8 @@ func runScanErrorTest(t *testing.T, tc TestCase) {
 		}
 	}
 	if wantError {
-		assert.Truef(t, parser.ErrorType != NO_ERROR, "Expected scanner error, but got none")
+		assert.Truef(t, parser.Err != nil, "Expected scanner error, but got none")
```
ccoVeille, line 91: You could have used assert.Errorf here

internal/libyaml/scanner_test.go
```diff
@@ -88,8 +88,8 @@ func runScanErrorTest(t *testing.T, tc TestCase) {
...
 	} else {
-		assert.Truef(t, parser.ErrorType == NO_ERROR, "Expected no scanner error, but got %v", parser.ErrorType)
+		assert.Truef(t, parser.Err == nil, "Expected no scanner error, but got %v", parser.Err)
```
ccoVeille, line 93: You could have used assert.NoErrorf here

decode_test.go
```diff
@@ -1177,7 +1178,10 @@ func TestUnmarshalErrors(t *testing.T) {
 		t.Run(fmt.Sprintf("test %d: %q", i, item.data), func(t *testing.T) {
 			var value any
 			err := yaml.Unmarshal([]byte(item.data), &value)
-			assert.ErrorMatchesf(t, item.error, err, "Partial unmarshal: %#v", value)
+			if err == nil {
+				t.Fatalf("got nil; want error %q - Partial unmarshal: %#v", item.error, value)
```
ccoVeille, line 1182: You could have used assert.Errorf here

decode_test.go
```diff
@@ -1187,7 +1191,10 @@ func TestDecoderErrors(t *testing.T) {
 		t.Run(fmt.Sprintf("test %d: %q", i, item.data), func(t *testing.T) {
 			var value any
 			err := yaml.NewDecoder(strings.NewReader(item.data)).Decode(&value)
-			assert.ErrorMatchesf(t, item.error, err, "Partial unmarshal: %#v", value)
```
ccoVeille, line 1190: Why removing errorMatches ?

you wanted exact equality ? Adding ^ and $ to the regexp would be enough no?

ccoVeille:
Shouldn't libyaml.Parse refactored to return an err instead of a bool and set an error into the struct ?

I remind a PR about this, no?. Maybe they could be addressed separately anyway. I'm fine with both approaches. But I'm raising the point

> [!NOTE]
> EDIT: In fact it's already in progress in #152.
> - https://github.com/yaml/go-yaml/pull/152

ingydotnet (approved):
LGTM

andig:
It seems this PR has moved the ParserError after rc3. Any idea where it's gone? This is a breaking change for us :/

PR #183, opened 2025-11-23
add Node.DecodeWithOptions

RaduBerinde:
This PR adds a `DecodeWithOptions()` method to `Node` which allows strict unmarshaling. Without this, it is impossible to require strict unmarshaling from inside a custom `UnmarshalYAML(node *yaml.Node)` handler. This is important for many applications where we don't want a typo in a configuration file to be silently ignored.

See https://github.com/go-yaml/yaml/issues/460 for more info. Note that this is functionally a regression in v3/v4; in v2 the decoder flags are automatically used by custom unmarshalers.

This is an adaptation of https://github.com/go-yaml/yaml/pull/691

### DIFF
--- a/node_test.go
+++ b/node_test.go
@@ -3174,6 +3174,46 @@ func TestNodeOmitEmpty(t *testing.T) {
 	assert.ErrorMatches(t, "yaml: cannot encode node with unknown kind 0", err)
 }
 
+func TestNodeDecodeWithOptions(t *testing.T) {
+	node := yaml.Node{
+		Kind: yaml.MappingNode,
+		Tag:  "!!map",
+		Content: []*yaml.Node{{
+			Kind:  yaml.ScalarNode,
+			Value: "a",
+			Tag:   "!!str",
+		}, {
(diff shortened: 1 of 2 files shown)


RaduBerinde:
To clarify on copilot's summary, this does not fully address the regression from v2 where decoder flags were automatically propagated to custom unmarshalers - we don't know if we should be strict or not inside `UnmarshalYAML`. But this at least allows us to always be strict. To address the regression fully, we would need to add a new version of the `Unmarshaler` interface.

ccoVeille:
Hi @RaduBerinde 

Please take a look at #162

I feel like it addresses the issue you reported.

Could you could review and confirm?

Thanks

RaduBerinde:
Oh! It does! That is great! Do we expect that to be merged before final v4?

ccoVeille:
> Oh! It does! That is great! Do we expect that to be merged before final v4?

I hope so

cc @ingydotnet

ingydotnet:
Ping me on Tuesday

PR #184, opened 2025-11-23
[internal] adds tests for internal files

### LINKED ISSUE #149: Do we need tests for the internal and cmd dirs?
```
$ go test ./...
ok  	go.yaml.in/yaml/v4	2.286s
?   	go.yaml.in/yaml/v4/cmd/go-yaml	[no test files]
?   	go.yaml.in/yaml/v4/internal/libyaml	[no test files]
ok  	go.yaml.in/yaml/v4/internal/testutil/assert	(cached)
ok  	go.yaml.in/yaml/v4/yts	0.099s
```

Also should make test trigger `go test ./...` instead of just `go test`?

colinjlacy:
This branch adds unit tests in the integration
style to the internal/libyaml directory. it
focuses primarily on the majority use cases and
does not get too creative with edge-case yaml
examples, which can likely be filled in as issues
are opened and bugs are squashed. The majority of
these tests were written by Claude in cursor.

To test, simply run:
```
make test-all
go test ./... --cover
```

fixes #149 in part

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -66,9 +66,6 @@ count ?= 1
 
 # Test rules:
 test: $(GO-DEPS)
-	go test$(if $v, -v) -vet=off .
-
-test-cover: $(GO-DEPS)
 	go test$(if $v, -v) -vet=off --cover . ./internal/...
 
 test-data: $(YTS-DIR)


ingydotnet:
> ```
> go test ./... --cover
> ```

Can you add `test-cover` rule to the Makefile for that?

internal/libyaml/reader_test.go
```diff
@@ -0,0 +1,358 @@
...
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
```
ccoVeille: I'm unsure we need to apply the legacy license to file we add.

They do use old code that were licensed under these Licenses

But here, these are files we add for the yaml/go-yaml project.

I wouldn't use a license header, or if I do, I would clearly state these new files are licensed under the license of the yaml/go-yaml project.

So I would simply use

```suggestion
// SPDX-License-Identifier: Apache-2.0
```


What do you think @ingydotnet ?

internal/libyaml/reader_test.go
```diff
@@ -0,0 +1,358 @@
...
+	}
+	if parser.Problem != "test problem" {
+		t.Errorf("setReaderError() Problem = %q, want \"test problem\"", parser.Problem)
+	}
+	if parser.ProblemOffset != 10 {
+		t.Errorf("setReaderError() ProblemOffset = %d, want 10", parser.ProblemOffset)
+	}
+	if parser.ProblemValue != 0x1234 {
+		t.Errorf("setReaderError() ProblemValue = %#x, want 0x1234", parser.ProblemValue)
+	}
```
ccoVeille: Here and everywhere, you could have used the assertion package we created

https://pkg.go.dev/go.yaml.in/yaml/v4@v4.0.0-rc.3/internal/testutil/assert

But why not

internal/libyaml/reader_test.go
```diff
@@ -0,0 +1,358 @@
...
+func TestYamlStringReadHandler(t *testing.T) {
+	parser := NewParser()
+	input := []byte("test data")
+	parser.input = input
+	parser.input_pos = 0
+
+	buffer := make([]byte, 10)
+	n, err := yamlStringReadHandler(&parser, buffer)
+
+	if err != nil && err != io.EOF {
```
ccoVeille: Here and everywhere `errors.Is` is the way to check error equality

GNUmakefile
```diff
@@ -68,6 +68,9 @@ count ?= 1
 test: $(GO-DEPS)
 	go test$(if $v, -v) -vet=off .
 
+test-cover: $(GO-DEPS)
+	go test$(if $v, -v) ./... --cover
+
```
ccoVeille, line 70: I know @ingydotnet asked you to add this, but I would have simply used this

```suggestion
test: $(GO-DEPS)
go test$(if $v, -v) -vet=off --cover . 

```

Because the cover is fine in the context of the test, here you are adding cover for everything, while we stated we cannot use ./...

If there is a need for having a test-cover rule, you should use this

```suggestion
test: $(GO-DEPS)
go test$(if $v, -v) -vet=off .

test-cover: $(GO-DEPS)
	go test$(if $v, -v) -vet=off --cover .

```

internal/libyaml/reader_test.go
ingydotnet: I agree, it doesn't belong.

GNUmakefile
colinjlacy: @ccoVeille sorry, I didn't know we aren't using `./...`. Do you mind explaining why that is?

ccoVeille: It's because the existence of yaml test suite in yts.

You can find some context here:

https://github.com/yaml/go-yaml/pull/171#discussion_r2487945015

I would like to suggest you to read all resolved comments, including thr one opened by Copilot.

We tried to fix things with #171 #172, #173 and #174

But there was no agreement, so #175 was opened.

I would say we have people who use `make test`command to launch tests, and Gophers with muscle memories who use `go test ./...`

It's uneasy to make everyone happy 😅

So for now, there is no real progress on the situation.

I would love getting feedbacks from you here.

ccoVeille: But for now, using this could be a solution for your PR

```suggestion
test: $(GO-DEPS)
go test$(if $v, -v) -vet=off --cover . internal/...

```

colinjlacy: Thanks for the context. I can see this has been a back-and-forth issue. I'm an old Go hand as well, so the muscle memory is real for me too. Still, I understand the other perspective entirely. I added the modified line as you suggested and we can adjust from there if/when more tests are added.

GNUmakefile
```diff
@@ -68,6 +68,9 @@ count ?= 1
 test: $(GO-DEPS)
 	go test$(if $v, -v) -vet=off .
 
+test-cover: $(GO-DEPS)
+	go test$(if $v, -v) -vet=off --cover . ./internal/...
```
ccoVeille, line 69: Let's continue there

If you take a look at my previous comment, you could see I was suggesting this


```suggestion
test: $(GO-DEPS)
	go test$(if $v, -v) -vet=off --cover . ./internal/...
```

But if there is real will to have a separate `make test-cover` this should be used.

```suggestion
test: $(GO-DEPS)
	go test$(if $v, -v) -vet=off . ./internal/...
 
test-cover: $(GO-DEPS)
	go test$(if $v, -v) -vet=off --cover . ./internal/...
```

ccoVeille (approved):
I didn't review all tests. But based in the fact you are not updating the code, but adding tests I feel like I can approve anyway.

colinjlacy:
Closing in favor of #194

PR #189, opened 2025-11-28
libyaml/reader: adds control char support

### LINKED ISSUE #179: YAML 1.2: Error parsing C1 control character inside double-quoted string
```go
package main

import (
	"fmt"

	"go.yaml.in/yaml/v4"
)

func main() {
	var str string
	err := yaml.Unmarshal([]byte("\"\u0080\""), &str)
	if err != nil {
		fmt.Println(err)
	}
}
```

This program should exit without printing anything, but it reports the error `yaml: control characters are not allowed`. To quote the YAML 1.2 specification, “YAML processors must allow all non-C0 characters inside quoted scalars.”

colinjlacy:
This change allows C1 control characters in YAML
values, as per the YAML 1.2 spec. This should
help with moving the need toward v1.2 compliance.

To test:
- run `make cli`
- run `./go-yaml -t <<<$'"\u0080"'`
- you should get the following output:
```
- {Token: STREAM-START}
- {Token: SCALAR, Value: "\x80", Style: Double}
- {Token: STREAM-END}
```

fixes #179

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -927,10 +927,14 @@ var unmarshalTests = []struct {
 		"true #" + strings.Repeat(" ", 512*3),
 		"true",
 	},
-	// Control characters
+	// C1 Control characters
 	{
 		"\u0080",
 		"\u0080",
+	},
+	{
+		"\u009F",
+		"\u009F",
 	},
 
 	// CRLF


internal/libyaml/reader.go
```diff
@@ -364,16 +364,15 @@ func (parser *Parser) updateBuffer(length int) bool {
...
 			switch {
 			case value == 0x09:
 			case value == 0x0A:
 			case value == 0x0D:
-			case value >= 0x20 && value <= 0x7E:
-			case value == 0x85:
-			case value >= 0xA0 && value <= 0xD7FF:
+			case value >= 0x20 && value <= 0xD7FF:
 			case value >= 0xE000 && value <= 0xFFFD:
 			case value >= 0x10000 && value <= 0x10FFFF:
```
ccoVeille, line 391: Could you add a comment on each case with the character it's about when there is only one, and a text for explaining the range.

The hyperlink is good, but when reading code or history it's good to have a clear idea of what it's about

ccoVeille, line 391: Also, I feel like some of the one you add are only supported in YAML 1.2 spec, and not by previous specs. So maybe you could add an information about it in comment for each line you add. It would be easier later to enable/disable them via an option such as allowing YAML 1.1 strict only in the parsing.

decode_test.go
```diff
@@ -927,6 +927,11 @@ var unmarshalTests = []struct {
 		"true #" + strings.Repeat(" ", 512*3),
 		"true",
 	},
+	// Control characters
```
ingydotnet: Change to `// C1 Control characters` and add a test for `\u009f`

ingydotnet (changes requested):

decode_test.go
ingydotnet: Awesome. Thx!

PR #191, opened 2025-11-30
Return `ParserError` on unknown anchor reference

### LINKED ISSUE #190: Unknown anchor causes an error without source location information
## Repro

```go
b := []byte("*foo\n")

var node yaml.Node
if err := yaml.Unmarshal(b, &node); err != nil {
	switch e := err.(type) {
	case *yaml.ParserError:
		fmt.Fprintf(os.Stderr, "ParserError:%d:%d: %s\n", e.Line, e.Column, e.Message)
	case *yaml.TypeError:
		for _, u := range e.Errors {
			fmt.Fprintf(os.Stderr, "TypeError:%d:%d: %s\n", u.Line, u.Column, u.Error())
		}
	default:
		fmt.Fprintf(os.Stderr, "UNEXPECTED ERROR: %s\n", e.Error())
	}
}
```

Run the above code and check the output.

## Expected behavior

`ParseError` or `TypeError` is reported for the unknown anchor reference `foo` so that we can know the line and column where the error occurred.

## Actual behavior

The output is:

```
UNEXPECTED ERROR: yaml: unknown anchor 'foo' referenced
```

This means that `yaml.Unmarshal` returned neither `yaml.ParseError` nor `yaml.TypeError`. I checked the `err` instance and it was actually a plain `fmt.Errorf` error object. It only keeps the error message.

```
&errors.errorString{s:"yaml: unknown anchor 'foo' referenced"}
```

rhysd:
Fixes #190

This PR makes the decoder return `ParserError` instead of standard `fmt.Errorf` when detecting a reference to an unknown anchor so that it can keep the source location.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -216,7 +216,8 @@ func (p *parser) alias() *Node {
 	n := p.node(AliasNode, "", "", string(p.event.Anchor))
 	n.Alias = p.anchors[n.Value]
 	if n.Alias == nil {
-		failf("unknown anchor '%s' referenced", n.Value)
+		msg := fmt.Sprintf("unknown anchor '%s' referenced", n.Value)
+		fail(&ParserError{msg, n.Line, n.Column})
 	}
 	p.expect(libyaml.ALIAS_EVENT)
 	return n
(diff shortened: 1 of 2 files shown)


ccoVeille (approved):

ingydotnet (approved):
lgtm

PR #193, opened 2025-12-03
Fixes flow sequence colon tokenizing

### LINKED ISSUE #188: Failure to parse `[ :foo ]`
Scalars starting with colon in a flow collection fail in the parser stage:

```
$ make go-yaml 
go build -o go-yaml ./cmd/go-yaml
$ ./go-yaml -e <<<'- :foo'
- {Event: DOCUMENT-START}
- {Event: SEQUENCE-START}
- {Event: SCALAR, Value: ':foo'}
- {Event: SEQUENCE-END}
- {Event: DOCUMENT-END}
$ ./go-yaml -e <<<'[ :foo ]'
2025/11/26 08:34:43 Failed to process events:failed to decode YAML: yaml: did not find expected node content
$ 
```

colinjlacy:
This fixes #188 , in which YAML values like `[ :foo ]` fail to be parsed based on inaccurate handling of the `':'`, according to expected behavior listed [in the spec](https://yaml.org/spec/1.2.2/#rule-ns-plain-char). 

To test, run `make test-all`. 

**Current state:** this will result in a known failing test (DBG4) to pass, which results in exit 1. Looking for guidance on how to address this.

### DIFF
--- a/internal/libyaml/scanner.go
+++ b/internal/libyaml/scanner.go
@@ -883,8 +883,8 @@ func (parser *Parser) isFlowSequence() bool {
 	if len(parser.tokens) == 0 {
 		return false
 	}
-	return parser.tokens[len(parser.tokens)-1].Type == FLOW_ENTRY_TOKEN ||
-		parser.tokens[len(parser.tokens)-1].Type == FLOW_SEQUENCE_START_TOKEN
+	previousToken := parser.tokens[len(parser.tokens)-1]
+	return previousToken.Type == FLOW_ENTRY_TOKEN || previousToken.Type == FLOW_SEQUENCE_START_TOKEN
 }
 
 func (parser *Parser) simpleKeyIsValid(simple_key *SimpleKey) (valid, ok bool) {
--- a/yts/known-failing-tests
+++ b/yts/known-failing-tests
@@ -80,6 +80,7 @@ TestYAMLSuite/CFD4/EventComparisonTest
 TestYAMLSuite/CN3R/EventComparisonTest
 TestYAMLSuite/CVW2/UnmarshalTest
 TestYAMLSuite/CVW2/EventComparisonTest
+TestYAMLSuite/DBG4/JSONComparisonTest
 TestYAMLSuite/DFF7/MarshalTest
 TestYAMLSuite/DK3J/EventComparisonTest
 TestYAMLSuite/DK3J/MarshalTest
@@ -106,7 +107,6 @@ TestYAMLSuite/FRK4/EventComparisonTest
 TestYAMLSuite/G4RS/EventComparisonTest
 TestYAMLSuite/G5U8/UnmarshalTest
 TestYAMLSuite/G5U8/EventComparisonTest
-TestYAMLSuite/00#07/EventComparisonTest
 TestYAMLSuite/HRE5/UnmarshalTest
 TestYAMLSuite/HRE5/EventComparisonTest
 TestYAMLSuite/HWV9/UnmarshalTest


ccoVeille:
> Current state: this will result in a known failing test (DBG4) to pass, which results in exit 1. Looking for guidance on how to address this.

Well this sounds like a good news.

I feel like the only thing missing is to remove the needed lines from

https://github.com/yaml/go-yaml/blob/main/yts%2Fknown-failing-tests

So line 83 and 84

colinjlacy:
This is interesting. I can reproduce this failure locally with `make test-yts`, which just runs `go test ./yts`. But if I run `make test-all` it passes. I'm digging into it, but if anyone has ideas on what may be happening behind the scenes they'd be welcome

internal/libyaml/scanner.go
```diff
@@ -879,6 +879,14 @@ func (parser *Parser) fetchNextToken() (ok bool) {
 		"found character that cannot start any token")
 }
 
+func (parser *Parser) isFlowSequence() bool {
+	if len(parser.tokens) == 0 {
+		return false
+	}
+	return parser.tokens[len(parser.tokens)-1].Type == FLOW_ENTRY_TOKEN ||
+		parser.tokens[len(parser.tokens)-1].Type == FLOW_SEQUENCE_START_TOKEN
+}
```
ccoVeille, line 888: This syntax is a bit too complicated. First you access to this twice

		parser.tokens[len(parser.tokens)-1]

It leads to readability issue, I feel.

I would like to suggest you two things, that could be done separately or together.

- you could use a variable with a meaningful name to store `parser.tokens[len(parser.tokens)-1]`
- you could use a switch on `parser.tokens[len(parser.tokens)-1]` or this variable to check the value of the type.

ingydotnet:
https://play.yaml.com/#IyBPdXRzaWRlIGZsb3cgY29sbGVjdGlvbjoKLSA6OnZlY3RvcgotICI6IC0gKCkiCi0gVXAsIHVwLCBhbmQgYXdheSEKLSAtMTIzCi0gaHR0cDovL2V4YW1wbGUuY29tL2ZvbyNiYXIKIyBJbnNpZGUgZmxvdyBjb2xsZWN0aW9uOgotIFsgOjp2ZWN0b3IsCiAgIjogLSAoKSIsCiAgIlVwLCB1cCBhbmQgYXdheSEiLAogIC0xMjMsCiAgaHR0cDovL2V4YW1wbGUuY29tL2ZvbyNiYXIgXQo=

This is DBG4. You can select any test suite test from there.

Also: https://play.yaml.com/#WyA6Zm9vIF0=

colinjlacy:
The good news is I've narrowed down what I think is the issue in `make test-yts`:
- the YAML Unmarshalled type of the input `-123` is an `Int`
- the JSON Unmarshalled type is a `float64`

This is failing `reflect.DeepEqual`, and would also explain why the printed values look identical. I'm going to look for similar tests to see if/how this is overcome in those instances

colinjlacy:
The problem I've been hitting is actually a new failing test, unfortunately: `TestYAMLSuite/DBG4/JSONComparisonTest`. The current interpretation of `-123` in `yaml.Unmarshall` is different (and superior, IMO) to the interpretation of that same set of characters in `json.Unmarshall`. `json.Unmarshall` automatically treats any number as a `float64` [according to the docs](https://pkg.go.dev/encoding/json#Unmarshal). The only way to get this test to pass is to change how `yaml.Unmarshall` interprets `-123` to make it conform to `json.Unmarshall`, which would be a breaking change. So let's not do that. 

The positive is that we're still seeing three less known failing tests after this fix. So we've got that going for us.

ingydotnet:
Is this ready for review, @colinjlacy ?
Request me when it is.

ingydotnet:
> The problem I've been hitting is actually a new failing test, unfortunately: `TestYAMLSuite/DBG4/JSONComparisonTest`. The current interpretation of `-123` in `yaml.Unmarshall` is different (and superior, IMO) to the interpretation of that same set of characters in `json.Unmarshall`. `json.Unmarshall` automatically treats any number as a `float64` [according to the docs](https://pkg.go.dev/encoding/json#Unmarshal). The only way to get this test to pass is to change how `yaml.Unmarshall` interprets `-123` to make it conform to `json.Unmarshall`, which would be a breaking change. So let's not do that.
> 
> The positive is that we're still seeing three less known failing tests after this fix. So we've got that going for us.

I'm glad you brought this up. Unlike JSON, YAML rules for what values get assigned which yaml tags (types) is intended to be configurable. In the spec this is called YAML Schema. I'm working on a simple configuration for specifying whatever rules you want. The gist of this is that people will soon be able to do either way above very easily. The default YAML Schema will of course not change here. But we can have a preset JSON one that people can use, no sweat, for full JSON compat.

PR #194, opened 2025-12-06
Data driven tests for internal/libyaml

ingydotnet:
Refactored from and replaces @colinjlacy's https://github.com/yaml/go-yaml/pull/184

Uses super clean, simple, consistent and documented foo_test.yaml files for test data, and minimal foo_test.go files for the code running the tests.

Added internal/libyaml/README.md that explains both the code and the test layout.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -66,7 +66,7 @@ count ?= 1
 
 # Test rules:
 test: $(GO-DEPS)
-	go test$(if $v, -v) -vet=off .
+	go test$(if $v, -v) -vet=off --cover . ./internal/...
 
 test-data: $(YTS-DIR)
 
--- a/internal/libyaml/README.md
+++ b/internal/libyaml/README.md
@@ -1,0 +1,533 @@
+# internal/libyaml
+
+This package provides low-level YAML processing functionality through a 3-stage
+pipeline: Scanner → Parser → Emitter.
+It implements the libyaml C library functionality in Go.
+
+## Directory Overview
+
+The `internal/libyaml` package implements the core YAML processing stages:
+
+1. **Scanner** - Tokenizes YAML text into tokens
+2. **Parser** - Converts tokens into events following YAML grammar rules
+3. **Emitter** - Serializes events back into YAML text
+
+## File Organization
+
+### Main Source Files
+
+- **scanner.go** - YAML scanner/tokenizer implementation
+- **parser.go** - YAML parser (tokens → events)
+- **emitter.go** - YAML emitter (events → YAML output)
+- **api.go** - Public API for Parser and Emitter types
+- **yaml.go** - Core types and constants (Event, Token, enums)
+- **reader.go** - Input handling and encoding detection
+- **writer.go** - Output handling
+- **yamlprivate.go** - Internal types and helper functions
+
+### Test Files
+
+- **scanner_test.go** - Scanner tests
+- **parser_test.go** - Parser tests
+- **emitter_test.go** - Emitter tests
+- **api_test.go** - API tests
+- **yaml_test.go** - Utility function tests
+- **reader_test.go** - Reader tests
+- **writer_test.go** - Writer tests
+- **yamlprivate_test.go** - Character classification tests
+- **loader_test.go** - Data loader scalar resolution tests
+- **yamldatatest_test.go** - YAML test data loading framework
+- **yamldatatest_loader.go** - YAML test data loader with scalar type resolution (exported for reuse)
+
+### Test Data Files (in `testdata/`)
+
+- **scanner.yaml** - Scanner test cases
+- **parser.yaml** - Parser test cases
+- **emitter.yaml** - Emitter test cases
+- **api.yaml** - API test cases
+- **yaml.yaml** - Utility function test cases
+- **reader.yaml** - Reader test cases
+- **writer.yaml** - Writer test cases
+- **yamlprivate.yaml** - Character classification test cases
+- **loader.yaml** - Data loader scalar resolution test cases
+
+## Processing Pipeline
+
+### 1. Scanner (scanner.go)
+
+The scanner converts YAML text into tokens.
+
+**Input**: Raw YAML text (string or []byte)
+**Output**: Stream of tokens
+
+**Token types include**:
+- `SCALAR_TOKEN` - Plain, quoted, or block scalar values
+- `KEY_TOKEN`, `VALUE_TOKEN` - Mapping key/value indicators
+- `BLOCK_MAPPING_START_TOKEN`, `FLOW_MAPPING_START_TOKEN` - Mapping delimiters
+- `BLOCK_SEQUENCE_START_TOKEN`, `FLOW_SEQUENCE_START_TOKEN` - Sequence delimiters
+- `ANCHOR_TOKEN`, `ALIAS_TOKEN` - Anchor definitions and references
+- `TAG_TOKEN` - Type tags
+- `DOCUMENT_START_TOKEN`, `DOCUMENT_END_TOKEN` - Document boundaries
+
+**Responsibilities**:
+- Character encoding detection (UTF-8, UTF-16LE, UTF-16BE)
+- Line break normalization
+- Indentation tracking
+- Quote and escape sequence handling
+
+### 2. Parser (parser.go)
+
+The parser converts tokens into events following YAML grammar rules.
+
+**Input**: Stream of tokens from Scanner
+**Output**: Stream of events
+
+**Event types include**:
+- `STREAM_START_EVENT`, `STREAM_END_EVENT` - Stream boundaries
+- `DOCUMENT_START_EVENT`, `DOCUMENT_END_EVENT` - Document boundaries
+- `SCALAR_EVENT` - Scalar values
+- `MAPPING_START_EVENT`, `MAPPING_END_EVENT` - Mapping boundaries
+- `SEQUENCE_START_EVENT`, `SEQUENCE_END_EVENT` - Sequence boundaries
+- `ALIAS_EVENT` - Anchor references
+
+**Responsibilities**:
+- Implementing YAML grammar and validation
+- Managing document directives (%YAML, %TAG)
+- Resolving anchors and aliases
+- Tracking implicit vs explicit markers
+- Style preservation (plain, single-quoted, double-quoted, literal, folded)
+
+### 3. Emitter (emitter.go)
+
+The emitter converts events back into YAML text.
+
+**Input**: Stream of events
+**Output**: YAML text
+
+**Responsibilities**:
+- Style selection (plain/quoted scalars, block/flow collections)
+- Formatting control (canonical mode, indentation, line width)
+- Character encoding
+- Anchor and tag serialization
+- Document marker generation (---, ...)
+
+**Configuration options**:
+- `Canonical` - Emit in canonical YAML form
+- `Indent` - Indentation width (2-9 spaces)
+- `Width` - Line width (-1 for unlimited)
+- `Unicode` - Enable Unicode character output
+- `LineBreak` - Line break style (LN, CR, CRLN)
+
+## Testing Framework
+
+### Test Architecture
+
+The testing framework uses a data-driven approach:
+
+1. **Test data** is stored in YAML files in the `testdata/` directory
+2. **Test logic** is implemented in Go files (`*_test.go`)
+3. **One-to-one pairing**: Each `testdata/foo.yaml` has a corresponding `foo_test.go`
+
+**Benefits**:
+- Easy to add new test cases without writing Go code
+- Test data is human-readable and self-documenting
+- Test logic is reusable across many test cases
+- Test data is separated from test code for clarity
+- Tests can become a common suite for multiple YAML frameworks
+
+### Test Data Files
+
+Each YAML file contains test cases for a specific component:
+
+- **scanner.yaml** - Scanner/tokenization tests
+  - Token sequence verification
+  - Token property validation (value, style)
+  - Error detection
+
+- **parser.yaml** - Parser/event generation tests
+  - Event sequence verification
+  - Event property validation (anchor, tag, value, directives)
+  - Error detection
+
+- **emitter.yaml** - Emitter/serialization tests
+  - Event-to-YAML conversion
+  - Configuration options testing
+  - Roundtrip testing (parse → emit)
+  - Writer integration
+
+- **api.yaml** - API constructor and method tests
+  - Constructor validation
+  - Method behavior and state changes
+  - Panic conditions
+  - Cleanup verification
+
+- **yaml.yaml** - Utility function tests
+  - Enum String() methods
+  - Style accessor methods
+
+- **reader.yaml** - Reader/input handling tests
+  - Encoding detection (UTF-8, UTF-16LE, UTF-16BE)
+  - Buffer management
+  - Error handling
+
+- **writer.yaml** - Writer/output handling tests
+  - Buffer flushing
+  - Output handlers (string, io.Writer)
+  - Error conditions
+
+- **yamlprivate.yaml** - Character classification tests
+  - Character type predicates (isAlpha, isDigit, isHex, etc.)
+  - Character conversion functions (asDigit, asHex, width)
+  - Unicode handling
+
+- **loader.yaml** - Data loader scalar resolution tests
+  - Numeric type resolution (integers, floats)
+  - Boolean and null value handling
+  - String vs numeric type disambiguation
+  - Mixed-type collections
+
+### Test Framework Implementation
+
+The test framework is implemented in `yamldatatest_loader.go` and `yamldatatest_test.go`:
+
+**Core functions**:
+- `LoadYAML(data []byte) (interface{}, error)` - Parses YAML using libyaml parser with scalar type resolution (exported)
+- `UnmarshalStruct(target interface{}, data map[string]interface{}) error` - Populates structs (exported)
+- `LoadTestCases(filename string) ([]TestCase, error)` - Loads and parses test YAML files
+- `coerceScalar(value string) interface{}` - Resolves scalar strings to appropriate Go types (int, float64, bool, nil, string)
+
+**Core types**:
+- `TestCase` struct - Umbrella structure containing fields for all test types
+  - Uses `interface{}` for flexible field types
+  - Post-processing converts generic fields to specific types
+
+**Post-processing**:
+After loading, the framework processes test data:
+- Converts `Want` (interface{}) to `WantEvents`, `WantTokens`, or `WantSpecs` based on test type
+- Converts `Want` (interface{}) to `WantContains` (handles both scalar and sequence)
+- Converts `Checks` to field validation specifications
+
+### Test Types
+
+#### Scanner Tests
+
+**scan-tokens** - Verify token sequence
+
+```yaml
+- scan-tokens:
+    name: Simple scalar
+    yaml: |-
+      hello
+    want:
+    - STREAM_START_TOKEN
+    - SCALAR_TOKEN
+    - STREAM_END_TOKEN
+```
+
+**scan-tokens-detailed** - Verify token properties
+
+```yaml
+- scan-tokens-detailed:
+    name: Single quoted scalar
+    yaml: |-
+      'hello world'
+    want:
+    - STREAM_START_TOKEN
+    - SCALAR_TOKEN:
+        style: SINGLE_QUOTED_SCALAR_STYLE
+        value: hello world
+    - STREAM_END_TOKEN
+```
+
+**scan-error** - Verify error detection
+
+```yaml
+- scan-error:
+    name: Invalid character
+    yaml: "\x01"
+```
+
+#### Parser Tests
+
+**parse-events** - Verify event sequence
+
+```yaml
+- parse-events:
+    name: Simple mapping
+    yaml: |
+      key: value
+    want:
+    - STREAM_START_EVENT
+    - DOCUMENT_START_EVENT
+    - MAPPING_START_EVENT
+    - SCALAR_EVENT
+    - SCALAR_EVENT
+    - MAPPING_END_EVENT
+    - DOCUMENT_END_EVENT
+    - STREAM_END_EVENT
+```
+
+**parse-events-detailed** - Verify event properties
+
+```yaml
+- parse-events-detailed:
+    name: Anchor and alias
+    yaml: |
+      - &anchor value
+      - *anchor
+    want:
+    - STREAM_START_EVENT
+    - DOCUMENT_START_EVENT
+    - SEQUENCE_START_EVENT
+    - SCALAR_EVENT:
+        anchor: anchor
+        value: value
+    - ALIAS_EVENT:
+        anchor: anchor
+    - SEQUENCE_END_EVENT
+    - DOCUMENT_END_EVENT
+    - STREAM_END_EVENT
+```
+
+**parse-error** - Verify error detection
+
+```yaml
+- parse-error:
+    name: Error state
+    yaml: |
+      key: : invalid
+```
+
+#### Emitter Tests
+
+**emit** - Emit events and verify output contains expected strings
+
+```yaml
+- emit:
+    name: Simple scalar
+    data:
+    - STREAM_START_EVENT:
+        encoding: UTF8_ENCODING
+    - DOCUMENT_START_EVENT:
+        implicit: true
+    - SCALAR_EVENT:
+        value: hello
+        implicit: true
+        style: PLAIN_SCALAR_STYLE
+    - DOCUMENT_END_EVENT:
+        implicit: true
+    - STREAM_END_EVENT
+    want: hello
+```
+
+**emit-config** - Emit with configuration
+
+```yaml
+- emit-config:
+    name: Custom indent
+    conf:
+      indent: 4
+    data:
+    - STREAM_START_EVENT:
+        encoding: UTF8_ENCODING
+    - DOCUMENT_START_EVENT:
+        implicit: true
+    - MAPPING_START_EVENT:
+        implicit: true
+        style: BLOCK_MAPPING_STYLE
+    # ... more events
+    want: key
+```
+
+**roundtrip** - Parse → emit, verify output
+
+```yaml
+- roundtrip:
+    name: Roundtrip
+    yaml: |
+      key: value
+      list:
+        - item1
+        - item2
+    want:
+    - key
+    - value
+    - item1
+```
+
+**emit-writer** - Emit to io.Writer
+
+```yaml
+- emit-writer:
+    name: Writer
+    data:
+    - STREAM_START_EVENT:
+        encoding: UTF8_ENCODING
+    # ... more events
+    want: test
+```
+
+#### API Tests
+
+**api-new** - Test constructors
+
+```yaml
+- api-new:
+    name: New parser
+    with: NewParser
+    test:
+    - nil: [raw-buffer, false]
+    - cap: [raw-buffer, 512]
+    - nil: [buffer, false]
+    - cap: [buffer, 1536]
+```
+
+**api-method** - Test methods and field state
+
+```yaml
+- api-method:
+    name: Parser set input string
+    with: NewParser
+    byte: true
+    call: [SetInputString, 'key: value']
+    test:
+    - eq: [input, 'key: value']
+    - eq: [input-pos, 0]
+    - nil: [read-handler, false]
+```
+
+**api-panic** - Test methods that should panic
+
+```yaml
+- api-panic:
+    name: Parser set input string twice
+    with: NewParser
+    byte: true
+    init: [SetInputString, first]
+    call: [SetInputString, second]
+    want: must set the input source only once
+```
+
+**api-delete** - Test cleanup
+
+```yaml
+- api-delete:
+    name: Parser delete
+    with: NewParser
+    byte: true
+    init: [SetInputString, test]
+    test:
+    - len: [input, 0]
+    - len: [buffer, 0]
+```
+
+**api-new-event** - Test event constructors
+
+```yaml
+- api-new-event:
+    name: New stream start event
+    call: [NewStreamStartEvent, UTF8_ENCODING]
+    test:
+    - eq: [Type, STREAM_START_EVENT]
+    - eq: [encoding, UTF8_ENCODING]
+```
+
+#### Utility Tests
+
+**enum-string** - Test String() methods of enums
+
+```yaml
+- enum-string:
+    name: Scalar style plain
+    enum: [ScalarStyle, PLAIN_SCALAR_STYLE]
+    want: Plain
+```
+
+**style-accessor** - Test style accessor methods
+
+```yaml
+- style-accessor:
+    name: Event scalar style
+    test: [ScalarStyle, DOUBLE_QUOTED_SCALAR_STYLE]
+```
+
+#### Loader Tests
+
+**scalar-resolution** - Test scalar type resolution
+
+```yaml
+- scalar-resolution:
+    name: Positive integer
+    yaml: "42"
+    want: 42
+
+- scalar-resolution:
+    name: Negative float
+    yaml: "-2.5"
+    want: -2.5
+```
+
+**Resolution order**:
+1. Boolean (true, false)
+2. Null (null keyword only)
+3. Hexadecimal integer (0x prefix)
+4. Float (contains .)
+5. Decimal integer
+6. String (fallback)
+
+### Common Keys in Test YAML Files
+
+Test cases use a **type-as-key** format where the test type is the map key:
+
+```yaml
+- test-type:
+    name: Test case name
+    # ... other fields
+```
+
+**Common fields**:
+- **name** - Test case name (title case convention)
+- **yaml** - Input YAML string to test
+- **want** - Expected result (format varies by test type)
+  - For api-panic: string containing expected panic message substring
+  - For scan-error/parse-error: boolean (defaults to true if omitted; set to false if no error expected)
+  - For enum-string: string representing expected String() output
+  - For other types: varies (may be sequence or scalar)
+- **data** - For emitter tests: list of event specifications to emit
+- **conf** - For emitter config tests: emitter configuration options
+- **with** - For API tests: constructor name (NewParser, NewEmitter)
+- **call** - For API tests: method call [MethodName, arg1, arg2, ...]
+- **init** - For API panic tests: setup method call before main method
+- **byte** - For API tests: boolean flag to convert string args to []byte
+- **test** - For API tests: list of field validation checks in format `operator: [field, value]` where operator is one of: nil, cap, len, eq, gte, len-gt.
+- **test** - For style-accessor tests: array of [Method, STYLE] where Method is the accessor method (e.g., ScalarStyle) and STYLE is the style constant (e.g., DOUBLE_QUOTED_SCALAR_STYLE).
+- **enum** - For enum tests: array of [Type, Value] where Type is the enum type (e.g., ScalarStyle) and Value is the constant (e.g., PLAIN_SCALAR_STYLE)
+
+**Note on scalar type resolution**: Unquoted scalar values in test data are automatically resolved to appropriate Go types (int, float64, bool, nil) by the `LoadYAML` function. Quoted scalars remain as strings.
+
+### Running Tests
+
+```bash
+# Run all tests in the package
+go test ./internal/libyaml
+
+# Run specific test file
+go test ./internal/libyaml -run TestScanner
+go test ./internal/libyaml -run TestParser
+go test ./internal/libyaml -run TestEmitter
+go test ./internal/libyaml -run TestAPI
+go test ./internal/libyaml -run TestYAML
+go test ./internal/libyaml -run TestLoader
+
+# Run specific test case (using subtest name)
+go test ./internal/libyaml -run TestScanner/Block_sequence
+go test ./internal/libyaml -run TestParser/Anchor_and_alias
+go test ./internal/libyaml -run TestEmitter/Flow_mapping
+go test ./internal/libyaml -run TestLoader/Scientific_notation_lowercase_e
+
+# Run with verbose output
+go test -v ./internal/libyaml
+
+# Run with coverage
+go test -cover ./internal/libyaml
+```
--- a/internal/libyaml/api_test.go
+++ b/internal/libyaml/api_test.go
@@ -1,0 +1,627 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package libyaml
+
+import (
+	"fmt"
+	"reflect"
+	"strconv"
+	"strings"
+	"testing"
+
+	"go.yaml.in/yaml/v4/internal/testutil/assert"
+)
+
+func TestAPI(t *testing.T) {
+	RunTestCases(t, "api.yaml", map[string]TestHandler{
+		"api-new":       runAPINewTest,
+		"api-method":    runAPIMethodTest,
+		"api-panic":     runAPIPanicTest,
+		"api-delete":    runAPIDeleteTest,
+		"api-new-event": runAPINewEventTest,
+	})
+}
+
+func runAPINewTest(t *testing.T, tc TestCase) {
+	t.Helper()
+
+	obj := createObject(t, tc.Constructor)
+	runFieldChecks(t, obj, tc.Checks)
+}
+
+func runAPIMethodTest(t *testing.T, tc TestCase) {
+	t.Helper()
+
+	obj := createObject(t, tc.Constructor)
+
+	// Run setup if any
+	if tc.Setup != nil {
+		if setupList, ok := tc.Setup.([]interface{}); ok && len(setupList) > 0 {
+			callMethodFromList(t, obj, setupList, tc.Bytes)
+		}
+	}
+
+	// Call the main method
+	callMethodFromList(t, obj, tc.Method, tc.Bytes)
+
+	// Run checks
+	runFieldChecks(t, obj, tc.Checks)
+}
+
+func runAPIPanicTest(t *testing.T, tc TestCase) {
+	t.Helper()
+
+	obj := createObject(t, tc.Constructor)
+
+	// Run setup if any
+	if tc.Setup != nil {
+		if setupList, ok := tc.Setup.([]interface{}); ok && len(setupList) > 0 {
+			callMethodFromList(t, obj, setupList, tc.Bytes)
+		}
+	}
+
+	// The main method call should panic
+	// Want can be either a string or a single-element sequence
+	var wantMsg string
+	switch v := tc.Want.(type) {
+	case string:
+		wantMsg = v
+	case []interface{}:
+		if len(v) > 0 {
+			msg, ok := v[0].(string)
+			assert.Truef(t, ok, "Want[0] should be string, got %T", v[0])
+			wantMsg = msg
+		} else {
+			t.Fatalf("Want slice is empty, expected at least one element")
+		}
+	default:
+		t.Fatalf("want must be a string or sequence, got %T", tc.Want)
+	}
+	assert.PanicMatchesf(t, wantMsg, func() {
+		callMethodFromList(t, obj, tc.Method, tc.Bytes)
+	}, "Expected panic: %s", wantMsg)
+}
+
+func runAPIDeleteTest(t *testing.T, tc TestCase) {
+	t.Helper()
+
+	obj := createObject(t, tc.Constructor)
+
+	// Run setup if any
+	if tc.Setup != nil {
+		if setupList, ok := tc.Setup.([]interface{}); ok && len(setupList) > 0 {
+			callMethodFromList(t, obj, setupList, tc.Bytes)
+		}
+	}
+
+	// Call Delete method
+	callMethodFromList(t, obj, []interface{}{"Delete"}, false)
+
+	// Run checks after delete
+	runFieldChecks(t, obj, tc.Checks)
+}
+
+func runAPINewEventTest(t *testing.T, tc TestCase) {
+	t.Helper()
+
+	event := createEventFromList(t, tc.Method, tc.Bytes)
+	runFieldChecks(t, &event, tc.Checks)
+}
+
+// createObject creates a Parser or Emitter based on constructor name
+func createObject(t *testing.T, constructor string) interface{} {
+	t.Helper()
+	switch constructor {
+	case "NewParser":
+		p := NewParser()
+		return &p
+	case "NewEmitter":
+		e := NewEmitter()
+		return &e
+	default:
+		t.Fatalf("unknown constructor: %s", constructor)
+	}
+	return nil
+}
+
+// createEventFromList creates an Event from a method list [constructor, args...]
+func createEventFromList(t *testing.T, methodList []interface{}, useBytes bool) Event {
+	t.Helper()
+	if len(methodList) == 0 {
+		t.Fatalf("empty method list")
+	}
+
+	constructor, ok := methodList[0].(string)
+	if !ok {
+		t.Fatalf("constructor should be string, got %T", methodList[0])
+	}
+	args := methodList[1:]
+
+	switch constructor {
+	case "NewStreamStartEvent":
+		if len(args) != 1 {
+			t.Fatalf("%s expects 1 argument, got %d", constructor, len(args))
+		}
+		encoding := parseArg(t, args[0])
+		return NewStreamStartEvent(Encoding(encoding))
+	case "NewStreamEndEvent":
+		if len(args) != 0 {
+			t.Fatalf("%s expects 0 arguments, got %d", constructor, len(args))
+		}
+		return NewStreamEndEvent()
+	case "NewDocumentEndEvent":
+		if len(args) != 1 {
+			t.Fatalf("%s expects 1 argument, got %d", constructor, len(args))
+		}
+		implicit := parseBoolArg(t, args[0])
+		return NewDocumentEndEvent(implicit)
+	case "NewAliasEvent":
+		if len(args) != 1 {
+			t.Fatalf("%s expects 1 argument, got %d", constructor, len(args))
+		}
+		anchor := parseStringArg(t, args[0], useBytes)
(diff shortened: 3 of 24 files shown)


go.mod
```diff
@@ -1,3 +1,7 @@
 module go.yaml.in/yaml/v4
 
 go 1.18
+
+require go.yaml.in/yaml/v3 v3.0.4
+
+replace go.yaml.in/yaml/v3 => ./.cache/yaml-v3
```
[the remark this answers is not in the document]
ingydotnet: The Makefile does do that already,

internal/libyaml/README.md
```diff
@@ -0,0 +1,493 @@
...
+  - STREAM_START_TOKEN
+  - SCALAR_TOKEN
+  - STREAM_END_TOKEN
+```
+
+**scan-tokens-detailed** - Verify token properties
+
+```yaml
+- name: Single quoted scalar
+  type: scan-tokens_detailed
```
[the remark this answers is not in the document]
ingydotnet: fixed in 2b43629

internal/libyaml/reader.go
```diff
@@ -363,31 +363,18 @@ func (parser *Parser) updateBuffer(length int) bool {
...
-			// DEL, C1 control
-			// incompatible with YAML versions <= 1.1
-			case value >= 0x7F && value <= 0x9F:
-			// and Basic Multilingual Plane (BMP),
+			case value == 0x85:
 			case value >= 0xA0 && value <= 0xD7FF:
-			// Additional Unicode Areas
 			case value >= 0xE000 && value <= 0xFFFD:
-			// 32 bit set
 			case value >= 0x10000 && value <= 0x10FFFF:
```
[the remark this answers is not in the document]
ingydotnet: Fixed.

.github/workflows/check-commit-message.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
     if: github.event.pull_request.user.login != 'dependabot[bot]'
     steps:
     - name: Checkout code
-      uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3  # v6.0.0
```
colinjlacy: These changes to the build step versions seem like they may have been unintentional. If so, can you revert them?

GNUmakefile
```diff
@@ -65,8 +69,8 @@ count ?= 1
 
 
 # Test rules:
-test: $(GO-DEPS)
-	go test$(if $v, -v) -vet=off .
+test: $(GO-DEPS) $(V3-DIR)
```
colinjlacy: I have no concerns about this change, but it looks like it may be pointing to version-specific test runs in the future. Is that the case?

decode_test.go
```diff
@@ -927,15 +927,6 @@ var unmarshalTests = []struct {
 		"true #" + strings.Repeat(" ", 512*3),
 		"true",
 	},
-	// C1 Control characters
```
colinjlacy, line 930: Please be sure not to strip these out

decode.go
```diff
@@ -216,8 +216,7 @@ func (p *parser) alias() *Node {
 	n := p.node(AliasNode, "", "", string(p.event.Anchor))
 	n.Alias = p.anchors[n.Value]
 	if n.Alias == nil {
-		msg := fmt.Sprintf("unknown anchor '%s' referenced", n.Value)
-		fail(&ParserError{msg, n.Line, n.Column})
```
colinjlacy, line 220: Do we want to remove the line, column numbers, and error type here?

internal/libyaml/emitter.go
```diff
@@ -1881,7 +1881,9 @@ func (emitter *Emitter) writeDoubleQuotedScalar(value []byte, allow_breaks bool)
 }
 
 func (emitter *Emitter) writeBlockScalarHints(value []byte) error {
-	if isSpace(value, 0) || isLineBreak(value, 0) {
+	if isSpace(value, 0) {
```
colinjlacy: Remember to remove commented-out code

go.mod
ccoVeille: There is something really strange to me here

Having the v4 to depends on v3 🤔

Either things should be shared between version, and we need something like an external repository like yts, either there is an error here.

internal/libyaml/api_test.go
```diff
@@ -0,0 +1,461 @@
...
+	case "NewParser":
+		p := NewParser()
+		return &p
+	case "NewEmitter":
+		e := NewEmitter()
+		return &e
+	default:
+		panic("unknown constructor: " + constructor)
+	}
+}
```
ccoVeille, line 125: This is a strange way to do. Is it because you wanted to mimics Go generics?

Without generics, usually such a thing in test is done by having an interface{} field and having this field sets to NewWhatever

I feel like I don't get the purpose you are trying to achieve.

internal/libyaml/reader_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+func runReaderSetErrorTest(t *testing.T, tc TestCase) {
+	t.Helper()
+
+	parser := NewParser()
+
+	// args should be [problem, offset, value]
+	problem := tc.Args[0].(string)
+	offset := tc.Args[1].(int)
+	value := tc.Args[2].(int)
```
ccoVeille: The Len of Args should be tested and anything else than 3 should stop processing further

internal/libyaml/testdata/api.yaml
```diff
@@ -0,0 +1,239 @@
+# SPDX-License-Identifier: Apache-2.0
```
ccoVeille, line 1: In Go testdata are in testdata folder by convention. This is a special folder. There should be no hypen

internal/libyaml/scanner_test.go
```diff
@@ -0,0 +1,101 @@
...
+		switch v := tc.Want.(type) {
+		case bool:
+			wantError = v
+		case []interface{}:
+			if len(v) > 0 {
+				if boolVal, ok := v[0].(bool); ok {
+					wantError = boolVal
+				}
+			}
+		}
```
ccoVeille, line 88: I'm surprised by the usage of type assertion everywhere in this test suite that you are building.

I have the strange impression that tests suite would be more complicated than the code it tests

.github/workflows/check-commit-message.yaml
ingydotnet: It think a couple unintentional things may have happened. That's why I put it in draft mode. Thank for looking. I'll patch things soon.

internal/libyaml/reader_test.go
```diff
@@ -0,0 +1,140 @@
...
+	parser := NewParser()
+	parser.SetInputString(tc.Input)
+
+	// Apply any setup
+	if tc.Setup != nil {
+		applySetup(t, &parser, tc.Setup)
+	}
+
+	// Get the length argument
+	length := tc.Args[0].(int)
```
ccoVeille: There are too many type assertion for me in this code. Also, a lot of them are not using the ", ok" pattern it sounds a source of panic for anyone who will try to add new tests or update existing one.

decode_test.go
ingydotnet: I think the way I started based on your PR, that change might not have been in you branch. I already fix that one.

internal/libyaml/yaml_data_test.go
```diff
@@ -0,0 +1,963 @@
...
+
+	// style_accessor tests
+	Accessor   string    `yaml:"method"`
+	StyleValue IntOrStr `yaml:"style"`
+
+	// api_method, api_panic, api_delete tests, and reader tests
+	Bytes  bool          `yaml:"bytes"`
+	Method []interface{} `yaml:"call"`
+	Setup  interface{}   `yaml:"setup"` // Can be []interface{} (api tests) or map[string]interface{} (reader tests)
+}
```
ccoVeille: I feel like the whole point of importing yaml v3 in test is to use yaml as a format to test yaml.

It sounds like to me we reached the aphorism: When you have an hammer everything is a nail. While the idea is elegant, it brings something too complicated. We don't have to bring yaml to test yaml. It's inception.

I feel like you imported yaml v3 because you would have face a cycle dependency if you tried to import the package at root level.

Using the json format would have sounded better, as it has no dependency.

My problem here is because the way Go handle go.mod and go.sum

It means that something importing yaml/v4 will see yaml/v3 as a dependency, and go.sum of someone importing the v4 will not only have the v3 as a dependency, but the v3 dependencies will also be added to anyone using v4.

So anything we managed to remove from v4 by cleaning its dependencies, will come back because v3 still have them, and so they will stay in v4 for ever.

I might be wrong here, I don't have a computer to test.

Also, having v4 tests depending on v3 means that if something is broken in v3, and we want to fix in v4, we will have to either fix v3 first because otherwise v4 tests would never pass, or patch test with hack to make it pass.

ccoVeille: Please note that as most of the libyaml methods are exported 

Everything in libyaml could be tested at the root level, and so yaml package would be available, and so no need to use the v3

But because the way Go considers then libyaml there would have 0% code coverage.

So it might something we don't want.

go.mod
ingydotnet: Let me explain.

The tests are written in basic yaml and so they need a yaml decoder to decode them, in order to work. This could be any yaml loader, not even ours. 

Of course I wanted it to be ours. I couldn't use the current version 4 as it caused a circular dep. I didn't need v4 because this is just simple yaml where I need the data. I decided to use the most recent v3 of ours. It is perfectly fine.

I could have just depended on it but we don't have any deps for this library so that would be weird. Especially since we have all the code we need right here. So I used a replace and had the Makefile make sure it was on disk.

I find it a pretty great solution, but open to suggestions...

.github/workflows/check-commit-message.yaml
ingydotnet: Removed unintentional changes based on using a fetch of your PR that didn't have some commits in it.

GNUmakefile
ingydotnet: It's just using a stable version of our YAML decoder for driving the yaml based tests.
Bootstrapping...

decode.go
ingydotnet: fixed. sorry.

internal/libyaml/testdata/api.yaml
ingydotnet: Will fix. I just made up `test-data` myself. I had no idea that `testdata` was a convention. Was going to use `data` at first. What a coincidence.

internal/libyaml/yaml_data_test.go
ingydotnet: v3 is a known stable yaml decoder and all we need. there is no chance of things going badly with it.

ingydotnet: > Using the json format would have sounded better, as it has no dependency.

YAML is an excellent format for defining test data. JSON wouldn't really work well because it can't do multiline values in a readable way. The beauty of these tests is that the YAML being tested looks exactly as you would see it in real life.

ingydotnet: > Also, having v4 tests depending on v3 means that if something is broken in v3, and we want to fix in v4, we will have to either fix v3 first because otherwise v4 tests would never pass, or patch test with hack to make it pass.

This will never happen. This is just dead simple YAML. I can't imagine ever triggering a bug in v3 with this test data.

internal/libyaml/testdata/api.yaml
ingydotnet: Fixed in ab3f733

internal/libyaml/reader_test.go
ingydotnet: Fixed in 8c51bbf 3a25a63 76ca2c1 5d077c8

internal/libyaml/yaml_data_test.go
ingydotnet: Got rid of replace and made this a test only dep in 9ed2252

go.mod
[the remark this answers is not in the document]
ingydotnet: Fixed in 9ed2252

ingydotnet: Reworked this in 9ed2252 to work much better

internal/libyaml/emitter.go
ingydotnet: Fixed in 86919b1

internal/libyaml/api_test.go
ingydotnet: We are using a data driven approach. We use data strings that then map to functions in Go code (or could be functions in Python code). 
The data is not Go specific. Any data string names could have been chosen here. They just happen to look the same as the Go function names in this case...

internal/libyaml/scanner_test.go
ingydotnet: This complexity exists to enable:
- Language-agnostic test data (YAML files could be reused by other YAML libraries)
- Easy addition of new test cases without writing Go code
- Comprehensive API testing with minimal boilerplate

internal/libyaml/yamldatatest_test.go
```diff
@@ -0,0 +1,963 @@
...
+	v := reflect.ValueOf(obj)
+	if v.Kind() == reflect.Ptr {
+		v = v.Elem()
+	}
+	field := v.FieldByName(fieldName)
+	if !field.IsValid() {
+		panic(fmt.Sprintf("field %s not found", fieldName))
+	}
+	return field.Interface()
+}
```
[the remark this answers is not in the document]
ingydotnet: fixed

internal/libyaml/yamldatatest_test.go
```diff
@@ -0,0 +1,963 @@
...
+	}
+	if config.LineBreak != "" {
+		// Parse line break style if needed
+		switch config.LineBreak {
+		case "LN":
+			emitter.SetLineBreak(LN_BREAK)
+		case "CR":
+			emitter.SetLineBreak(CR_BREAK)
+		case "CRLF":
+			emitter.SetLineBreak(CRLN_BREAK)
```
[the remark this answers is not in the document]
ingydotnet: not a bug

internal/libyaml/reader_test.go
ingydotnet: Fixed in 7b7e371

internal/libyaml/yaml_data_test.go
ccoVeille: As I said, I'm fine with the idea of using yaml. The repository is a showcase of YAML, so it's a good idea.

But right now, I feel like it would be a very bad thing to add v3 as a dependency of the v4

I'll do a summary here, for everyone, the problems:
- the way Go works: something added as a dependency, even for tests, becomes a dependency of the code that would use go-yaml
- having whatever project or language depending on a previous version of the project sounds a problem
- adding tests to internal is something interesting, but not mandatory.
- if adding these tests leads to adding this dependency, I would prefer to do not test them.
- instead the code of libyaml could be tested at the root level.
- the code organization and the places where the code are tested could be changed, to try to solve things.
- if using yaml in tests bring such level of complexity it should not be used.

Let's wait for other people feedbacks, before going further.

ingydotnet: It's my understanding that Go module graph pruning will ensure that v3 is not a transitive dep of v4.
I'll test that. You should too.

ingydotnet: My testing so far indicates that v3 would show up in the dependency graph.
I'll try alternate layouts.

ingydotnet: I was able to get rid of v3 altogether by just using internals/libyaml functions to load the test YAML files.

See: 26318c5

ingydotnet:
I have a few more YAML gardening tasks to do before I make this ready again...

internal/libyaml/yamldatatest_loader.go
```diff
@@ -0,0 +1,399 @@
...
+		if _, err := fmt.Sscanf(value, "0x%x", &intVal); err == nil {
+			return intVal
+		}
+	}
+
+	// Try decimal int
+	var intVal int
+	if _, err := fmt.Sscanf(value, "%d", &intVal); err == nil {
+		return intVal
+	}
```
ccoVeille, line 40: Decimal value can be negative

I would detect the presence of a leading hyphen and use int, and use uint otherwise

internal/libyaml/yamldatatest_loader.go
```diff
@@ -0,0 +1,399 @@
...
+
+	// Default to string
+	return value
```
ccoVeille, line 43: We could have float too no ?

internal/libyaml/yaml_data_test.go
ccoVeille: Thanks

ccoVeille (commented):
Thanks for the refactoring.

It's way better and now I'm confident it can be merged at some point.

I saw your comment about the fact you might have a few refactoring to do, but I made some minor feedbacks

internal/libyaml/yamldatatest_loader.go
ingydotnet: does `%d` not handle negatives?

ingydotnet: Already works. Full testing in f688763

internal/libyaml/yamldatatest_loader.go
ingydotnet: float support added and tested in f688763

ingydotnet:
> I have a few more YAML gardening tasks to do before I make this ready again...

Gardening done. Made the YAML super clean. Gonna run by copilot once more then ask for team reviews.

internal/libyaml/yamldatatest_loader.go
ingydotnet: Oops. Forgot to add test files there. See d20111c

internal/libyaml/yamldatatest_loader.go
ingydotnet: See d20111c instead!

ingydotnet:
I squashed the history to a single commit. Ready for review.

internal/libyaml/yamldatatest_loader.go
ccoVeille: Indeed %d (scanf) is not \d (regexp)

Thanks for adding the test

internal/libyaml/testdata/emitter.yaml
ccoVeille, line 1: Now everything is done. I can tell I appreciate the format for defining the tests.

It would complete if not fully replace yts data at dom point depending on the direction you want to go.

internal/libyaml/testdata/api.yaml
ccoVeille, line 1: I would like to suggest removing the _test.yaml prefix to use .yaml one.

You are in testdata I feel like it's enough. Also the _test suffix sounds like a Go thing, while the definition of your tests are now language agnostic

internal/libyaml/api_test.go
```diff
@@ -0,0 +1,527 @@
...
+	}
+
+	// The main method call should panic
+	// Want can be either a string or a single-element sequence
+	var wantMsg string
+	switch v := tc.Want.(type) {
+	case string:
+		wantMsg = v
+	case []interface{}:
+		msg, ok := v[0].(string)
```
ccoVeille: A nil/empty slice would panic

This pattern can be found at various place. They should all be addressed

internal/libyaml/api_test.go
```diff
@@ -0,0 +1,527 @@
...
+
+func runAPINewEventTest(t *testing.T, tc TestCase) {
+	t.Helper()
+
+	event := createEventFromList(tc.Method, tc.Bytes)
+	runFieldChecks(t, &event, tc.Checks)
+}
+
+// createObject creates a Parser or Emitter based on constructor name
+func createObject(constructor string) interface{} {
```
ccoVeille: Here and everywhere where you use panic, I would pass t *testing.T as argument and call t.Fatal or t.Fatalf

internal/libyaml/api_test.go
```diff
@@ -0,0 +1,527 @@
...
+}
+
+// getField retrieves a field from a struct, handling special field names
+func getField(t *testing.T, v reflect.Value, fieldName string) reflect.Value {
+	t.Helper()
+
+	// Handle special field names like buffer-0, buffer-1
+	if strings.HasPrefix(fieldName, "buffer-") {
+		var bufferIndex int
+		_, err := fmt.Sscanf(fieldName, "buffer-%d", &bufferIndex)
```
ccoVeille, line 467: I saw the comment of invalid value being checked somewhere else

I didn't see where, but so my comment might be irrevelant

but I would like to make sure `buffer - - 42` (spaces added for clarity) won't accept -42 as buffer value

internal/libyaml/api_test.go
```diff
@@ -0,0 +1,527 @@
...
+
+	v := reflect.ValueOf(obj)
+	if v.Kind() == reflect.Ptr {
+		v = v.Elem()
+	}
+
+	for _, check := range checks {
+		// Handle nil checks
+		if len(check.Nil) >= 2 {
+			fieldName, _ := check.Nil[0].(string)
```
ccoVeille: Here and everywhere else I wouldn't ignore the ok value that the type assertion return and either handle it with a t.Fatal or add it to the condition

I'm a bit worried a non string field would be returned as "", false by the type assertion, and so empty string will be used as field name leading to either have a wrong message being reported, or a test to pass when it wasn't supposed.

Please note,maybe it cannot happen. But when I see things that are ignored, I'm suspicious

internal/libyaml/api_test.go
```diff
@@ -0,0 +1,527 @@
...
+			field := getField(t, v, fieldName)
+			if field.IsValid() && minLen > 0 {
+				if field.Len() <= minLen {
+					t.Errorf("%s len = %d, want > %d", fieldName, field.Len(), minLen)
+				}
+			}
+		}
+
+		// Handle eq checks
+		if len(check.Eq) >= 2 {
```
ccoVeille: Is there a valid case when having more than 2 arguments shouldn't be considered as an error?

The test suite shouldbe human error prone

internal/libyaml/api_test.go
```diff
@@ -0,0 +1,527 @@
...
+	}
+	return field
+}
+
+// getIntValue extracts an integer value from a field
+func getIntValue(t *testing.T, field reflect.Value, fieldName string) int {
+	t.Helper()
+
+	if field.CanInterface() {
+		switch v := field.Interface().(type) {
```
ccoVeille: I thought reflection would give you access to the type without doing type assertion

internal/libyaml/api_test.go
```diff
@@ -0,0 +1,527 @@
...
+			return int(v)
+		default:
+			t.Fatalf("%s: expected int field, got %T", fieldName, v)
+		}
+	}
+
+	// For unexported fields
+	switch field.Kind() {
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		return int(field.Int())
```
ccoVeille, line 494: Isn't there a CanInt method that you could use ?

internal/libyaml/api_test.go
```diff
@@ -0,0 +1,527 @@
...
+	}
+
+	field = getField(t, v, fieldName)
+	if !field.IsValid() {
+		return
+	}
+
+	// Parse constant if it's a string that looks like a constant name
+	var expectedInt int
+	hasExpectedInt := false
```
ccoVeille: This would be the cannoical way in Go

```suggestion
	var hasExpectedInt bool
```

internal/libyaml/api_test.go
```diff
@@ -0,0 +1,527 @@
...
+	expected := expectedValue
+	if str, ok := expectedValue.(string); ok && looksLikeConstant(str) {
```
ccoVeille, line 559: A switch on type assertion should be considered

internal/libyaml/api_test.go
```diff
@@ -0,0 +1,527 @@
...
+		hasExpectedInt = true
+	}
+
+	// Get value based on type (handle unexported fields)
+	var got interface{}
+
+	if field.CanInterface() {
+		// For exported fields, convert expected to field's type
+		if hasExpectedInt {
+			expected = reflect.ValueOf(expectedInt).Convert(field.Type()).Interface()
```
ccoVeille, line 573: You should use CanConvert here and report an error if it's false

Also once converted, it's a good thing to check CanInterface on it

internal/libyaml/api_test.go
```diff
@@ -0,0 +1,527 @@
...
+	}
+}
+
+// parseBoolArg parses a boolean argument
+func parseBoolArg(arg interface{}) bool {
+	switch v := arg.(type) {
+	case bool:
+		return v
+	case string:
+		return v == "true"
```
ccoVeille: So anything that is a string not equal to `"true"` is false?

So TRUE is false, "foo" is false.

I would check for string type and return an error for anything else than `"true"` or `"false"`

You do it right in parseConstant

internal/libyaml/api_test.go
```diff
@@ -0,0 +1,527 @@
...
+	// Handle boolean values
+	if name == "true" {
+		return 1
+	}
+	if name == "false" {
+		return 0
+	}
+
+	// Try parsing as int first
+	if val, err := strconv.Atoi(name); err == nil {
```
ccoVeille, line 316: Be careful of int overflow.

An uint larger than math.MaxInt would overflow and you will have unexpected results of error message.

A test should cover this.

internal/libyaml/api_test.go
```diff
@@ -0,0 +1,527 @@
...
+		return val
+	}
+
+	// Use IntOrStr to parse other constants
+	ios := IntOrStr{}
+	err := ios.FromValue(name)
+	if err != nil {
+		panic(fmt.Sprintf("failed to parse constant %q: %v", name, err))
```
ccoVeille: As mentioned before, here and everywhere, mrgjid should receive a *testing.T, call t.Helper() and use t.Fatalf in such conditions

Panic should be avoided

internal/libyaml/api_test.go
ccoVeille: Same logic about the 2and argument not being checked and return an error

internal/libyaml/api_test.go
ccoVeille: I would expect to see a call to Kind() and not a fyie assertion

internal/libyaml/api_test.go
ccoVeille, line 494: CanInt should be considered

internal/libyaml/api_test.go
```diff
@@ -0,0 +1,527 @@
...
+			t.Fatalf("%s: expected int field, got %T", fieldName, v)
+		}
+	}
+
+	// For unexported fields
+	switch field.Kind() {
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		return int(field.Int())
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+		return int(field.Uint())
```
ccoVeille, line 496: CanUint should be considered

internal/libyaml/api_test.go
ccoVeille: OverflowInt also

internal/libyaml/api_test.go
```diff
@@ -0,0 +1,527 @@
...
+			return
+		} else {
+			t.Errorf("%s: buffer field is not a byte slice", fieldName)
+			return
+		}
+	}
+
+	field = getField(t, v, fieldName)
+	if !field.IsValid() {
+		return
```
ccoVeille, line 552: Should wevsee a comment here or a call to t.Error?

internal/libyaml/yamldatatest_loader.go
```diff
@@ -0,0 +1,503 @@
...
+package libyaml
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+// coerceScalar converts a YAML scalar string to an appropriate Go type
+func coerceScalar(value string) interface{} {
```
ccoVeille, line 12: Do you (or yaml) support numerical separator like `1_000_000` for one million?

internal/libyaml/yamldatatest_loader.go
```diff
@@ -0,0 +1,503 @@
...
+
+	// Handle conversions
+	switch fieldType.Kind() {
+	case reflect.String:
+		if str, ok := value.(string); ok {
+			field.SetString(str)
+			return nil
+		}
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		if intVal, ok := value.(int); ok {
```
ccoVeille: Int64 will overflow int on i386 arch here (int32)

CanInt and/or OverflowInt should be considered

internal/libyaml/yamldatatest_loader.go
```diff
@@ -0,0 +1,503 @@
...
+			return nil
+		}
+	case reflect.Bool:
+		if boolVal, ok := value.(bool); ok {
+			field.SetBool(boolVal)
+			return nil
+		}
+	case reflect.Float32, reflect.Float64:
+		if floatVal, ok := value.(float64); ok {
+			field.SetFloat(floatVal)
```
ccoVeille: CanFloat should be considered

internal/libyaml/yamldatatest_loader.go
```diff
@@ -0,0 +1,503 @@
...
+			} else {
+				m, ok = item.(map[string]interface{})
+				if !ok {
+					return fmt.Errorf("slice element %d: expected map for struct, got %T", i, item)
+				}
+				// Normalize type-as-key format
+				m = normalizeTypeAsKey(m)
+			}
+
+			if err := UnmarshalStruct(elem.Addr().Interface(), m); err != nil {
```
ccoVeille, line 471: When using Addr, CanAddr is a good friend

internal/libyaml/yamldatatest_test.go
```diff
@@ -0,0 +1,1045 @@
...
+	case "ALIAS_TOKEN":
+		return ALIAS_TOKEN
+	case "ANCHOR_TOKEN":
+		return ANCHOR_TOKEN
+	case "TAG_TOKEN":
+		return TAG_TOKEN
+	case "SCALAR_TOKEN":
+		return SCALAR_TOKEN
+	default:
+		panic(fmt.Sprintf("unknown token type: %s", s))
```
ccoVeille: Here also panic should be repldfed by a t.Fatal

encode_test.go
```diff
@@ -136,6 +137,18 @@ var marshalTests = []struct {
 		map[string]any{"v": ""},
 		"v: \"\"\n",
 	},
+
+	// Issue #65: https://github.com/yaml/go-yaml/issues/65 - Validate handling newlines at the start.
+	{
+		map[string]any{"v": "\nhi"},
```
ccoVeille: This seems to come from another PR

internal/libyaml/testdata/api.yaml
ingydotnet: Yeah at first the `foo_test.go` and `foo_test.yaml` files were in the same directory, so it made sense that they have the same base name. But under `testdata/` we can drop the `_test`. 

Fixed in c2e5bb1

internal/libyaml/api_test.go
ingydotnet: Fixed in 69894a0

internal/libyaml/api_test.go
ingydotnet: Fixed in 68ac0fc

internal/libyaml/yamldatatest_test.go
ingydotnet: Fixed in 68ac0fc

internal/libyaml/api_test.go
ingydotnet: Fixed in 23e0334

internal/libyaml/api_test.go
ingydotnet: Fixed in 68ac0fccb9fc56ec01b92192344f064a817ad246

internal/libyaml/api_test.go
ingydotnet: Fixed in 23e03342d42a23cda04971c88bf9308d92054fc5

internal/libyaml/yamldatatest_loader.go
ingydotnet: No. This is only for internal test usage. If someone uses that in a test they can add support then.

encode_test.go
ingydotnet: Might be from #184
@colinjlacy ?

internal/libyaml/api_test.go
ingydotnet: Fixed in 2a77003

encode_test.go
colinjlacy: Nope, not me

colinjlacy: Could be Claude? See https://github.com/yaml/go-yaml/pull/194/files#diff-a10b737cb7f52611c26e4507f1bc652425a66b09346785c31ca4f0698c1a2fbaR1883

internal/libyaml/yamldatatest_loader.go
ingydotnet: Fixed in 5937a7b

internal/libyaml/yamldatatest_loader.go
ingydotnet: Fixed in 5937a7b

internal/libyaml/yamldatatest_loader.go
ingydotnet: Fixed in 5937a7b

internal/libyaml/scanner_test.go
```diff
@@ -0,0 +1,95 @@
...
+		expected = append(expected, ParseTokenType(t, tokenStr))
+	}
+
+	assert.Equalf(t, len(expected), len(types), "scanTokens() got %d tokens, want %d", len(types), len(expected))
+	for i, tt := range expected {
+		assert.Equalf(t, tt, types[i], "token[%d] = %v, want %v", i, types[i], tt)
+	}
+}
+
+func runScanTokensDetailedTest(t *testing.T, tc TestCase) {
```
colinjlacy, line 41: It's not obvious what the value is of having `runParseEventsTest` and `runParseEventsDetailedTest`. Maybe worth adding a comment to explain why both are needed?

internal/libyaml/api_test.go
ingydotnet: Fixed in 17feef8

internal/libyaml/api_test.go
ingydotnet: Fixed in 17feef8

internal/libyaml/api_test.go
ingydotnet: Fixed in 17feef8

internal/libyaml/api_test.go
ingydotnet: Fixed in 17feef8

internal/libyaml/api_test.go
ingydotnet: Fixed in 17feef8

internal/libyaml/yamldatatest_loader.go
```diff
@@ -0,0 +1,503 @@
...
+//	type TestCase struct {
+//	    Name string `yaml:"name"`
+//	    Data string `yaml:"data"`
+//	}
+//	var tc TestCase
+//	err := libyaml.UnmarshalStruct(&tc, map[string]interface{}{
+//	    "name": "test1",
+//	    "data": "hello",
+//	})
+func UnmarshalStruct(target interface{}, data map[string]interface{}) error {
```
colinjlacy, line 198: I get why you're doing this and it does make sense. Are we worried about any possibility of this becoming misaligned with the root level method?

internal/libyaml/api_test.go
ingydotnet: Fixed in 119b781

internal/libyaml/api_test.go
ingydotnet: Fixed in 119b781

internal/libyaml/api_test.go
ingydotnet: Fixed in 119b781

internal/libyaml/api_test.go
ingydotnet: Fixed in 119b781

internal/libyaml/api_test.go
ingydotnet: Fixed in 119b781

internal/libyaml/api_test.go
ingydotnet: Fixed in 9caf051

internal/libyaml/yamldatatest_loader.go
ingydotnet: I doubt it but we can worry about it if that happens.

internal/libyaml/scanner_test.go
ingydotnet: We can improve in a follow up PR

encode_test.go
ingydotnet: Reverted in 13bbebf

internal/libyaml/testdata/emitter.yaml
ingydotnet: I think I've addressed everything now. All passing.

ingydotnet:
@ccoVeille @colinjlacy 

I think I've addressed everything. 
I'm sure there's more to improve but I feel like this is solid.

Can you review again? 
If you find things that aren't blockers for this PR, consider filing new issues and we can address them quickly.

ingydotnet:
Squashed to a single commit again. Refs in resolved comments should still work...

internal/libyaml/yaml_test.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	// Want can be either a string or a single-element sequence
+	var want string
+	switch v := tc.Want.(type) {
+	case string:
+		want = v
+	case []interface{}:
+		if len(v) > 0 {
+			want = v[0].(string)
```
ccoVeille: there is a possible type assertion panic here, if the slice is not a slice of string

internal/libyaml/yamldatatest_loader.go
```diff
@@ -0,0 +1,546 @@
...
+		return nil
+	}
+
+	// Handle conversions
+	switch fieldType.Kind() {
+	case reflect.String:
+		if str, ok := value.(string); ok {
+			field.SetString(str)
+			return nil
+		}
```
ccoVeille, line 302: [CanSet](https://pkg.go.dev/reflect#Value.CanSet) should be used as the documentation of [SetString](https://pkg.go.dev/reflect#Value.SetString) says

internal/libyaml/yamldatatest_loader.go
```diff
@@ -0,0 +1,546 @@
...
+				return fmt.Errorf("value %d overflows %v", intVal, fieldType)
+			}
+			field.SetUint(u64)
+			return nil
+		}
+	case reflect.Bool:
+		if boolVal, ok := value.(bool); ok {
+			field.SetBool(boolVal)
+			return nil
+		}
```
ccoVeille: I feel it should be simply this

```suggestion
		field.SetBool(valueRefl.Bool())
```

internal/libyaml/yamldatatest_loader.go
```diff
@@ -0,0 +1,546 @@
...
+			if !field.CanFloat() {
+				return fmt.Errorf("field cannot store float values")
+			}
+			f64 := float64(intVal)
+			if field.OverflowFloat(f64) {
+				return fmt.Errorf("value %d overflows %v", intVal, fieldType)
+			}
+			field.SetFloat(f64)
+			return nil
+		}
```
ccoVeille: Unless I'm wrong this can be done, and then avoid type assertion

```suggestion
		if !field.CanFloat() {
			return fmt.Errorf("field cannot store float values")
		}
		f64 := valueRefl.Float()
		if field.OverflowFloat(f64) {
			return fmt.Errorf("value %f overflows %v", f64, fieldType)
		}
		field.SetFloat(f64)
		return nil
```

internal/libyaml/testdata/loader.yaml
```diff
@@ -0,0 +1,139 @@
...
+    want: 255
+
+- scalar-resolution:
+    name: Float with decimal point
+    yaml: "3.14"
+    want: 3.14
+
+- scalar-resolution:
+    name: Negative float
+    yaml: "-2.5"
```
ccoVeille, line 42: Could you add a test for an *integer* greater than math.MaxInt64 ?

For example `"9223372036854775808"`: 9223372036854775807 (math.MaxInt64) + 1

Same for an int lower than math.MinInt64 ?

For example `" -9223372036854775809"` : -9223372036854775808 (math.MinInt64) - 1

internal/libyaml/testdata/loader.yaml
```diff
@@ -0,0 +1,139 @@
...
+    want: 1.5
+
+- scalar-resolution:
+    name: Scientific notation lowercase e
+    yaml: "1.5e10"
+    want: 15000000000.0
+
+- scalar-resolution:
+    name: Scientific notation uppercase E
+    yaml: "1.5E10"
```
ccoVeille: Could you add a test with a *float* greater than math.MaxUint64 here ?

internal/libyaml/yamldatatest_loader.go
```diff
@@ -0,0 +1,546 @@
...
+import (
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+// coerceScalar converts a YAML scalar string to an appropriate Go type
+func coerceScalar(value string) interface{} {
+	// Try bool and null
+	switch value {
```
ccoVeille, line 14: Here it sounds oriented to YAML 1.2, no ?

Maybe a strings.ToLower here to support True TRUE

What about `y`, `f`, `on`, `off` also ?

Please note all of these could be reported as invalid here, with a specific error message

Same logic with NULL

internal/libyaml/yamldatatest_loader.go
```diff
@@ -0,0 +1,546 @@
...
+	if len(value) > 2 && (strings.HasPrefix(value, "0x") || strings.HasPrefix(value, "0X")) {
+		var intVal int
+		// Use %x to parse the hex digits after the 0x/0X prefix
+		if _, err := fmt.Sscanf(value[2:], "%x", &intVal); err == nil {
+			return intVal
+		}
+	}
+
+	// Try float (must check before int because %d will parse "1.5" as "1")
+	if strings.Contains(value, ".") || strings.ContainsAny(value, "eE") {
```
ccoVeille: This would sound logic

```suggestion
	if strings.ContainsAny(value, ".eE") {
```

internal/libyaml/yamldatatest_loader.go
```diff
@@ -0,0 +1,546 @@
...
+	}
+
+	// Try hex int (0x or 0X prefix)
+	if len(value) > 2 && (strings.HasPrefix(value, "0x") || strings.HasPrefix(value, "0X")) {
+		var intVal int
+		// Use %x to parse the hex digits after the 0x/0X prefix
+		if _, err := fmt.Sscanf(value[2:], "%x", &intVal); err == nil {
+			return intVal
+		}
+	}
```
ccoVeille: What about a simple

```suggestion
	// Try hex int (0x or 0X prefix)
	var intVal int
	// Use %x to parse the hex digits after the 0x/0X prefix
	if _, err := fmt.Sscanf(strings.ToLower(value), "0x%x", &intVal); err == nil {
		return intVal
	}
```

And so no hasprefix, and value[2:]

internal/libyaml/yamldatatest_loader.go
```diff
@@ -0,0 +1,546 @@
...
+			if !field.CanInt() {
+				return fmt.Errorf("field cannot store int values")
+			}
+			i64 := int64(intVal)
+			if field.OverflowInt(i64) {
+				return fmt.Errorf("value %d overflows %v", intVal, fieldType)
+			}
+			field.SetInt(i64)
+			return nil
+		}
```
ccoVeille, line 306: I would have expected something like this

```suggestion
	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
		if !field.CanInt() {
			return fmt.Errorf("field cannot store int values")
		}
		i64 := valueRefl.Int()
		if field.OverflowInt(i64) {
			return fmt.Errorf("value %v overflows %v", value, fieldType)
		}
		field.SetInt(i64)
		return nil
	}
```

I don't think there is a reason to use type assertion once you have the reflection value.

internal/libyaml/yamldatatest_loader.go
```diff
@@ -0,0 +1,546 @@
...
+			if intVal < 0 {
+				return fmt.Errorf("cannot convert negative int %d to %v", intVal, fieldType)
+			}
+			u64 := uint64(intVal)
+			if field.OverflowUint(u64) {
+				return fmt.Errorf("value %d overflows %v", intVal, fieldType)
+			}
+			field.SetUint(u64)
+			return nil
+		}
```
ccoVeille: I would suggest this here

```suggestion
		if !field.CanUint() {
			return fmt.Errorf("field cannot store uint values")
		}
		if !valueRefl.CanUint() {
			if valueRefl.Int() < 0 {
				return fmt.Errorf("cannot convert negative %v to %v", value, fieldType)
			}
			return fmt.Errorf("cannot convert %v to %v", value, fieldType)
		}
		u64 := valueRefl.Uint()
		if field.OverflowUint(u64) {
			return fmt.Errorf("value %v overflows %v", value, fieldType)
		}
		field.SetUint(u64)
		return nil
	}
```

internal/libyaml/yaml_test.go
ingydotnet: fixed in a165e0bf54612231a1c883bd59d365c1ab4cfab9

internal/libyaml/yamldatatest_loader.go
ingydotnet: Fixed in 963778d4248c580e3fd0d1de6f0117b1fe750acb

internal/libyaml/yamldatatest_loader.go
ingydotnet: `internal/libyaml/yamldatatest_loader.go` is only for internal test parsing and I don't need/want more than is currently needed. Technically we don't need floats I think but left them in.

internal/libyaml/testdata/loader.yaml
ingydotnet: `internal/libyaml/yamldatatest_loader.go` is only for internal test parsing and I don't need/want more than is currently needed. Technically we don't need floats I think but left them in.

internal/libyaml/testdata/loader.yaml
ingydotnet: `internal/libyaml/yamldatatest_loader.go` is only for internal test parsing and I don't need/want more than is currently needed. Technically we don't need floats I think but left them in.

internal/libyaml/api_test.go
```diff
@@ -0,0 +1,604 @@
...
+	t.Helper()
+
+	v := reflect.ValueOf(obj)
+	if v.Kind() == reflect.Ptr {
+		v = v.Elem()
+	}
+
+	for _, check := range checks {
+		// Handle nil checks
+		if len(check.Nil) >= 2 {
```
ccoVeille: Nitpicking here: shouldn't we report as error with a t.Fatal any field of the struct that require to have a len() > 0 but len() != 2 ?

Something like this (pseudo code)

```go
func hasLength(t *testing.T, s int, v []any) bool {

  if len(v) == 0 {
    return false
  }

  if len(v) != s {
    return true
  }

  t.Fatalf("length differs: want %d got %d")
}
```

Then this can be applied on all if statement

```suggestion
		if hasLength(t, check.Nil, 2) {
```

internal/libyaml/api_test.go
```diff
@@ -0,0 +1,604 @@
...
+		case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+			got = int(field.Int())
+			if hasExpectedInt {
+				expected = expectedInt
+			}
+		case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+			got = int(field.Uint())
+			if hasExpectedInt {
+				expected = expectedInt
+			}
```
ccoVeille, line 605: We would have expected something close to this, no?

(Copy pasted from another method of the same PR)
 
```
        case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64: 
                 if !field.CanInt() { 
                         return fmt.Errorf("field cannot store int values") 
                 } 
                 i64 := valueRefl.Int() 
                 if field.OverflowInt(i64) { 
                         return fmt.Errorf("value %v overflows %v", value, fieldType) 
                 } 
                 field.SetInt(i64) 
                 return nil 
         case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64: 
                 if !field.CanUint() { 
                         return fmt.Errorf("field cannot store uint values") 
                 } 
                 if !valueRefl.CanUint() { 
                         if valueRefl.Int() < 0 { 
                                 return fmt.Errorf("cannot convert negative %v to %v", value, fieldType) 
                         } 
                         return fmt.Errorf("cannot convert %v to %v", value, fieldType) 
                 } 
                 u64 := valueRefl.Uint() 
                 if field.OverflowUint(u64) { 
                         return fmt.Errorf("value %v overflows %v", value, fieldType) 
                 } 
                 field.SetUint(u64) 
                 return nil
```

internal/libyaml/api_test.go
ccoVeille: Of simply checking

```go
if len(check.Nil) >= 1 {
   if len(check.Nil) != 2 {
      t.Fatalf("Nil check expects exactly 2 args, got %d", len(check.Nil))
   }
   
   // the current code
}
```

ingydotnet: d2003eb

internal/libyaml/api_test.go
ingydotnet: d2003eb

PR #195, opened 2025-12-08
node: add NewDecoder function that allows customizing decode behaviour in Unmarshalers

juliusmh:
This PR adds `*Node.NewDecoder` function that allows setting options on the decoder, namely `KnownFields`, in `Unmarshaler` implementations. This is achieved by augmenting the `Decoder` type with a `node *Node` field. 

Things I considered:
- Make `Decoder` an interface. This would be a breaking change.
- Add a `DecodeWithOptions`, but we already have a decoder so why not reuse it? All development going into `Decoder` will automatically be available here too.
- Replace `Decoder.parser` with some interface or `func() *Node` instead of two fields. I'm open for this. E.g.:
 
```go
type Decoder struct {
  getNode func() *Node
  knownFields bool
}
func NewDecoder() *Decoder {
  p := newParserFromReader(r)
  return &Decoder{getNode: p.parse}
}
func (n *Node) NewDecoder() *Decoder {
  return &Decoder{
    getNode:     func()*Node { return n },
    knownFields: false,
  }
}
```

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -1709,6 +1709,25 @@ func TestObsoleteUnmarshalerRetry(t *testing.T) {
 	assert.DeepEqual(t, obsoleteSliceUnmarshaler([]int{1}), su)
 }
 
+type knownFieldsUnmarshaler[T any] struct {
+	dest T
+}
+
+func (u *knownFieldsUnmarshaler[T]) UnmarshalYAML(node *yaml.Node) error {
+	d := node.NewDecoder()
+	d.KnownFields(true)
+	return d.Decode(&u.dest)
+}
+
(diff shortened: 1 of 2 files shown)


ingydotnet:
See #212

RaduBerinde:
> See #212

Maybe I missed it but I can't tell what part of #212 addresses this?

ingydotnet:
Does the (new as of a few hours ago) addition of node.Load and node.Dump in #212 solve these needs?

If so can we close this?

RaduBerinde:
Yes, thank you!

ingydotnet:
Closing this as I feel it is now covered by #212 's node.Load and node.Dump

PR #196, opened 2025-12-08
Accept int(0) as time.Duration

juliusmh:
Copy of: https://github.com/go-yaml/yaml/pull/876

Integer zero or string "0" are valid according to time.ParseDuration. See the original PR for a lengthier description.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -619,6 +619,9 @@ func (d *decoder) scalar(n *Node, out reflect.Value) bool {
 			if !isDuration && !out.OverflowInt(int64(resolved)) {
 				out.SetInt(int64(resolved))
 				return true
+			} else if isDuration && resolved == 0 {
+				out.SetInt(0)
+				return true
 			}
 		case int64:
 			if !isDuration && !out.OverflowInt(resolved) {
--- a/decode_test.go
+++ b/decode_test.go
@@ -309,6 +309,16 @@ var unmarshalTests = []struct {
 		"a: 3s",
 		map[string]time.Duration{"a": 3 * time.Second},
 	},
+	// Zero duration as a string.
+	{
+		"a: '0'",
+		map[string]time.Duration{"a": 0},
+	},
+	// Zero duration as an int.
+	{
+		"a: 0",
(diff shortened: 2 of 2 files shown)


ingydotnet:
Our basic stance on v4 is that it behaves like v3 by default but has options for almost anything...

juliusmh:
> Would this be acceptable to be enabled by an option?

You mean an option for `Decoder`? Do you have a naming suggestion, perhaps `Decoder.ZeroIntDuration` ?

ingydotnet:
> > Would this be acceptable to be enabled by an option?
> 
> You mean an option for `Decoder`? Do you have a naming suggestion, perhaps `Decoder.ZeroIntDuration` ?

No, I just meant in general. Could this not be the default?

We are working on a new options system. Hoping to roll that out in the next couple weeks.
`ZeroIntDuration` would be a fine name for it, but don't do the work yet, until the options system lands.

ingydotnet:
See #212 for the potential new options system.

ccoVeille:
For record, I made some tests last weekend. I forgot to talk about it.
 
I don't see a problem in accepting `0` as zero duration, considering `"0"` and `'0'` are already accepted as valid zero duration.

So I agree, this should be parsed as a slice of 4 zero time.Duration 

```yaml
durations: 
  - 0s
  - '0'
  - "0"
  - 0     # this is not supported
```

Also, I confirm the reason behind the fact `'0'`and `"0"` works as zero duration is the fact time.ParseDuration accepts `string("0")` as the zero duration.

I feel like it should be OK to support `0` duration as a sane default for v4.

FYI, I checked and goccy/go-yaml doesn't support `0` as a zero duration.

Also, for records, none of the following notations are currently parsed as zero duration

```yaml
durations: 
  - 0.0
  - .0
  - 0.
  - '0.0'
  - "0.0"
```

And I don't think they should. I mention them for exhaustivity purpose and as candidates for unit tests.

ingydotnet:
@ccoVeille given that, I approve this asis. 
If you do as well, I'll merge it.

PR #204, opened 2025-12-15
Bugfix: allocate 4 bytes instead of 32 bytes

inteon:
This looks like a typo in the `read` function.

The maximum `width` is 4 (see https://github.com/yaml/go-yaml/blob/1c3a4ee7046215c76a8fcdca6bd6210894186394/internal/libyaml/yamlprivate.go#L238).
To me, it looks like we incorrectly allocated 32 bytes when we read a character that can be max 4 bytes long.
Often `read` will be called just once, for 1 character and will unnecessarily allocate space for at least 8 characters.

### DIFF
--- a/internal/libyaml/scanner.go
+++ b/internal/libyaml/scanner.go
@@ -548,7 +548,7 @@ func (parser *Parser) read(s []byte) []byte {
 		panic("invalid character sequence")
 	}
 	if len(s) == 0 {
-		s = make([]byte, 0, 32)
+		s = make([]byte, 0, 4)
 	}
 	if w == 1 && len(s)+w <= cap(s) {
 		s = s[:len(s)+1]


ingydotnet:
@inteon Can you please describe your PRs more fully?

ingydotnet:
@inteon I asked Claude to explain this to me. https://gist.github.com/ingydotnet/25dfd9ea1368d606281a7afa9d8934ff

It looks like this would be a bad change. 

If you agree, please close.

ingydotnet:
@inteon Please respond by 2026/01/09 or we'll close this.

inteon:
@ingydotnet I'll close this. Might look into this later & check all locations where this is called to understand how much should be allocated.

ingydotnet:
Thanks @inteon 🙏

PR #205, opened 2025-12-15
Simplify read

inteon:
depends on https://github.com/yaml/go-yaml/pull/204

Simplifies the `read` function.

### DIFF
--- a/internal/libyaml/scanner.go
+++ b/internal/libyaml/scanner.go
@@ -541,17 +541,10 @@ func (parser *Parser) read(s []byte) []byte {
 	if w == 0 {
 		panic("invalid character sequence")
 	}
-	if len(s) == 0 {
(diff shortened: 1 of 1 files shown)


ingydotnet:
@inteon ,

Did you review the blame commits for that code to see why it was done that way?
e6da37e7, 5b18502a, f4d271a8

Is simpler actually better here? If so, can you report your findings and explain why?

ingydotnet:
I'm guessing if #204 is the wrong direction, then this would be as well.

Please advise, @inteon

ingydotnet:
@inteon Please respond by 2026/01/09 or we'll close this.

inteon:
See https://github.com/yaml/go-yaml/pull/204#issuecomment-3699525789

PR #206, opened 2025-12-15
Simplify simple key tracking

inteon:
Refactors the simple key tracking, dropping the `simple_keys_by_tok` map.

### DIFF
--- a/internal/libyaml/scanner.go
+++ b/internal/libyaml/scanner.go
@@ -662,12 +662,25 @@ func (parser *Parser) fetchMoreTokens() error {
 		if parser.tokens_head < len(parser.tokens)-2 {
 			// If a potential simple key is at the head position, we need to fetch
 			// the next token to disambiguate it.
-			head_tok_idx, ok := parser.simple_keys_by_tok[parser.tokens_parsed]
-			if !ok {
+
+			var first_key int
+			found_potential_key := false
+
+			if len(parser.simple_key_stack) > 0 {
+				// Found a simple key on the stack
+				first_key = parser.simple_key_stack[0].token_number
+				found_potential_key = true
+			} else if parser.simple_key_possible {
+				// Found a 'current' simple key (which was not pushed to the stack yet)
+				first_key = parser.simple_key.token_number
+				found_potential_key = true
+			}
+
+			if !found_potential_key {
+				// We don't have any potential simple keys
 				break
-			} else if valid, err := parser.simpleKeyIsValid(&parser.simple_keys[head_tok_idx]); err != nil {
-				return err
-			} else if !valid {
+			} else if parser.tokens_parsed != first_key {
+				// We have not reached the potential simple key yet.
 				break
 			}
 		}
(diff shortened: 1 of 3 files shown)


ingydotnet:
@inteon Please explain your full intentions for creating this PR.

I get that it is a refactoring, but it doesn't seem to reduce the amount of code.

ingydotnet:
@inteon this assessment seems invaluable (to me) assuming it is correct:

https://gist.github.com/ingydotnet/664ff8cdb6c7c4b758e84228a9d7f3f2

Let me know what your take is on that. Anything you might want to change or add?

ccoVeille:
Just in case it's about reviewing only the last commit, as the other were in #152 and got merged.

The commit I talk about
https://github.com/yaml/go-yaml/commit/e6f0a0e4491bf33e192ac0ca99116a8aff3ae7fa

I don't think there is a need to rebase and fix the conflicts until we agree on the fact we need this commit 

I will now review the changes

internal/libyaml/yaml.go
```diff
@@ -630,8 +629,8 @@ type Parser struct {
 	indents []int // The indentation levels stack.
 
 	simple_key_allowed bool        // May a simple key occur at the current position?
-	simple_keys        []SimpleKey // The stack of simple keys.
-	simple_keys_by_tok map[int]int // possible simple_key indexes indexed by token_number
+	simple_key         SimpleKey   // The last simple key.
```
ccoVeille: I don't think we need this. Or at least, I would prefer we don't add it

internal/libyaml/scanner.go
```diff
@@ -675,30 +675,45 @@ func (parser *Parser) fetchMoreTokens() bool {
...
-			if !ok {
+
+			var first_key int
+			found_potential_key := false
+
+			if len(parser.simple_key_stack) > 0 {
+				// Found a simple key on the stack
+				first_key = parser.simple_key_stack[0].token_number
+				found_potential_key = true
+			} else if parser.simple_key.possible {
```
ccoVeille: I feel lile the simpleKey is only there to hold the current simpleKey

Can't we always push the current key to the stack and check is OK to be there in the stack ? 

Isn't what `possible` field is about ?

I would prefer if we could avoid to have a simplekey stored at the parser level

If not suitable, I feel like the onky things we check from parser.simplekey field are possible and required

Can't we simply have simpleKeyPossible, simpleKeyRequired bool at the parser level ?

internal/libyaml/scanner.go
```diff
@@ -976,19 +958,16 @@ const max_flow_level = 10000
...
 func (parser *Parser) increaseFlowLevel() error {
-	// Reset the simple key on the next level.
-	parser.simple_keys = append(parser.simple_keys, SimpleKey{
-		possible:     false,
-		required:     false,
-		token_number: parser.tokens_parsed + (len(parser.tokens) - parser.tokens_head),
-		mark:         parser.mark,
-	})
+	if parser.simple_key.possible {
+		parser.simple_key_stack = append(parser.simple_key_stack, parser.simple_key) // push current key
```
ingydotnet: I'd prefer to not see eol comments that make lines this long.

ingydotnet (approved):
This looks like a good refactoring to me.

internal/libyaml/scanner.go
ccoVeille: @inteon what do you think about this ?

cc @ingydotnet

inteon: There are definitely different ways that this can be implemented.
I will try a few different alternatives when I have some spare time.

ccoVeille: Thanks

inteon: I moved `simple_key_possible bool` to the parser level.

internal/libyaml/yaml.go
inteon: This approach is very similar to how we handle indents: we have `indent  int` and `indents []int`.

ccoVeille (approved):

PR #207, opened 2025-12-15
Pass event as argument to emit

inteon:
TLDR; Instead of setting the event field on `encoder`, we can just pass the event to `emit()`.

The PR removes a field from the struct that was there apparently just to be set then retrieved by the emit method. Passing it to emit directly seems a better way to do this. There is no reason store the event for ever in the struct. The solution in this PR allows the garbage collector to do its job, and it could (slightly) reduce the memory footprint.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -41,7 +41,6 @@ var (
 
 type encoder struct {
 	emitter               libyaml.Emitter
-	event                 libyaml.Event
 	out                   []byte
 	flow                  bool
 	indent                int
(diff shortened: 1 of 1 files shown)


ingydotnet:
@inteon Please explain your full intentions for creating this PR.

ccoVeille:
While Tim may explain it, I would say I like the PR as it removes a field from the struct that was there apparently just to be set then retrieved `emit` method. I agree that passing it to emit sounds be the way to do. There is no reason store the event for ever in the struct. The solution in this PR allows the garbage collector to do its job, and it would reduce the memory footprint

ingydotnet:
@ccoVeille thanks.
@inteon can you fix the commit msg CI error?

inteon:
@ingydotnet done

PR #212, opened 2025-12-19
Add Dump/Load, Options API

ingydotnet:
**This PR is the heart and soul of our upcoming v4 release.**

> Note: You might want to read these user guides before digging into the PR specifics:
> https://github.com/yaml/go-yaml/tree/load-dump-options-plugins/doc

It combines 4 topics that we need to see working in concert, in order to get things right for v4:

* A new Dump/Load top level API.
* A very flexible options system.
  * This started with, but replaces #162
* Data driven tests (See #210) for writing lots of tests for this to work.

I'd like to hear lots of feedback but at this point keep it at the macro level as much as possible.

To be sure, we are still committed to keeping the v3 API be the default for smooth transitions.
We've even made it trivial to use the new Dump/Load API starting with v3 (or v2) defaults!

### DIFF
--- a/options.go
+++ b/options.go
@@ -64,11 +64,12 @@ func WithIndent(indent int) Option {
 // If compact is false, '- ' is not treated as part of the indentation.
 // When called without arguments, defaults to true.
 func WithCompactSeqIndent(compact ...bool) Option {
+	if len(compact) > 1 {
+		panic("yaml: WithCompactSeqIndent accepts at most one argument")
+	}
+	val := len(compact) == 0 || compact[0]
 	return func(o *options) error {
-		if len(compact) > 1 {
-			return errors.New("yaml: WithCompactSeqIndent accepts at most one argument")
-		}
-		o.compactSeqIndent = len(compact) == 0 || compact[0]
+		o.compactSeqIndent = val
 		return nil
 	}
 }
@@ -79,11 +80,12 @@ func WithCompactSeqIndent(compact ...bool) Option {
 // that do not correspond to any fields in the target struct.
 // When called without arguments, defaults to true.
 func WithKnownFields(knownFields ...bool) Option {
+	if len(knownFields) > 1 {
+		panic("yaml: WithKnownFields accepts at most one argument")
+	}
+	val := len(knownFields) == 0 || knownFields[0]
 	return func(o *options) error {
-		if len(knownFields) > 1 {
-			return errors.New("yaml: WithKnownFields accepts at most one argument")
-		}
-		o.knownFields = len(knownFields) == 0 || knownFields[0]
+		o.knownFields = val
 		return nil
 	}
 }
@@ -97,11 +99,12 @@ func WithKnownFields(knownFields ...bool) Option {
 // This is useful when you expect exactly one document and want behavior
 // similar to [Unmarshal].
 func WithSingleDocument(singleDocument ...bool) Option {
+	if len(singleDocument) > 1 {
+		panic("yaml: WithSingleDocument accepts at most one argument")
+	}
+	val := len(singleDocument) == 0 || singleDocument[0]
 	return func(o *options) error {
-		if len(singleDocument) > 1 {
-			return errors.New("yaml: WithSingleDocument accepts at most one argument")
-		}
-		o.singleDocument = len(singleDocument) == 0 || singleDocument[0]
+		o.singleDocument = val
 		return nil
 	}
 }
@@ -130,11 +133,12 @@ func WithLineWidth(width int) Option {
 //
 // The default is true.
 func WithUnicode(unicode ...bool) Option {
+	if len(unicode) > 1 {
+		panic("yaml: WithUnicode accepts at most one argument")
+	}
+	val := len(unicode) == 0 || unicode[0]
 	return func(o *options) error {
-		if len(unicode) > 1 {
-			return errors.New("yaml: WithUnicode accepts at most one argument")
-		}
-		o.unicode = len(unicode) == 0 || unicode[0]
+		o.unicode = val
 		return nil
 	}
 }
@@ -148,11 +152,12 @@ func WithUnicode(unicode ...bool) Option {
 //
 // The default is true.
 func WithUniqueKeys(uniqueKeys ...bool) Option {
+	if len(uniqueKeys) > 1 {
+		panic("yaml: WithUniqueKeys accepts at most one argument")
+	}
+	val := len(uniqueKeys) == 0 || uniqueKeys[0]
 	return func(o *options) error {
-		if len(uniqueKeys) > 1 {
-			return errors.New("yaml: WithUniqueKeys accepts at most one argument")
-		}
-		o.uniqueKeys = len(uniqueKeys) == 0 || uniqueKeys[0]
+		o.uniqueKeys = val
 		return nil
 	}
 }
@@ -166,11 +171,12 @@ func WithUniqueKeys(uniqueKeys ...bool) Option {
 //
 // The default is false.
 func WithCanonical(canonical ...bool) Option {
+	if len(canonical) > 1 {
+		panic("yaml: WithCanonical accepts at most one argument")
+	}
+	val := len(canonical) == 0 || canonical[0]
 	return func(o *options) error {
-		if len(canonical) > 1 {
-			return errors.New("yaml: WithCanonical accepts at most one argument")
-		}
-		o.canonical = len(canonical) == 0 || canonical[0]
+		o.canonical = val
 		return nil
 	}
 }
@@ -196,11 +202,12 @@ func WithLineBreak(lineBreak LineBreak) Option {
 // When false (default), the marker is omitted for the first document.
 // When called without arguments, defaults to true.
 func WithExplicitStart(explicit ...bool) Option {
+	if len(explicit) > 1 {
+		panic("yaml: WithExplicitStart accepts at most one argument")
+	}
+	val := len(explicit) == 0 || explicit[0]
 	return func(o *options) error {
-		if len(explicit) > 1 {
-			return errors.New("yaml: WithExplicitStart accepts at most one argument")
-		}
-		o.explicitStart = len(explicit) == 0 || explicit[0]
+		o.explicitStart = val
 		return nil
 	}
 }
@@ -211,11 +218,12 @@ func WithExplicitStart(explicit ...bool) Option {
 // When false (default), the marker is omitted.
 // When called without arguments, defaults to true.
 func WithExplicitEnd(explicit ...bool) Option {
+	if len(explicit) > 1 {
+		panic("yaml: WithExplicitEnd accepts at most one argument")
+	}
+	val := len(explicit) == 0 || explicit[0]
 	return func(o *options) error {
-		if len(explicit) > 1 {
-			return errors.New("yaml: WithExplicitEnd accepts at most one argument")
-		}
-		o.explicitEnd = len(explicit) == 0 || explicit[0]
+		o.explicitEnd = val
 		return nil
 	}
 }
@@ -229,11 +237,12 @@ func WithExplicitEnd(explicit ...bool) Option {
 //
 // When false (default), all collections use block style.
 func WithFlowSimpleCollections(flow ...bool) Option {
+	if len(flow) > 1 {
+		panic("yaml: WithFlowSimpleCollections accepts at most one argument")
+	}
+	val := len(flow) == 0 || flow[0]
 	return func(o *options) error {
-		if len(flow) > 1 {
-			return errors.New("yaml: WithFlowSimpleCollections accepts at most one argument")
-		}
-		o.flowSimpleCollections = len(flow) == 0 || flow[0]
+		o.flowSimpleCollections = val
 		return nil
 	}
 }


yaml.go
```diff
@@ -87,7 +88,74 @@ type Marshaler interface {
...
+
+	// Process through load plugins
+	for _, p := range l.opts.plugins {
+		if lp, ok := p.(LoadPlugin); ok {
+			node, err = lp.ProcessLoadNode(node)
+			if err != nil {
+				return err
+			}
+		}
+	}
```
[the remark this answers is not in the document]
ingydotnet: f0f893e

encode.go
```diff
@@ -104,14 +126,67 @@ func (e *encoder) marshalDoc(tag string, in reflect.Value) {
 	if node != nil && node.Kind == DocumentNode {
 		e.nodev(in)
 	} else {
-		e.event = libyaml.NewDocumentStartEvent(nil, nil, true)
+		// Use !explicitStart for implicit flag (true = implicit/no marker)
+		e.event = libyaml.NewDocumentStartEvent(nil, nil, !e.explicitStart)
 		e.emit()
 		e.marshal(tag, in)
-		e.event = libyaml.NewDocumentEndEvent(true)
+		// Use !explicitEnd for implicit flag
```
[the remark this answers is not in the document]
ingydotnet: f0f893e

yaml.go
```diff
@@ -217,14 +289,153 @@ func unmarshal(in []byte, out any, strict bool) (err error) {
...
+		// Process through dump plugins
+		for _, p := range d.opts.plugins {
+			if dp, ok := p.(DumpPlugin); ok {
+				processedNode, err := dp.ProcessDumpNode(&node)
+				if err != nil {
+					return err
+				}
+				node = *processedNode
+			}
+		}
```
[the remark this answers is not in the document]
ingydotnet: f0f893e

ingydotnet:
**Note:** I put this in Draft Mode because it will likely take a quite a while to get things right.

But feel free to start making high level reviews of things.

Questions appreciated!

felixfontein:
I've tried `go run .` in `example/comment_plugins_comparison`, and was a bit surprised that there were no comments in the output of the v3 and v4 comment plugins. I guess this is caused by `node.Encode(v)` dumping the node as YAML and re-parsing it as a node, but I haven't dug deeper into it. Also right now the v3 and v4 comment plugins are identical - I guess this is intentional *for now* and will change eventually?

ingydotnet:
> I've tried `go run .` in `example/comment_plugins_comparison`, and was a bit surprised that there were no comments in the output of the v3 and v4 comment plugins. I guess this is caused by `node.Encode(v)` dumping the node as YAML and re-parsing it as a node, but I haven't dug deeper into it. 

I'll take a look. Thanks for trying things out. I didn't write the examples by hand and I haven't yet looked at them closely.

> Also right now the v3 and v4 comment plugins are identical - I guess this is intentional _for now_ and will change eventually?

Yep. I just picked Comments for an example plugin, but I haven't yet put a lot of effort into plugins. I was mostly concentrating on making the absence extremely clean and flexible. Now I can dig into the plugins more.

The only thing I'm fairly sure about with plugins is that you import them and then construct a new plugin object with a specific configuration. That's what you pass to `yaml.WithPlugins(comments, schema, format)` etc. Those would be 3 configured plugin objects. 

Also you can configure plugins (and options) with Go code or YAML (or JSON, since JSON is YAML).

options.go
```diff
@@ -0,0 +1,453 @@
...
+)
+
+// LineBreak represents the line ending style for YAML output.
+type LineBreak = libyaml.LineBreak
+
+// Line break constants for different platforms.
+const (
+	LNBreak   = libyaml.LN_BREAK   // Unix-style \n (default)
+	CRBreak   = libyaml.CR_BREAK   // Old Mac-style \r
+	CRLNBreak = libyaml.CRLN_BREAK // Windows-style \r\n
```
ccoVeille: Usually constants from same type are sharing a name prefix, here they are sharing a suffix.

Having a common prefix helps to find them regrouped in the documentation and IDE completion

Here a common prefix could be LineBreak

options.go
```diff
@@ -0,0 +1,453 @@
...
+	}
+}
+
+// WithCompactSeqIndent configures whether the sequence indicator '- ' is
+// considered part of the indentation when dumping YAML content.
+//
+// If compact is true, '- ' is treated as part of the indentation.
+// If compact is false, '- ' is not treated as part of the indentation.
+// When called without arguments, defaults to true.
+func WithCompactSeqIndent(compact ...bool) Option {
```
ccoVeille, line 66: This is use of variadic for one arg only, sounds strange to me in Go.

But it seems to be used 

https://github.com/search?q=language%3Ago+%22...bool%22+-is%3Afork+-is%3Aarchived&type=code&p=2

Bringing some Go expert and friends here:

@alexandear @brackendawson @chavacava what do you think about this ?

options.go
```diff
@@ -0,0 +1,453 @@
...
+		ExplicitStart         *bool   `yaml:"explicit-start"`
+		ExplicitEnd           *bool   `yaml:"explicit-end"`
+		FlowSimpleCollections *bool   `yaml:"flow-simple-coll"`
+		KnownFields           *bool   `yaml:"known-fields"`
+		SingleDocument        *bool   `yaml:"single-document"`
+		UniqueKeys            *bool   `yaml:"unique-keys"`
+	}
+	if err := Unmarshal([]byte(yamlStr), &cfg); err != nil {
+		return nil, err
+	}
```
ccoVeille, line 313: In such case I would use a map[string]string then loop on field names, this way I could report fields that are not supported.

This can be very useful to spot typos.

But the struct is pretty convenient, so maybe both can be done.

I don't think we already have an option yo report fields that are not mapped, right ?

options.go
```diff
@@ -0,0 +1,453 @@
...
+	o := &options{
+		// v4 defaults
+		indent:           2,
+		compactSeqIndent: true,
+		lineWidth:        80,
+		unicode:          true,
+		uniqueKeys:       true,
+		canonical:        false,
+		lineBreak:        libyaml.LN_BREAK,
+	}
```
ccoVeille, line 435: Maybe this, as the canonical and line break are not specific to v4

```suggestion
	o := &options{
		canonical:        false,
		lineBreak:        libyaml.LN_BREAK,

		// v4 defaults
		indent:           2,
		compactSeqIndent: true,
		lineWidth:        80,
		unicode:          true,
		uniqueKeys:       true,
	}
```

options.go
```diff
@@ -0,0 +1,453 @@
...
+// applyOptions applies the given options to a new options struct.
+// Starts with v4 defaults.
+func applyOptions(opts ...Option) (*options, error) {
+	o := &options{
+		// v4 defaults
+		indent:           2,
+		compactSeqIndent: true,
+		lineWidth:        80,
+		unicode:          true,
+		uniqueKeys:       true,
```
ccoVeille, line 434: Another way to consider the v4 default would be to prepend V4 to the opts list.

This way the default are really the v4 ones

options.go
```diff
@@ -0,0 +1,453 @@
...
+// Usage:
+//
+//	yaml.Dump(&data, yaml.V3)
+//	yaml.Dump(&data, yaml.V3, yaml.WithIndent(2))
+var V3 = Options(
+	WithIndent(4),
+	WithLineWidth(80),
+	WithUnicode(true),
+	WithUniqueKeys(true),
+)
```
ccoVeille, line 402: There is a bug according to me here.

I would add CompactSeqIdent false as someone may have V4 and append V3 options

BTW, same remark for v2

See also https://github.com/yaml/go-yaml/pull/212#discussion_r2636612303

options.go
```diff
@@ -0,0 +1,453 @@
...
+
+// applyOptions applies the given options to a new options struct.
+// Starts with v4 defaults.
+func applyOptions(opts ...Option) (*options, error) {
+	o := &options{
+		// v4 defaults
+		indent:           2,
+		compactSeqIndent: true,
```
ccoVeille, line 431: Please note this means v2 and v3 won't have compactSeqIndent false if used, see https://github.com/yaml/go-yaml/pull/212#discussion_r2636611677

plugin/comment/v3/comment.go
```diff
@@ -0,0 +1,43 @@
+//
+// Copyright (c) 2025 The go-yaml Project Contributors
+// SPDX-License-Identifier: Apache-2.0
+//
+
+// Package v3 provides a comment plugin with v3-specific behavior.
+package v3
```
ccoVeille: While the organization seems logic I'm a bit afraid by two things:

- the way Go behaves with vX packages, I know there is a special behavior with root level packages. Maybe it's OK, but I raise it just in case
- because of the way Go use import it means this packages would be imported with this name `v3`, so in the code someone would use `v3.Plugin` for telling comment are handled with the v3 way
- collision and confusion: we may have multiple v3 packages, one for comment,onr for map/slice/whatever

I feel like something like this would be expected

```
plugins
   nocomment/comment.go
   v3comment/comment.go
   v4comment/comment.go
   ...
   v3map/foo.go
   v3whatever/bar.go
```

Also this way v3 things are regrouped together, same for v4.

Another option would be

```
plugins
   commentremoval/comment.go
   commentv3/comment.go
   commentv5/comment.go
   ...
   mapv3/foo.go
   whateverv3/bar.go
```

util/godoc
```diff
@@ -0,0 +1,120 @@
...
+Usage: $0 <command>
+
+Commands:
+  serve   Start the documentation server
+  stop    Stop the documentation server
+  open    Open documentation in browser (starts server if needed)
+  status  Check if server is running
+
+Environment variables:
+  GODOC_BIN   Path to godoc binary (default: .cache/local/bin/godoc)
```
ccoVeille: Well, godoc is deprecated, it doesn't support features that were added to recent Go versions

`go doc` is the way to do now

https://pkg.go.dev/cmd/doc

So there is no need for an extra tool, I feel.

Go doc is here for a while now, Go 1.18 is supported.

But old Go version might have some arguments, here you will have to check by calling `go help doc` and test.

encode.go
```diff
@@ -31,29 +31,51 @@ import (
...
+	explicitEnd           bool
+	flowSimpleCollections bool
 }
 
-func newEncoder() *encoder {
+// newEncoder creates a new YAML encoder with the given options.
+//
+// The writer parameter specifies the output destination for the encoder.
+// If writer is nil, the encoder will write to an internal buffer.
+func newEncoder(writer io.Writer, opts *options) *encoder {
```
ccoVeille, line 59: This would lead to use `newEncoder(nil, nil)` in some context.

It's uneasy to understand.

That's why I used a Writer option in my implementation. So I could pass a simple nil when I don't need options or writer.

But maybe you disliked the idea to use an option for this. Then another solution is to add two variables 

var noOptions *opts
var noWriter ioWriter

This way these two are nil, and can be used when calling `newEncoder(noWriter, noOptions)`

This pattern is used in http package with http.NoBody variable.

yaml.go
```diff
@@ -118,7 +190,7 @@ func (dec *Decoder) KnownFields(enable bool) {
 // See the documentation for Unmarshal for details about the
 // conversion of YAML into a Go value.
 func (dec *Decoder) Decode(v any) (err error) {
-	d := newDecoder()
+	d := newDecoder(nil)
```
ccoVeille, line 202: See https://github.com/yaml/go-yaml/pull/212#discussion_r2636635483

For my suggestion of being able to use `newDecoder(noOptions)`

yaml.go
```diff
@@ -217,14 +293,156 @@ func unmarshal(in []byte, out any, strict bool) (err error) {
 //	yaml.Marshal(&T{F: 1}} // Returns "a: 1\nb: 0\n"
 func Marshal(in any) (out []byte, err error) {
 	defer handleErr(&err)
-	e := newEncoder()
+	e := newEncoder(nil, nil)
```
ccoVeille: This nil, nil is cryptic here when reading only this line

See  https://github.com/yaml/go-yaml/pull/212#discussion_r2636635483 about the suggestions I'm doing to resolve it

options.go
ingydotnet: That makes sense. These were obviously inherited from libyaml's, But I am fine with adding a common prefix.

options.go
ingydotnet: Personally, this is way too verbose for me and I'm trying to make the new API really enjoyable.

`yaml.WithCompactSeqIndent()` is long. `WithCompactSeqIndent(true)` is too long and overstating the already obvious. 

That said, it's not my hill to die on...

yaml.go
```diff
@@ -141,7 +213,7 @@ func (dec *Decoder) Decode(v any) (err error) {
 // See the documentation for Unmarshal for details about the
 // conversion of YAML into a Go value.
 func (n *Node) Decode(v any) (err error) {
```
RaduBerinde, line 226: #162 also provided a way to pass options to `Node.Decode`, which addresses the problem described here: https://github.com/go-yaml/yaml/issues/460

Maybe we can add a `Load()` method that takes options?

plugin/comment/v3/comment.go
felixfontein: Maybe also `comment_removal`, `comment_v3`, `comment_v4`, ... so it's easier to parse the names while reading.

ccoVeille: Yes, but such package names are discouraged by the Go convention/specs.

"By convention, packages are given lower case, single-word names; there should be no need for underscores or mixedCaps."
Source: [Effective Go: Package Names](https://go.dev/doc/effective_go#package-names)
​
​"Good package names are short and clear. They are lower case, with no under_scores or mixedCaps."
Source: [The Go Blog: Package Names](https://go.dev/blog/package-names?hl=fr-FR)

That's why I suggested that

Another solution, in another direction could be to have the following packages organization

```
pluglin/
  nocomment
  v3default/
     comment
     whatever
   v4default/
     comment
     whatever
```

This way people can use `comment` package from wherever they want, and we can assume quite no one would use `comment` with v3 and v4 in the sane file, and if they do they will be able to use import aliases.

doc/dump-load-api.md
```diff
@@ -0,0 +1,554 @@
+# YAML Encoding and Decoding API
+
+This guide covers go-yaml's encoding (dumping) and decoding (loading) APIs,
+from simple one-liners to advanced streaming with full configuration control.
+
+## Recommended API for v4+
+
+**For new code in v4+, prefer `Dump`/`Load` over `Marshal`/`Unmarshal` and
+`Encode`/`Decode`.**
```
colinjlacy, line 9: Would we consider deprecating `Marshal`/`Unmarshal` and `Encode`/`Decode` in v4, with the intent of removing them in v5?

doc/dump-load-api.md
```diff
@@ -0,0 +1,554 @@
...
+`Marshal`/`Unmarshal` and `Encode`/`Decode` are retained in v4 to smooth the
+transition from v3:
+
+- They work **without options** and continue using **v3 semantics** (4-space
+  indent, non-compact sequences)
+- They're perfect for upgrading existing v3 code with minimal changes
+- For new code, use `Dump`/`Load` instead—they can replicate the old behavior
+  and much more
+
+### Flexibility
```
colinjlacy, line 50: This is why I would say it would be advantageous to deprecate now. There's a clear and concise migration path, and once it's set we can guide users to standardize on this.

util/godoc
ingydotnet: Claude says `go doc` is only for terminal. godoc was to see docs in browser. They said that pkgsite is the modern replacement for godoc. I'm adding `make doc` for terminal and switching browser to pkgsite.

ingydotnet: actually `go doc -http -all` does basically same thing so I'll simplify to that.

ingydotnet: fixed in f62ca2e

encode.go
ingydotnet: 0c77057

yaml.go
ingydotnet: 0c77057

yaml.go
ingydotnet: 0c77057

plugin/comment/v3/comment.go
ingydotnet: go-yaml will support a number of plugin "types". These are named APIs for hooking into various parts of the code. For each type we will offer at least one plugin package/implementation. Anyone else can publish alternative  packages that implement the API for that plugin type.

Some plugin types might be:
- `comment`
- `schema`
- `resolver`
- `constructor`
- `format`
- `function`
- `envvars`

I imagine our plugin packages would be organized like:
```
plugin/schema/go-yaml/plugin.go
plugin/resolver/go-yaml/plugin.go
plugin/comment/strategy1/plugin.go
plugin/comment/strategy2/plugin.go
```

All plugins can be used with v2, v3, v4, v5, etc defaults. In general, there is nothing version specific about plugins.

We might have a `plugin/comment/v3/...` plugin that does comments the way v3 does now, but we could also call that `plugin/comment/niemeyer/...`.

I don't want to bikeshed these names any more until we have a usable set of plugins from at least 3 types.

options.go
ingydotnet: I'm not 100% sure what you mean here but I don't really want to tie options to a specific version.
I just want to make it easy for the Dump/Load API to easily transition from v2 or Marshal API.

encode.go
```diff
@@ -104,14 +133,67 @@ func (e *encoder) marshalDoc(tag string, in reflect.Value) {
 	if node != nil && node.Kind == DocumentNode {
 		e.nodev(in)
 	} else {
-		e.event = libyaml.NewDocumentStartEvent(nil, nil, true)
+		// Use !explicitStart for implicit flag (true = implicit/no marker)
+		e.event = libyaml.NewDocumentStartEvent(nil, nil, !e.explicitStart)
```
ccoVeille: Maybe we could add libyaml.NoVersionDirective and libyaml.NoTagDirective here that would be nil the same way you added for noWriter

yaml.go
```diff
@@ -87,7 +88,78 @@ type Marshaler interface {
 // See the documentation of Marshal for the format of tags and a list of
```
ccoVeille, line 88: A godoc link could be added here with this syntax, you will see the difference by launching go doc -http

 ```suggestion
 // See the documentation of [Dump] for the format of tags and a list of
```

File: options.go (hunk 4 above)
Comment (ingydotnet) [reply options.go]: ad3b9dc

File: yaml.go [hunk 21]
```diff
@@ -87,7 +88,78 @@ type Marshaler interface {
...
 // supported tag options.
 func Unmarshal(in []byte, out any) (err error) {
```
ccoVeille, line 92: I would add something like this in the go doc of Unmarshal

```suggestion
//
// Unmarshal was the only available method to Decode YAML before the v4.
// For compatibility purpose the behavior of the v3 was added for this method in the v4,
// but you should consider using [Load] for now, as it comes with better settings, and can be configured via [Option]
func Unmarshal(in []byte, out any) (err error) {
```

Then I'm unsure whether v4 could add the Go special ``// Deprecated:` tag in the go doc, to make it clearly appears as deprecated, but still available.

yaml.go
```diff
@@ -87,7 +88,78 @@ type Marshaler interface {
...
+		decoder: newDecoder(o),
+		opts:    o,
+	}, nil
+}
+
+// Load reads the next YAML-encoded document from its input and stores it
+// in the value pointed to by v.
+//
+// Returns io.EOF when there are no more documents to read.
+// If WithSingleDocument option was set and a document was already read,
```
ccoVeille, line 125:
```suggestion
// Returns [io.EOF] when there are no more documents to read.
// If [WithSingleDocument] option was set and a document was already read,
```

options.go
ingydotnet: 134ec7b

yaml.go
```diff
@@ -217,14 +293,156 @@ func unmarshal(in []byte, out any, strict bool) (err error) {
 //	yaml.Marshal(&T{F: 1}} // Returns "a: 1\nb: 0\n"
 func Marshal(in any) (out []byte, err error) {
```
ccoVeille, line 339: Same remarks here about:
- mentioning [Dump] and Marshal behavior was kept for compatibility purpose
- using `//Deprecated: ` tag

yaml.go
```diff
@@ -141,7 +213,7 @@ func (dec *Decoder) Decode(v any) (err error) {
 // See the documentation for Unmarshal for details about the
 // conversion of YAML into a Go value.
 func (n *Node) Decode(v any) (err error) {
-	d := newDecoder()
+	d := newDecoder(nil)
```
ccoVeille, line 227: You can use noOptions here

yaml.go
ccoVeille, line 202: Use 


```suggestion
	d := newDecoder(noOptions)
```

yaml.go
```diff
@@ -87,7 +88,78 @@ type Marshaler interface {
...
+
+// Load reads the next YAML-encoded document from its input and stores it
+// in the value pointed to by v.
+//
+// Returns io.EOF when there are no more documents to read.
+// If WithSingleDocument option was set and a document was already read,
+// subsequent calls return io.EOF.
+//
+// See the documentation for [Unmarshal] for details about the
+// conversion of YAML into a Go value.
```
ccoVeille: Here also I would suggest this

```suggestion
// See the documentation for [Load] for details about the
// conversion of YAML into a Go value.
```

See https://github.com/yaml/go-yaml/pull/212#discussion_r2637333299

yaml.go
```diff
@@ -217,14 +293,156 @@ func unmarshal(in []byte, out any, strict bool) (err error) {
...
+	if err := d.Close(); err != nil {
+		return nil, err
+	}
+	return buf.Bytes(), nil
+}
+
+// DumpAll encodes multiple values as a multi-document YAML stream.
+//
+// Each value becomes a separate YAML document, separated by "---".
+// See [Marshal] for details about the conversion of Go values to YAML.
```
ccoVeille, line 371: See https://github.com/yaml/go-yaml/pull/212#discussion_r2637333299

```suggestion
// DumpAll encodes multiple values as a multi-document YAML stream.
//
// Each value becomes a separate YAML document, separated by "---".
// See [Dump] for details about the conversion of Go values to YAML.
```

yaml.go
```diff
@@ -217,14 +293,156 @@ func unmarshal(in []byte, out any, strict bool) (err error) {
...
 	defer e.destroy()
 	e.marshalDoc("", reflect.ValueOf(in))
 	e.finish()
 	out = e.out
 	return out, err
 }
 
+// Dump encodes a value to YAML with the given options.
+//
+// See [Marshal] for details about the conversion of Go values to YAML.
```
ccoVeille, line 351: I would do the opposite here, and add the documentation in Dump, and mention in Marshal and DumpAll to look at Dump


```suggestion
// Dump encodes a value to YAML with the given [Option]s.
// The structure of the generated document will reflect the structure of the value itself.
// Maps and pointers (to struct, string, int, etc) are accepted as the in value.
//
// Struct fields are only marshaled if they are exported (have an upper case
// first letter), and are marshaled using the field name lowercased as the
// default key. Custom keys may be defined via the "yaml" name in the field
// tag: the content preceding the first comma is used as the key, and the
// following comma-separated options are used to tweak the marshaling process.
// Conflicting names result in a runtime error.
//
// The field tag format accepted is:
//
//        `(...) yaml:"[<key>][,<flag1>[,<flag2>]]" (...)`
//
// The following flags are currently supported:
//
//        omitempty    Only include the field if it's not set to the zero
//                     value for the type or to empty slices or maps.
//                     Zero valued structs will be omitted if all their public
//                     fields are zero, unless they implement an IsZero
//                     method (see the [IsZeroer] interface type), in which
//                     case the field will be excluded if IsZero returns true.
//
//        flow         Dump using a flow style (useful for structs,
//                     sequences and maps).
//
//        inline       Inline the field, which must be a struct or a map,
//                     causing all of its fields or keys to be processed as if
//                     they were part of the outer struct. For maps, keys must
//                     not conflict with the yaml keys of other struct fields.
//
// In addition, if the key is "-", the field is ignored.
//
// For example:
//
//        type T struct {
//            F int `yaml:"a,omitempty"`
//            B int
//        }
//        yaml.Dump(&T{B: 2}) // Returns "b: 2\n"
//        yaml.Dump(&T{F: 1}} // Returns "a: 1\nb: 0\n"
//
```

options.go
ingydotnet: I don't understand. Are any of those specific to v4?
I think the comment is just pointing out that thos are v4 defaults, and thus the default options for Dump and Load.

yaml.go
```diff
@@ -217,14 +293,156 @@ func unmarshal(in []byte, out any, strict bool) (err error) {
...
+func Load(in []byte, out any, opts ...Option) error {
+	return unmarshal(in, out, opts...)
+}
+
+// LoadAll decodes all YAML documents from the input.
+//
+// Returns a slice containing all decoded documents. Each document is
+// decoded into an any value (typically map[string]any or []any).
+//
+// See [Unmarshal] for details about the conversion of YAML into Go values.
```
ccoVeille, line 429: See https://github.com/yaml/go-yaml/pull/212#discussion_r2637339722

```suggestion
// See [Load] for details about the conversion of YAML into Go values.
```

yaml.go
```diff
@@ -217,14 +293,156 @@ func unmarshal(in []byte, out any, strict bool) (err error) {
...
+	}
+	if err := d.Close(); err != nil {
+		return nil, err
+	}
+	return buf.Bytes(), nil
+}
+
+// Load decodes the first YAML document with the given options.
+//
+// See [Unmarshal] for details about the conversion of YAML into Go values.
```
ccoVeille: Same as https://github.com/yaml/go-yaml/pull/212#discussion_r2637333299

```suggestion
// Load decodes the first YAML document with the given [Option]s.
//
// Maps and pointers (to a struct, string, int, etc) are accepted as out
// values. If an internal pointer within a struct is not initialized,
// the yaml package will initialize it if necessary for unmarshalling
// the provided data. The out parameter must not be nil.
//
// The type of the decoded values should be compatible with the respective
// values in out. If one or more values cannot be decoded due to a type
// mismatches, decoding continues partially until the end of the YAML
// content, and a [*yaml.TypeError] is returned with details for all
// missed values.
//
// Struct fields are only unmarshalled if they are exported (have an
// upper case first letter), and are unmarshalled using the field name
// lowercased as the default key. Custom keys may be defined via the
// "yaml" name in the field tag: the content preceding the first comma
// is used as the key, and the following comma-separated options are
// used to tweak the marshaling process (see [Dump]).
// Conflicting names result in a runtime error.
//
// For example:
//
//        type T struct {
//            F int `yaml:"a,omitempty"`
//            B int
//        }
//        var t T
//        yaml.Load([]byte("a: 1\nb: 2"), &t)
//
// See the documentation of [Dump] for the format of tags and a list of
// supported tag options.
```

yaml.go
ingydotnet: and also a `node.Dump()` with options?

yaml.go
```diff
@@ -217,14 +293,156 @@ func unmarshal(in []byte, out any, strict bool) (err error) {
...
+		opts:    o,
+	}, nil
+}
+
+// Dump writes the YAML encoding of v to the stream.
+//
+// If multiple values are dumped to the stream, the second and subsequent
+// documents will be preceded with a "---" document separator.
+//
+// See the documentation for [Marshal] for details about the conversion of Go
```
ccoVeille, line 475: See https://github.com/yaml/go-yaml/pull/212#discussion_r2637339722

```suggestion
// See the documentation for [Dump] for details about the conversion of Go
```

options.go
```diff
@@ -0,0 +1,455 @@
...
+}
+
+// WithLineBreak sets the line ending style for YAML output.
+//
+// Available options:
+//   - LineBreakLN: Unix-style \n (default)
+//   - LineBreakCR: Old Mac-style \r
+//   - LineBreakCRLN: Windows-style \r\n
+//
+// The default is LineBreakLN.
```
ccoVeille, line 191:
```suggestion
// WithLineBreak sets the line ending style for YAML output.
//
// Available options:
//   - [LineBreakLN]: Unix-style \n (default)
//   - [LineBreakCR]: Old Mac-style \r
//   - [LineBreakCRLN]: Windows-style \r\n
//
// The default is [LineBreakLN].
```

options.go
```diff
@@ -0,0 +1,455 @@
...
+		return nil
+	}
+}
+
+// CombineOptions combines multiple options into a single Option.
+//
+// Deprecated: Use Options instead.
+func CombineOptions(opts ...Option) Option {
+	return Options(opts...)
+}
```
ccoVeille: Is there a need for this ?

options.go
```diff
@@ -0,0 +1,455 @@
...
+			optList = append(optList, WithLineBreak(LineBreakCRLN))
+		default:
+			return nil, errors.New("yaml: invalid line-break value (use ln, cr, or crln)")
+		}
+	}
+
+	return Options(optList...), nil
+}
+
+// WithPlugins adds multiple plugins at once.
```
ccoVeille:
```suggestion
// WithPlugins adds multiple [Plugin]s at once.
```

options.go
ccoVeille: OK then

options.go
ccoVeille: My point was about the fact the v4 setting doesn't set canonical or line break, and these are not part of the v4 settings, also that v2 and v3 also use these same two values for these two settings

yaml.go
```diff
@@ -87,7 +88,78 @@ type Marshaler interface {
...
+	parser   *parser
+	decoder  *decoder
+	opts     *options
+	docCount int
+}
+
+// NewLoader returns a new Loader that reads from r with the given options.
+//
+// The Loader introduces its own buffering and may read data from r
+// beyond the YAML values requested.
```
ccoVeille:
```suggestion
// NewLoader returns a new [Loader] that reads from r with the given options.
//
// The [Loader] introduces its own buffering and may read data from r
// beyond the YAML values requested.
```

yaml.go
```diff
@@ -87,7 +88,78 @@ type Marshaler interface {
...
+}
+
+// Load reads the next YAML-encoded document from its input and stores it
+// in the value pointed to by v.
+//
+// Returns io.EOF when there are no more documents to read.
+// If WithSingleDocument option was set and a document was already read,
+// subsequent calls return io.EOF.
+//
+// See the documentation for [Unmarshal] for details about the
```
ccoVeille:
```suggestion
// See the documentation for [Load] for details about the
```

yaml.go
ccoVeille: It would make sense indeed

yaml.go
```diff
@@ -87,7 +88,78 @@ type Marshaler interface {
 // See the documentation of Marshal for the format of tags and a list of
 // supported tag options.
 func Unmarshal(in []byte, out any) (err error) {
-	return unmarshal(in, out, false)
+	return unmarshal(in, out, V3)
```
ccoVeille, line 93: Maybe this would make it clearer no?


```suggestion
	return Load(in, out, V3)
```

yaml.go
ingydotnet: @RaduBerinde, this is a good call.

Added in 1dd8861 (with new example/ programs)

Could you please review?

doc/dump-load-api.md
ingydotnet: I'm in favor of that. What do others think?

What do you imagine that deprecation process would look like?

ccoVeille: I would deprecate them all too.

Which means adding a `// Deprecated: ` flag in the godoc, and keep themgor now, and remove them in the v5

See https://github.com/yaml/go-yaml/pull/212#discussion_r2637320689

doc/dump-load-api.md
```diff
@@ -0,0 +1,635 @@
...
+implementations where you need to preserve options like `WithKnownFields()`.
+
+```go
+type Config struct {
+    Name string `yaml:"name"`
+    Port int    `yaml:"port"`
+}
+
+func (c *Config) UnmarshalYAML(node *yaml.Node) error {
+    type plain Config
```
RaduBerinde, line 483: nit: this example is very confusing (why do we need another type?)

A more realistic usage would be something like:

```go
type SomeObject struct {
  // ...
}

func (o *SomeObject) UnmarshalYAML(node *yaml.Node) error {
  var cfg struct {
    Name string `yaml:"name"`
    Port int    `yaml:"port"`
  }
  if err := node.Load(&cfg, yaml.WithKnownFields()); err != nil {
    return err
  }
  o.init(cfg.Name, cfg.Port)
  return nil
}
```

encode.go
ingydotnet: done in 7ae5826

yaml.go
ingydotnet: @ccoVeille I really want high level observations at this point. I've addressed several of your low level points with fixes, but this is not what I want to be doing yet. I'm going to mark these as resolved for the time being. We need to get the big picture headed in the right direction first.

options.go
ingydotnet: ba786e7

options.go
RaduBerinde: Could do `yaml.WithCompactSeqIndent()` and `yaml.WithNoCompactSeqIndent()`

yaml.go
RaduBerinde: Nice, that will do it!

doc/dump-load-api.md
RaduBerinde: Oh, I see why another type (otherwise it would just recurse). I guess this is a trick to force `WithKnownFields()` regardless of the decoder settings. Maybe just add a comment explaining that above `type plain Config`  (eg `define a type that does not implement UnmarshalYAML`).

doc/dump-load-api.md
ingydotnet: Done in bec97d4

options.go
ingydotnet: Going with this for now.

doc/dump-load-api.md
ingydotnet: and also c240f6b

cmd/go-yaml/event.go
```diff
@@ -160,11 +165,13 @@ func processEventsDecode(profuse, compact bool) error {
 				info := formatEventInfo(event, profuse)
 
 				var buf bytes.Buffer
-				enc := yaml.NewEncoder(&buf)
-				enc.SetIndent(2)
-				if err := enc.Encode([]*EventInfo{info}); err != nil {
+				enc, err := yaml.NewDumper(&buf, yaml.WithIndent(2))
```
ccoVeille: NewDumper uses yaml.V4 as default settings so the indent is already 2, no ?

```suggestion
				enc, err := yaml.NewDumper(&buf)
```

Please note such a line is repeated in many places over the files updated in this PR

cmd/go-yaml/main.go
```diff
@@ -182,12 +185,14 @@ func main() {
 
 				info := FormatNode(node)
 
-				// Use encoder with 2-space indentation
+				// Use dumper with 2-space indentation
```
ccoVeille: Is there a need to mention the new 2 space indentation now it's the default ?

See https://github.com/yaml/go-yaml/pull/212#discussion_r2637642333


And if it isn't worth mentioning it the comment line can be removed

```suggestion
```

cmd/go-yaml/event.go
ccoVeille: here and all along the files, we are now storing a Dumper in a variable named `enc`.

 maybe it could be renamed as it could create confusion

options.go
```diff
@@ -0,0 +1,448 @@
...
+//
+// Example:
+//
+//	yaml.Dump(&data, yaml.V4, yaml.WithPlugins(comments.V4()))
```
ccoVeille: I feel like the code in plugin is different than this comment.

There is no comments.V4() right now, right ?

plugin/comment/doc.go
```diff
@@ -0,0 +1,82 @@
...
+//	dumper, _ := yaml.NewDumper(writer, yaml.WithPlugins(none.New()))
+//	dumper.Dump(&node)
+//	dumper.Close()
+//
+// # Example: Using v4 preset with comment plugin
+//
+//	import "go.yaml.in/yaml/v4/plugin/comment/v4"
+//
+//	// v4 preset + explicit comment plugin
+//	dumper, _ := yaml.NewDumper(w, yaml.V4, yaml.WithPlugins(v4.New()))
```
ccoVeille: As we stated earlier, here

- https://github.com/yaml/go-yaml/pull/212#discussion_r2636621886
- https://github.com/yaml/go-yaml/pull/212#discussion_r2637307932

I feel like the plugin system, which is wonderful new feature, needs to be iterated a bit, in term of code and packages organization.

Having v3.New, v4.New is a problem to me.

Especially because other plugins than comment would also have these v3.New and v4.New

We can do better than that.

**I would like to suggest to move the plugin addition out of this PR.**

Alternatively, I'm also fine with merging like this if we are OK to work on a better organization for plugin after merging it. I can open PRs.

yaml.go
ccoVeille: OK, then.

I can open a PR to address all the improvements I raised in another PR

ccoVeille: OK, then.

I can open a PR to address all the improvements I raised in another PR

options.go
ccoVeille: Reopening for visibility purpose

plugin/comment/v3/comment.go
RaduBerinde:
```
plugin/schema/go-yaml/plugin.go
plugin/resolver/go-yaml/plugin.go
```
Are both of these a `goyaml` package? If I want to use both of these, I'd have to manually rename one of the imports?

I would include the "type" in the final package name, e.g.
```
plugin/schema/schemagoyaml
plugin/resolver/resolvergoyaml
plugin/comment/commentv3
plugin/comment/commentv4
```
and perhaps drop the second level organization altogether:
```
plugin/schemagoyaml
plugin/resolvergoyaml
plugin/commentv3
plugin/commentv4
```
`commentv3.New()` is a lot more readable than `v3.New()`

I would also consider putting the plugins in a separate `yaml/go-yaml-plugins` repository which can have more frequent releases and which you can update (as a dependency) separately from the main repo.

I second the separation of plugins from the rest of the PR.

ingydotnet: Those are fine suggestions. I was just explaining the plugin type concept.

plugin/comment/doc.go
ingydotnet: I started this with:

> This PR is really big, and it will get bigger before it gets smaller.

Absolutely the plugins should be in a separate PR, but the point of doing them together was to make sure the new API and options system worked with the plugins system. 

I still want to get a few plugins working here before pulling out the plugin system and merging this PR.

options.go
ingydotnet: wip

cmd/go-yaml/event.go
ingydotnet: de1b804

ingydotnet: de1b804

ingydotnet:
Well all that being said, I'm seeing good reason to get this merged asap (sans plugins) so that we can use the new api and options for other PRs right away. I'll make those changes.

ingydotnet:
OK, plugins work removed. Ready for real review and possible merge!

doc/options.md
```diff
@@ -0,0 +1,655 @@
...
+Your CI system expects 2-space indented YAML:
+
+```go
+import "go.yaml.in/yaml/v4/options/v2"
+
+func GenerateCI(config *CIConfig) ([]byte, error) {
+    var buf bytes.Buffer
+
+    dumper, err := yaml.NewDumper(&buf,
+        v2.Opts(),
```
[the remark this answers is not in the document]
ingydotnet: ae4487e

doc.go
```diff
@@ -0,0 +1,118 @@
...
+//	    yaml.WithIndent(4),
+//	)
+//
+// Load options from YAML configuration files:
+//
+//	opts, err := yaml.OptsYAML(configYAML)
+//	dumper, err := yaml.NewDumper(w, opts)
+//
+// See the [go.yaml.in/yaml/v4/options] package and its version-specific
+// subpackages (v2, v3, v4) for details.
```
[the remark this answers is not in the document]
ingydotnet: ae4487e

example/README.md
```diff
@@ -0,0 +1,151 @@
...
+// Dump
+data, _ := yaml.Marshal(&config)
+```
+
+### Streaming with Options
+
+```go
+// Load with options
+loader, _ := yaml.NewLoader(reader,
+    yaml.WithKnownFields(true),
```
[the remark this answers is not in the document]
ingydotnet: ae4487e

example/README.md
```diff
@@ -0,0 +1,151 @@
...
+dumper.Close()
+```
+
+### Using Version Presets
+
+```go
+import "go.yaml.in/yaml/v4/options/v4"
+
+dumper, _ := yaml.NewDumper(writer,
+    v4.Opts(),
```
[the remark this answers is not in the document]
ingydotnet: ae4487e

doc/options.md
```diff
@@ -0,0 +1,655 @@
...
+**Example output with different indents:**
+
+```yaml
+# WithIndent(2)
+servers:
+  web:
+    host: localhost
+    port: 8080
+
+# WithIndent(4) - default
```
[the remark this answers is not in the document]
ingydotnet: ae4487e

example/node_load_strict_unmarshaler/main.go
```diff
@@ -0,0 +1,54 @@
...
+`
+
+	var config1 Config
+	err := yaml.Load([]byte(validYAML), &config1)
+	if err != nil {
+		log.Fatal(err)
+	}
+	fmt.Printf("✓ Valid config loaded: %+v\n\n", config1)
+
+	// Invalid YAML with unknown field - should fail
```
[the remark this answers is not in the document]
ingydotnet: ae4487e

yts/known-failing-tests
```diff
@@ -210,7 +210,6 @@ TestYAMLSuite/010/UnmarshalTest
 TestYAMLSuite/010/EventComparisonTest
 TestYAMLSuite/YJV2/UnmarshalTest
 TestYAMLSuite/YJV2/EventComparisonTest
-TestYAMLSuite/229Q/JSONComparisonTest
```
ccoVeille, line 213: For records and archive purpose, here was what not passing until now

```
--- FAIL: TestYAMLSuite (0.00s)
    --- FAIL: TestYAMLSuite/229Q (0.00s)
        test_suite_test.go:72: Running test: testdata/data-2022-01-17/229Q
            Description: Spec Example 2.4. Sequence of Mappings
        --- FAIL: TestYAMLSuite/229Q/JSONComparisonTest (0.00s)
            test_suite_test.go:40: NAME::: TestYAMLSuite/229Q/JSONComparisonTest, false, false
            test_suite_test.go:229: Test: testdata/data-2022-01-17/229Q
                Description: Spec Example 2.4. Sequence of Mappings
                
                Error: YAML unmarshal vs JSON unmarshal mismatch
                Expected: [map[avg:0.278 hr:65 name:Mark McGwire] map[avg:0.288 hr:63 name:Sammy Sosa]]
                Got     : [map[avg:0.278 hr:65 name:Mark McGwire] map[avg:0.288 hr:63 name:Sammy Sosa]]
```

I'm a bit surprised this is changing, even if I'm happy a new test pass, I don't understand why

.gitignore
```diff
@@ -1,5 +1,5 @@
 /.cache/
 /.claude/
+/CLAUDE.md
```
colinjlacy, line 3: I'm fine with this being gitignored for now, but we should start to formulate a plan for adding files like this to the repo.

cmd/go-yaml/main.go
```diff
@@ -222,25 +226,27 @@ func ProcessNodeUnmarshal() error {
...
-		if err := yaml.Unmarshal(doc, &node); err != nil {
-			return fmt.Errorf("failed to unmarshal YAML to node: %w", err)
+		if err := yaml.Load(doc, &node); err != nil {
+			return fmt.Errorf("failed to load YAML to node: %w", err)
 		}
 
 		info := FormatNode(node)
 
-		// Use encoder with 2-space indentation
+		// Use dumper with 2-space indentation
```
colinjlacy: [nit]: since this isn't explicitly set here, I'd remove this comment.

decode.go
```diff
@@ -316,13 +316,19 @@ var (
...
 		stringMapType:  stringMapType,
 		generalMapType: generalMapType,
-		uniqueKeys:     true,
+		uniqueKeys:     true, // default when no options provided
 	}
 	d.aliases = make(map[*Node]bool)
+
+	if opts != nil {
+		d.knownFields = opts.knownFields
+		d.uniqueKeys = opts.uniqueKeys
```
colinjlacy, line 331: Just so we're clear - and I'm tired so I might not be reading this right - but I do believe that passing in a partial struct (e.g. `options{knownFields: ...}`) will result in `uniqueKeys` being false. Is that the intention?

encode.go
```diff
@@ -497,7 +581,7 @@ func (e *encoder) node(node *Node, tail string) {
 
 	case SequenceNode:
 		style := libyaml.BLOCK_SEQUENCE_STYLE
-		if node.Style&FlowStyle != 0 {
+		if node.Style&FlowStyle != 0 || e.isSimpleCollection(node) {
```
colinjlacy, line 587: I'm curious about why these cases were added here.

example/README.md
```diff
@@ -0,0 +1,149 @@
...
+## Common Patterns
+
+### Basic Load and Dump
+
+```go
+import "go.yaml.in/yaml/v4"
+
+// Load
+var config Config
+yaml.Unmarshal(yamlData, &config)
```
colinjlacy: Can't help but notice we're using `Unmarshal` here instead of `Load`. Is that intended?

colinjlacy: Can't help but notice we're using `Unmarshal` here instead of `Load`. Is that intended?

example/README.md
```diff
@@ -0,0 +1,149 @@
...
+
+// Dump
+data, _ := yaml.Marshal(&config)
```
colinjlacy: Same as above - intended?

example/README.md
```diff
@@ -0,0 +1,149 @@
...
+    yaml.WithSingleDocument(),
+)
+loader.Load(&config)
+
+// Dump with options
+dumper, _ := yaml.NewDumper(writer,
+    yaml.WithIndent(2),
+)
+dumper.Dump(&config)
+dumper.Close()
```
colinjlacy, line 134: Does `dumper` need to be closed as a read stream for memory safety? If so, is that documented somewhere?

example/dumper_indent_comparison/main.go
```diff
@@ -0,0 +1,121 @@
...
+		dumper, err := yaml.NewDumper(&buf, yaml.WithIndent(indent))
+		if err != nil {
+			panic(err)
+		}
+
+		if err := dumper.Dump(&cfg); err != nil {
+			panic(err)
+		}
+
+		if err := dumper.Close(); err != nil {
```
colinjlacy, line 67: [nit]: I notice in these examples that we're not using the common `defer` keyword that tends to be used when calling things like `file.Close()`. I find that to be an extremely common approach in Golang, so I'm wondering if it was a choice to be more explicit with an in-line close here.

ingydotnet: I imagine (but am not sure) that since dumper writes to a file, that file won't be written to (or at least completed) until you call close. I don't know when defers trigger.

example/README.md
ingydotnet: A dumper.Dump is called iteratively, one value at a time, dumping one document at a time. So until you close it, it's not done. It's waiting for the next Dump.

decode.go
ingydotnet: I'm not sure.

encode.go
ingydotnet: This is for WithFlowSimpleCollections. Small, simple seqs and maps get written in single line flow style.

cmd/go-yaml/main.go
ingydotnet: 7bd160b

example/README.md
ingydotnet: 7bd160b

ingydotnet: Read 7bd160b commit msg.

decode.go
ingydotnet: Read 7bd160b commit msg for details. Claude say the behavior is correct but possibly confusing. Comments were added.

example/README.md
ingydotnet: 7bd160b

options.go
```diff
@@ -0,0 +1,428 @@
...
+	}
+}
+
+// WithSingleDocument configures the Loader to only process the first document
+// in a YAML stream. After the first document is loaded, subsequent calls to
+// Load will return io.EOF.
+//
+// This is useful when you expect exactly one document and want behavior
+// similar to [Unmarshal].
+func WithSingleDocument() Option {
```
colinjlacy: This one stands out to me, not having an option to pass in variadic bool. It seems like something that would come up in an env-specific context or a CI pipeline, being able to pass in an env var `SINGLE_DOC` and have that determine the behavior. 
```
yaml.Load(y, yaml.WithSingleDocument(SINGLE_DOC))
```

Current code means having to be more verbose with:
```
if SINGLE_DOC {
  yaml.Load(y, yaml.WithSingleDocument())
} else {
  yaml.Load(y)
}
```
I'd prefer to see them all follow a similar functional signature, at least the ones with boolean args.

yaml.go
colinjlacy: I agree with @ccoVeille. The new Load method's comments shouldn't point people to the deprecated method that we want to move away from.

options.go
ingydotnet: 45c024f

ingydotnet:
@colinjlacy fixed docs away from marshal in 478fccd

ingydotnet:
@colinjlacy @ccoVeille 
The primary goal of this pull request was to add options and the dump/load APIs.
Along the way we decided to deprecate the old APIs.
This leaves us with quite a bit of internal confusion in the code and docs wrt names and such.
I suggest we not get too hung up on it and leave the remaining cleanup for further pull requests.

decode.go
```diff
@@ -316,13 +316,21 @@ var (
 	ifaceType      = generalMapType.Elem()
 )
 
-func newDecoder() *decoder {
+func newDecoder(opts *options) *decoder {
 	d := &decoder{
 		stringMapType:  stringMapType,
 		generalMapType: generalMapType,
-		uniqueKeys:     true,
+		uniqueKeys:     true, // default for legacy Decoder API (when opts == nil)
```
RaduBerinde, line 323: [nit] Consider inverting `uniqueKeys` to `allowDuplicateKeys` so it defaults to false.

encode.go
```diff
@@ -30,30 +30,61 @@ import (
 	"go.yaml.in/yaml/v4/internal/libyaml"
 )
 
+// Sentinel values for newEncoder parameters.
```
RaduBerinde, line 33: Very nice!

options.go
```diff
@@ -0,0 +1,433 @@
...
+
+// WithCompactSeqIndent configures whether the sequence indicator '- ' is
+// considered part of the indentation when dumping YAML content.
+//
+// If compact is true, '- ' is treated as part of the indentation.
+// If compact is false, '- ' is not treated as part of the indentation.
+// When called without arguments, defaults to true.
+func WithCompactSeqIndent(compact ...bool) Option {
+	return func(o *options) error {
+		if len(compact) > 1 {
```
RaduBerinde: This check belongs outside of the function (can be a panic).

options.go
```diff
@@ -0,0 +1,433 @@
...
+			return errors.New("yaml: WithCompactSeqIndent accepts at most one argument")
+		}
+		o.compactSeqIndent = len(compact) == 0 || compact[0]
```
RaduBerinde: If someone calls `WithCompactSeqIndent(x...)`, we are capturing the slice `x`, which the caller can modify after the call (see https://go.dev/play/p/ajg0AzHil8D).

We should move this expression outside of the func and store the result into a variable.

options.go
```diff
@@ -0,0 +1,433 @@
...
+//
+// When enabled, loading will return an error if the YAML input contains fields
+// that do not correspond to any fields in the target struct.
+// When called without arguments, defaults to true.
+func WithKnownFields(knownFields ...bool) Option {
+	return func(o *options) error {
+		if len(knownFields) > 1 {
+			return errors.New("yaml: WithKnownFields accepts at most one argument")
+		}
+		o.knownFields = len(knownFields) == 0 || knownFields[0]
```
RaduBerinde: Ditto

options.go
```diff
@@ -0,0 +1,433 @@
...
+// When false, non-ASCII characters are escaped (e.g., "caf\u00e9").
+// When called without arguments, defaults to true.
+//
+// The default is true.
+func WithUnicode(unicode ...bool) Option {
+	return func(o *options) error {
+		if len(unicode) > 1 {
+			return errors.New("yaml: WithUnicode accepts at most one argument")
+		}
+		o.unicode = len(unicode) == 0 || unicode[0]
```
RaduBerinde: Ditto, and more below

decode.go
ingydotnet: hmm. this one is interesting. `duplicateKeys` should be `error`, `ignore`, or `replace`.
I'm not sure how I feel about the v4 default. Probably `error`.

Thoughts?

ingydotnet: (ignore takes first, replace takes last)

options.go
ingydotnet: Does 52e08e4 do what you want @RaduBerinde ?

RaduBerinde: Yes, thanks!

decode.go
ingydotnet: I can do this as a followup PR (before we cut the next release)

ingydotnet: @copilot make an issue for this, please and thank you.

ingydotnet: That didn't work but a chat with copilot resulted in https://github.com/yaml/go-yaml/issues/218

RaduBerinde:
🎉 Thank you @ingydotnet!

encode.go
```diff
@@ -30,30 +30,61 @@ import (
...
 
-func newEncoderWithWriter(w io.Writer) *encoder {
-	e := &encoder{
-		emitter: libyaml.NewEmitter(),
+	// Apply options if provided, otherwise use defaults
+	if opts != nil {
+		e.indent = opts.indent
+		e.lineWidth = opts.lineWidth
+		e.emitter.CompactSequenceIndent = opts.compactSeqIndent
+		e.emitter.SetWidth(opts.lineWidth)
```
RaduBerinde, line 69: There is an inconsistency here - we only call `e.emitter.SetWidth(opts.lineWidth)` when options are passed. We should move this call after the `if`.

RaduBerinde: I think it would be better to always use an options object, instead of having two separate code paths for the defaults (depending if an option is passed or not).

ingydotnet: @RaduBerinde Can you please file these thoughts as an issue?

RaduBerinde: Opened PR #224

PR #213, opened 2025-12-19
Fix a bug causing 'make' commands to run slowly

ingydotnet:
These shell commands run on every `make` command ad were searching `.cache/` and other ignored dirs.

Sometimes `make` tab completion took over 5 seconds, as did `time make`:
```
$ time make
make: Nothing to be done for 'default'.

real	0m0.018s
user	0m0.003s
sys	0m0.016s
```

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -1,7 +1,7 @@
 # Auto-install https://github.com/makeplus/makes at specific commit:
 MAKES := .cache/makes
 MAKES-LOCAL := .cache/local
-MAKES-COMMIT ?= 40dee58a023162aa71ed9dc8c93973c7d73d32d5
+MAKES-COMMIT ?= 4962658786cf52734b3e416e49cba2abf08402a6
 $(shell [ -d $(MAKES) ] || ( \
   git clone -q https://github.com/makeplus/makes $(MAKES) && \
   git -C $(MAKES) reset -q --hard $(MAKES-COMMIT)))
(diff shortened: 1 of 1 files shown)


GNUmakefile
```diff
@@ -31,11 +31,8 @@ CLI-BINARY := go-yaml
 
 # Setup and include go.mk and shell.mk:
 GO-FILES := \
-  $(shell \
-    find . \
-    -not \( -path ./.cache -prune \) \
-    -name '*.go' | \
-    sort)
+  $(shell find *.go cmd internal yts -name '*.go' | sort)
```
ccoVeille: It's a bit strange to me to hard code the list of folders here.

It means it could be forgotten to add a new folder here if it gets created.

What about this


```suggestion
  $(shell find * -name '*.go' | sort)
```

This is a way to ignore all .dotfile folders, including .cache

ingydotnet: The problem with that is when you use `git worktree` and the worktrees all have `.cache/` dirs.

ingydotnet: I do agree that it's weird to hard code the directories. 
Let's keep thinking about it.

ingydotnet: Fixed this to be dynamic, fast, and ignore worktrees.

GNUmakefile
```diff
@@ -30,12 +30,10 @@ YTS-DIR := yts/testdata/$(YTS-TAG)
...
 
 # Setup and include go.mk and shell.mk:
-GO-FILES := \
-  $(shell \
-    find . \
-    -not \( -path ./.cache -prune \) \
-    -name '*.go' | \
-    sort)
+REPO-DIRS := $(shell find * -maxdepth 0 -type d \
+	       ! -exec test -f {}/.git \; -print)
```
ccoVeille, line 37: Maybe a comment about why it matters to do that. People won't think about Git Worktree

ingydotnet: ca24855

PR #217, opened 2025-12-21
Add go modernize linter

banjoh:
Add linter to check usage of modern Go idioms.

Ref: https://github.com/yaml/go-yaml/issues/4

https://golangci-lint.run/docs/linters/configuration/#modernize https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/modernize

The PR might look large, but the material changes are
- Adding linter to `.golangci.yaml`
- `interface{} -> any` conversion

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -7,6 +7,7 @@ linters:
   - misspell
   - nolintlint
   - staticcheck
+  - modernize
   disable:
   - errcheck
   - ineffassign
--- a/cmd/go-yaml/event.go
+++ b/cmd/go-yaml/event.go
@@ -200,7 +200,7 @@ func processEventsUnmarshal(profuse, compact bool) error {
 		firstDoc = false
 
 		// For unmarshal mode, use interface{} first to avoid preserving comments
-		var data interface{}
+		var data any
 		if err := yaml.Unmarshal(doc, &data); err != nil {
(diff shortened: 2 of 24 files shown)


ingydotnet:
@banjoh Can you change the commit message to "Add go modernise linter", please?

banjoh:
I'll also need to fix the linter errors

banjoh:
> @banjoh Can you change the commit message to "Add go modernise linter", please?

Done.

ingydotnet:
Yeah that's quite the change. Gobsmacked here.

banjoh:
> A bit shocked that we had all this interface, and only became aware of it now.
> 
> Can someone explain that?

I think this might be code either written before `any` was introduced, or people still using `interface{}` out of habit. With the linter we'll be able to catch these. I'm still finding myself using `interface{}` (when not vibing), but thanks to the linter my pre-commit hooks always slap my hand

ccoVeille:
> A bit shocked that we had all this interface, and only became aware of it now.
> 
> Can someone explain that?

We were supporting Go 1.16 as go-yaml/yaml

https://github.com/yaml/go-yaml/blob/73290cb4afb161a1276969b4f21a7b6f6e1111b0/go.mod#L3

`any` was brought with Go 1.18, the minimal version we decided to support with the v4

PR #223, opened 2025-12-23
Major CLI updates plus CLI data testing

ingydotnet:


cmd/go-yaml/cli_test.go
```diff
@@ -0,0 +1,193 @@
...
+		t.Fatalf("Failed to find test files: %v", err)
+	}
+
+	if len(testFiles) == 0 {
+		t.Skip("No test files found in testdata/")
+	}
+
+	// Build the CLI binary if it doesn't exist
+	binaryPath := "../../go-yaml"
+	if _, err := os.Stat(binaryPath); os.IsNotExist(err) {
```
ccoVeille: You should stop here and report an error if the file exists but is a folder.

Think about someone who created a go-yaml folder.

cmd/go-yaml/cli_test.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+	}
+
+	// Process each test file
+	for _, testFile := range testFiles {
+		testFileName := filepath.Base(testFile)
+		t.Run(testFileName, func(t *testing.T) {
+			runTestFile(t, testFile, binaryPath)
+		})
+	}
```
ccoVeille: I wouldn't use go build, but I would launch go run on each test. It might be slower, but I dislike the idea of having a binary that might be obsolete

cmd/go-yaml/cli_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+	for _, test := range tests {
+		if test.expected == "" {
+			continue // Skip if no expected output for this format
+		}
+
+		t.Run(test.field, func(t *testing.T) {
+			// Run the CLI command
+			cmd := exec.Command(binaryPath, test.flag)
```
ccoVeille: I would use go run cmd/go-yaml/... here

Relying on an existing binary sound like a problem to me

cmd/go-yaml/cli_test.go
```diff
@@ -0,0 +1,193 @@
...
+		if i < len(expLines) {
+			expLine = expLines[i]
+		}
+		if i < len(actLines) {
+			actLine = actLines[i]
+		}
+
+		if expLine != actLine {
+			result.WriteString("Line ")
+			result.WriteString(strings.Repeat(" ", len(strings.TrimSpace(expLine))+1))
```
ccoVeille: Why having this line?

I mean it adds as many space as needed, but then add a \n

So it adds space padding for a \n

No one would see this padding right?

cmd/go-yaml/event.go
```diff
@@ -63,13 +66,27 @@ func ProcessEvents(profuse, compact, unmarshal bool) error {
 	return processEventsDecode(profuse, compact)
 }
 
-// processEventsDecode uses Loader.Load for YAML processing
+// processEventsDecode uses yaml.Decoder for YAML processing with implicit field augmentation
```
ccoVeille: This sounds a bad conflict resolution

Loader.Load is the one used now

cmd/go-yaml/event.go
```diff
@@ -78,252 +95,307 @@ func processEventsDecode(profuse, compact bool) error {
...
+			enc, err := yaml.NewDumper(&buf)
+			if err != nil {
+				return fmt.Errorf("failed to create dumper: %w", err)
+			}
+			if err := enc.Dump([]*EventInfo{info}); err != nil {
 				enc.Close()
-				fmt.Print(buf.String())
+				return fmt.Errorf("failed to dump event info: %w", err)
 			}
+			enc.Close()
```
ccoVeille: Here and everywhere, the error returned by enc.Close() is not checked

cmd/go-yaml/event.go
```diff
@@ -78,252 +95,307 @@ func processEventsDecode(profuse, compact bool) error {
...
 	}
 
+	// Get implicit flags from libyaml parser
+	implicitFlags, err := getDocumentImplicitFlags(input)
+	if err != nil {
+		return err
+	}
+
 	// Split input into documents
 	documents := bytes.Split(input, []byte("---"))
```
ccoVeille: Not related to this change, but what would happen with a document being like this

```yaml
a: "---"
b: "---"
```

Maybe it's OK, but I would do this


```suggestion
	documents := bytes.Split(append([]byte("\n"), bytes.TrimSpace(input), []byte("\n")), []byte("\n---\n"))
```

cmd/go-yaml/event.go
```diff
@@ -491,11 +564,176 @@ func formatEventInfo(event *Event, profuse bool) *EventInfo {
...
 		if event.StartLine == event.EndLine && event.StartColumn == event.EndColumn {
-			info.Pos = fmt.Sprintf("%d;%d", event.StartLine, event.StartColumn)
+			// Single position
+			info.Pos = fmt.Sprintf("%d/%d", event.StartLine, event.StartColumn)
+		} else if event.StartLine == event.EndLine {
+			// Range on same line
+			info.Pos = fmt.Sprintf("%d/%d-%d", event.StartLine, event.StartColumn, event.EndColumn)
+		} else {
+			// Range across different lines
+			info.Pos = fmt.Sprintf("%d/%d-%d/%d", event.StartLine, event.StartColumn, event.EndLine, event.EndColumn)
```
ccoVeille: The classic way to write it would be this


```suggestion
			info.Pos = fmt.Sprintf("%d:%d", event.StartLine, event.StartColumn)
		} else if event.StartLine == event.EndLine {
			// Range on same line
			info.Pos = fmt.Sprintf("%d:%d-%d", event.StartLine, event.StartColumn, event.EndColumn)
		} else {
			// Range across different lines
			info.Pos = fmt.Sprintf("%d:%d-%d:%d", event.StartLine, event.StartColumn, event.EndLine, event.EndColumn)
```

cmd/go-yaml/token.go
```diff
@@ -374,26 +375,47 @@ func formatTokenInfo(token *Token, profuse bool) *TokenInfo {
...
-			info.Pos = fmt.Sprintf("%d;%d", token.StartLine, token.StartColumn)
+			// Single position
+			info.Pos = fmt.Sprintf("%d/%d", token.StartLine, token.StartColumn)
+		} else if token.StartLine == token.EndLine {
+			// Range on same line
+			info.Pos = fmt.Sprintf("%d/%d-%d", token.StartLine, token.StartColumn, token.EndColumn)
 		} else {
-			info.Pos = fmt.Sprintf("%d;%d-%d;%d", token.StartLine, token.StartColumn, token.EndLine, token.EndColumn)
+			// Range across different lines
+			info.Pos = fmt.Sprintf("%d/%d-%d/%d", token.StartLine, token.StartColumn, token.EndLine, token.EndColumn)
```
ccoVeille: Same remark about the fact : should be used instead of /

cmd/go-yaml/parser.go
```diff
@@ -120,9 +107,61 @@ func (p *Parser) Next() (*Token, error) {
...
+			Value:       string(value),
+			CommentType: commentType,
+			StartLine:   int(comment.StartMark.Line) + 1,
+			StartColumn: int(comment.StartMark.Column),
+			EndLine:     int(comment.EndMark.Line) + 1,
+			EndColumn:   int(comment.EndMark.Column),
+		}
+		p.pendingTokens = append(p.pendingTokens, commentToken)
+	}
+}
```
ccoVeille: Column is 1-based no ?

```suggestion
			StartColumn: int(comment.StartMark.Column) + 1,
			EndLine:     int(comment.EndMark.Line) + 1,
			EndColumn:   int(comment.EndMark.Column) + 1,
		}
		p.pendingTokens = append(p.pendingTokens, commentToken)
	}
}
```

cmd/go-yaml/node.go
```diff
@@ -87,24 +125,199 @@ func formatStyle(s yaml.Style) string {
...
+	case yaml.SingleQuotedStyle:
+		return "single"
+	case yaml.LiteralStyle:
+		return "literal"
+	case yaml.FoldedStyle:
+		return "folded"
+	case yaml.FlowStyle:
+		return "flow"
+	default:
+		return "plain"
```
ccoVeille: I would use this.

```suggestion
	case 0:
		return "plain"
	default:
		return "unknown"
```

But maybe I'm too defensive.

But formatStyle returns ""

cmd/go-yaml/event.go
ingydotnet:
```
$ ./go-yaml <<<'
a: "---"
b: "---"
' -n
kind: Document
content:
  - kind: Mapping
    content:
      - kind: Scalar
        text: a
      - kind: Scalar
        style: Double
        text: '---'
      - kind: Scalar
        text: b
      - kind: Scalar
        style: Double
        text: '---'

File: cmd/go-yaml/cli_test.go (hunk 1 above)
Comment (ingydotnet) [reply cmd/go-yaml/cli_test.go]: Fixed all your issues in 35ce0bc

File: cmd/go-yaml/event.go (hunk 8 above)
Comment (ingydotnet) [reply cmd/go-yaml/event.go]: I made this change but it has the unfortunate side effect of quoting all the position values. That's why I chose the format that I did. Later though I think we can fix the encoder to not do that. It does that because of the YAML 1.1 rules. That's why I filed #231.

File: cmd/go-yaml/event.go (hunk 7 above)
Comment (ingydotnet) [reply cmd/go-yaml/event.go]: Not a bug:
```
$ ./go-yaml <<<'
> a: "---"
b: "---"
> ' -e
- {event: DOCUMENT-START}
- {event: MAPPING-START}
- {event: SCALAR, value: a}
- {event: SCALAR, value: '---', style: Double}
- {event: SCALAR, value: b}
- {event: SCALAR, value: '---', style: Double}
- {event: MAPPING-END}
- {event: DOCUMENT-END}
```

File: cmd/go-yaml/event.go (hunk 5 above)
Comment (ccoVeille) [reply cmd/go-yaml/event.go]: Unless I'm wrong. This one wasn't changed

Comment (ingydotnet) [reply cmd/go-yaml/event.go]: Reverted that. Still resolving conflicts.

File: cmd/go-yaml/event.go (hunk 6 above)
Comment (ccoVeille) [reply cmd/go-yaml/event.go]: This is still relevant I think

File: cmd/go-yaml/main.go [hunk 12]
```diff
@@ -31,11 +31,54 @@ func hasUsefulContent(n *yaml.Node) bool {
...
+
+	// Apply version preset (if not using explicit API flags)
+	if v2Mode {
+		opts = append(opts, yaml.V2)
+	} else if v3Mode {
+		opts = append(opts, yaml.V3)
+	} else {
+		// V4 is the default, no need to explicitly add it
+		opts = append(opts, yaml.V4)
+	}
```
ccoVeille: Can't we just have a mode that would be v2, v3 and v4 by default

cmd/go-yaml/main.go
```diff
@@ -53,14 +96,36 @@ func main() {
...
 	// Shared flags
 	longMode := flag.Bool("l", false, "Long (block) formatted output")
-	unmarshalMode := flag.Bool("u", false, "Use Unmarshal instead of Decode for YAML input")
-	marshalMode := flag.Bool("m", false, "Use Marshal instead of Encode for YAML output")
+
+	// Version flags (long form only)
+	var v2Mode bool
+	var v3Mode bool
+	flag.BoolVar(&v2Mode, "v2", false, "Use V2 option preset")
+	flag.BoolVar(&v3Mode, "v3", false, "Use V3 option preset")
```
ccoVeille: Following https://github.com/yaml/go-yaml/pull/223#discussion_r2660037466

For the same reason the fact we could use --v2 and --v3,let me think we should have a --preset=v2/v3/v4

cmd/go-yaml/main.go
ingydotnet: I don't understand what you mean by that. The 4 is the default mode.

cmd/go-yaml/main.go
ingydotnet: I don't understand that either. What is the purpose of a `==preset`? (I'm not arguing against it. I just don't understand it yet.)

ingydotnet:
I'm putting this back into draft mode because I have some newer and bigger plans for how this should work.

ingydotnet:
Need to be done over #240.

ingydotnet:
Branch `REBASED-240-cli-fixes-and-refactoring` has this branch rebased on top of #240 's `internals-refactoring`.
I'll continue work on that branch until #240 is merged.

ingydotnet:
@ccoVeille I think you'll like 6108ac5 because I replaced the few options I added with a generic `-o` `--option` flag to specify any possible combination of options!

cmd/go-yaml/cli_test.go
```diff
@@ -0,0 +1,223 @@
...
+	for i := 0; i < maxLines; i++ {
+		expLine := ""
+		actLine := ""
+
+		if i < len(expLines) {
+			expLine = expLines[i]
+		}
+		if i < len(actLines) {
+			actLine = actLines[i]
+		}
```
ccoVeille, line 320: Nitpicking: You reduced maxLine to the lowest of both slice, so here you can use that


```suggestion
		expLine: = expLines[i]
		actLine: = actLines[i]
```

ccoVeille, line 320: Nitpicking

In addition of https://github.com/yaml/go-yaml/pull/223#discussion_r2707129028

If you online use maxLines to do not overflow the lines index, you can do this


```suggestion
	var result strings.Builder
	for i := 0; i < len(expLines) && I < len(actLines); i++ {
		expLine := expLines[i]
		actLine = actLines[i]
```

cmd/go-yaml/cli_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+func TestCLI(t *testing.T) {
+	// Find all test files in testdata/
+	testFiles, err := filepath.Glob("testdata/*.yaml")
+	if err != nil {
+		t.Fatalf("Failed to find test files: %v", err)
+	}
+
+	if len(testFiles) == 0 {
+		t.Skip("No test files found in testdata/")
```
ccoVeille: Shouldn't this cause a failure instead of skipping?

cmd/go-yaml/cli_test.go
```diff
@@ -0,0 +1,223 @@
...
+
+		if i < len(expLines) {
+			expLine = expLines[i]
+		}
+		if i < len(actLines) {
+			actLine = actLines[i]
+		}
+
+		if expLine != actLine {
+			result.WriteString(fmt.Sprintf("Line %d:\n", i+1))
```
ccoVeille: This is usually written like this


```suggestion
			fmt.Fsprintf(result, "Line %d:\n", i+1)
```

cmd/go-yaml/event.go
ccoVeille: An issue was opened

cmd/go-yaml/json.go
```diff
@@ -15,17 +15,21 @@ import (
 	"go.yaml.in/yaml/v4"
 )
 
-// ProcessJSON reads YAML from stdin and outputs JSON encoding
-func ProcessJSON(pretty, unmarshal bool) error {
-	if unmarshal {
-		return processJSONUnmarshal(pretty)
+// ProcessJSON reads YAML from reader and outputs JSON encoding
+func ProcessJSON(reader io.Reader, pretty, unmarshalMode, decodeMode bool, opts []yaml.Option) error {
```
ccoVeille: Usually, we are using variadic


```suggestion
func ProcessJSON(reader io.Reader, pretty, unmarshalMode, decodeMode bool, opts ...yaml.Option) error {
```

cmd/go-yaml/json.go
```diff
@@ -15,17 +15,21 @@ import (
...
+	}
+	// Default: use Load API with options
+	return processJSONLoad(reader, pretty, opts)
 }
 
-// processJSONDecode uses Loader.Load for YAML processing
-func processJSONDecode(pretty bool) error {
-	loader, err := yaml.NewLoader(os.Stdin)
+// processJSONLoad uses Loader.Load for YAML processing with options
+func processJSONLoad(reader io.Reader, pretty bool, opts []yaml.Option) error {
```
ccoVeille: Same here


```suggestion
func processJSONLoad(reader io.Reader, pretty bool, opts ...yaml.Option) error {
```

cmd/go-yaml/main.go
```diff
@@ -15,12 +15,278 @@ import (
...
 const version = "4.0.0.1"
 
+// stringSlice is a custom flag type for collecting multiple -o flags
+type stringSlice []string
+
+func (s *stringSlice) String() string {
+	if s == nil {
+		return ""
+	}
+	return fmt.Sprintf("%v", *s)
```
ccoVeille: No formatting, this should be preferred

```suggestion
	return fmt.Sprint(*s)
```

ccoVeille: I'll add linters when the PR paces will slow down

cmd/go-yaml/node.go
```diff
@@ -146,24 +184,201 @@ func formatStyle(s yaml.Style) string {
...
-	// Show !!str only if it was explicit in the input
+	// In profuse mode, always show tag
+	if profuse {
+		return tag
+	}
+
+	// Default YAML tags - only show if they were explicit in the input
 	switch tag {
-	case "!!str", "!!map", "!!seq":
+	case "!!str", "!!map", "!!seq", "!!int", "!!float", "!!bool", "!!null":
```
ccoVeille, line 232: What about !!merge tag that I didn't know until yesterday

ccoVeille (approved):
The PR became wild with huge amount of code changes.

It's OK, it's a CLI used for debug purpose.

I'm a bit worried the CLI become as big as the library itself. It's uneasy to maintain.

But these changes are interesting and needed.

Some remarks anyway. But they can be ignored if there is a need to merge quickly.

cmd/go-yaml/event.go
```diff
@@ -494,11 +566,176 @@ func formatEventInfo(event *Event, profuse bool) *EventInfo {
...
+type DocumentImplicitFlags struct {
+	StartImplicit bool
+	EndImplicit   bool
+}
+
+// getDocumentImplicitFlags extracts implicit flags for all documents
+func getDocumentImplicitFlags(input []byte) ([]*DocumentImplicitFlags, error) {
+	p := libyaml.NewParser()
+	if len(input) == 0 {
+		input = []byte{'\n'}
```
colinjlacy: I'm curious about what the purpose of explicitly setting a new line here is. Why not return an error?

cmd/go-yaml/event.go
```diff
@@ -494,11 +566,176 @@ func formatEventInfo(event *Event, profuse bool) *EventInfo {
...
+
+		ev.Delete()
+	}
+}
+
+// getEventsFromParser parses YAML input and extracts events with implicit field information
+func getEventsFromParser(input []byte, profuse bool) ([]*Event, error) {
+	p := libyaml.NewParser()
+	if len(input) == 0 {
+		input = []byte{'\n'}
```
colinjlacy, line 374: Same as above.

cmd/go-yaml/node.go
ingydotnet: Explicit use of `!!merge` isn't really a thing. 
We can discuss later.

cmd/go-yaml/main.go
ingydotnet: 99481be

cmd/go-yaml/json.go
ingydotnet: 10e80ea

cmd/go-yaml/json.go
ingydotnet: 10e80ea

cmd/go-yaml/cli_test.go
ingydotnet: 3f28489

cmd/go-yaml/cli_test.go
ingydotnet: 38d41f7

cmd/go-yaml/event.go
ingydotnet: This is intentional behavior, not an error. Empty input is valid YAML (represents an empty stream with no documents). The libyaml parser needs at least one byte to initialize,  
  so []byte{'\n'} is a minimal valid input that produces the expected empty result. Returning an error would be incorrect - echo "" | go-yaml -e should work and produce no events.

cmd/go-yaml/cli_test.go
ingydotnet: This sets maxLines to the maximum (not minimum) of both lengths. This is intentional - the diff needs to show trailing lines when arrays have different lengths. The bounds checks are required because we iterate up to the longer array.

ingydotnet:
@colinjlacy I addressed @ccoVeille's issues and then made several important improvements to the CLI.

I think it's in a great state now. 

Review it as closely as you want. I don't think there's too much to worry about, since it's just the CLI.

cmd/go-yaml/cli_test.go
```diff
@@ -0,0 +1,334 @@
...
+
+// TestMain builds the CLI binary once before running tests
+func TestMain(m *testing.M) {
+	// Create a temporary directory for the test binary
+	tmpDir, err := os.MkdirTemp("", "go-yaml-test")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to create temp dir: %v\n", err)
+		os.Exit(1)
+	}
+	defer os.RemoveAll(tmpDir)
```
ccoVeille: t.TempDir could be used and you could simplify here

cmd/go-yaml/cli_test.go
```diff
@@ -0,0 +1,334 @@
...
+	Out  string `yaml:"out"`
+}
+
+// CmdTestSuite is a sequence of command test cases
+type CmdTestSuite []CmdTestCase
+
+// TestCLICommands runs command-based tests from testdata/cmd/*.yaml
+func TestCLICommands(t *testing.T) {
+	// Find all test files in testdata/cmd/
+	testFiles, err := filepath.Glob("testdata/cmd/*.yaml")
```
ccoVeille: This should be the way to be architecture agnostic

```suggestion
	testFiles, err := filepath.Glob(filepath.Join("testdata", "cmd", "*.yaml"))
```

cmd/go-yaml/cli_test.go
```diff
@@ -0,0 +1,334 @@
...
+	"no-unicode-yaml":         "-o no-unicode -y",
+	"canonical-explicit-yaml": "-o canonical,explicit -y",
+	"v2-indent4-yaml":         "-o v2,indent=4 -y",
+	"stream-node":             "-o stream -n",
+	"stream-NODE":             "-o stream -N",
+}
+
+func TestCLI(t *testing.T) {
+	// Find all test files in testdata/
+	testFiles, err := filepath.Glob("testdata/*.yaml")
```
ccoVeille:
```suggestion
	testFiles, err := filepath.Glob(filepath.Join("testdata", "*.yaml"))
```

cmd/go-yaml/json.go
```diff
@@ -54,12 +58,40 @@ func processJSONDecode(pretty bool) error {
 	return nil
 }
 
+// processJSONDecode uses deprecated Decoder.Decode for YAML processing (no options support)
+func processJSONDecode(reader io.Reader, pretty bool, opts []yaml.Option) error {
```
ccoVeille:
```suggestion
func processJSONDecode(reader io.Reader, pretty bool, opts ...yaml.Option) error {
```

cmd/go-yaml/main.go
```diff
@@ -15,27 +15,324 @@ import (
...
+			continue
+		}
+		opt, err := parseOneOption(trimmed)
+		if err != nil {
+			return nil, err
+		}
+		opts = append(opts, opt...)
+	}
+	return opts, nil
+}
```
ccoVeille, line 250: This doesn't make sense to me


```suggestion
func parseOptionFlags(s string) ([]yaml.Option, error) {
	parts := strings.Split(s, ",")
	var opts []yaml.Option
	for _, part := range parts {
		trimmed := strings.TrimSpace(part)
		if trimmed == "" {
			continue
		}
		opt, err := parseOneOption(trimmed)
		if err != nil {
			return nil, err
		}
		opts = append(opts, opt...)
	}
	return opts, nil
}
```

Am I missing something?

cmd/go-yaml/main.go
```diff
@@ -15,27 +15,324 @@ import (
...
+
+// hasUsefulContent checks if a node has meaningful content to display.
+// Stream nodes are only returned when explicitly requested via -o stream,
+// so we always show them when present.
 func hasUsefulContent(n *yaml.Node) bool {
-	if n.Kind != yaml.StreamNode {
-		return true
+	return true
+}
```
ccoVeille: This doesn't look like an interface, so why is there a function that simply returns true?

ingydotnet: 82699bd

cmd/go-yaml/json.go
ingydotnet: 8e514fd

cmd/go-yaml/cli_test.go
ingydotnet: e3c138e

cmd/go-yaml/cli_test.go
ingydotnet: e3c138e

cmd/go-yaml/cli_test.go
ingydotnet: ca2ee8b

cmd/go-yaml/main.go
ingydotnet: 71ef5de

ingydotnet:
I'm going to merge this now, without requiring a full review from 2 team members. 
I feel confident in the tests and behavior. 
And it's only a debugging CLI. The diff is huge:
```
li-fixes-and-refactoring $ git diff origin/main | wc -l
5421
```
That's a lot to expect people to review. 
This should be the last major refactor of this size; at least for a while. 
I want to use this code to help me with the plugin API. 

I discussed this with @colinjlacy in Slack and he was ok with me merging if I was. And I am.

ccoVeille:
I thought I had approved it yesterday

cmd/go-yaml/json.go
ccoVeille: I'm surprised by the commit.

It appears you were able to change the method signature without changing the calls to this method.

This sounds not normal.

I checked and you were passing nil as a []yaml.Option

So now nil is passed as a Option, which is valid because nil satisfy any function signature.

It could lead to panic if nil is used as a function.

So two options here, remove the nil passed as argument of the method. Or simply consider removing the variadic argument if you don't plan to pass an option to this method

ccoVeille: I'm surprised by the commit.

It appears you were able to change the method signature without changing the calls to this method.

This sounds not normal.

I checked and you were passing nil as a []yaml.Option

So now nil is passed as a Option, which is valid because nil satisfy any function signature.

It could lead to panic if nil is used as a function.

So two options here, remove the nil passed as argument of the method. Or simply consider removing the variadic argument if you don't plan to pass an option to this method

PR #224, opened 2025-12-24
use explicit legacy options

RaduBerinde:
The legacy code paths pass nil options to encoder/decoder, which
handle `nil` options by setting specific settings. This change
simplifies this code by using explicit `legacyOptions` instead of
`nil`.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -317,21 +317,13 @@ var (
 )
 
 func newDecoder(opts *options) *decoder {
-	d := &decoder{
+	return &decoder{
 		stringMapType:  stringMapType,
 		generalMapType: generalMapType,
-		uniqueKeys:     true, // default for legacy Decoder API (when opts == nil)
+		knownFields:    opts.knownFields,
+		uniqueKeys:     opts.uniqueKeys,
+		aliases:        make(map[*Node]bool),
 	}
-	d.aliases = make(map[*Node]bool)
-
-	if opts != nil {
-		// For Loader API, opts comes from applyOptions() which initializes
-		// uniqueKeys to true by default, so this preserves the same behavior
-		d.knownFields = opts.knownFields
-		d.uniqueKeys = opts.uniqueKeys
-	}
-
-	return d
 }
 
 func (d *decoder) terror(n *Node, tag string, out reflect.Value) {
--- a/encode.go
+++ b/encode.go
@@ -34,7 +34,6 @@ import (
 // These provide clarity at call sites, similar to http.NoBody.
 var (
 	noWriter           io.Writer                 = nil
-	noOptions          *options                  = nil
 	noVersionDirective *libyaml.VersionDirective = nil
 	noTagDirective     []libyaml.TagDirective    = nil
 )
@@ -56,26 +55,20 @@ type encoder struct {
 // The writer parameter specifies the output destination for the encoder.
 // If writer is nil, the encoder will write to an internal buffer.
 func newEncoder(writer io.Writer, opts *options) *encoder {
-	e := &encoder{
-		emitter: libyaml.NewEmitter(),
-	}
+	emitter := libyaml.NewEmitter()
+	emitter.CompactSequenceIndent = opts.compactSeqIndent
+	emitter.SetWidth(opts.lineWidth)
+	emitter.SetUnicode(opts.unicode)
+	emitter.SetCanonical(opts.canonical)
+	emitter.SetLineBreak(opts.lineBreak)
 
-	// Apply options if provided, otherwise use defaults
-	if opts != nil {
-		e.indent = opts.indent
-		e.lineWidth = opts.lineWidth
-		e.emitter.CompactSequenceIndent = opts.compactSeqIndent
-		e.emitter.SetWidth(opts.lineWidth)
-		e.emitter.SetUnicode(opts.unicode)
-		e.emitter.SetCanonical(opts.canonical)
-		e.emitter.SetLineBreak(opts.lineBreak)
-		e.explicitStart = opts.explicitStart
-		e.explicitEnd = opts.explicitEnd
-		e.flowSimpleCollections = opts.flowSimpleCollections
-	} else {
-		// Default values when no options provided
-		e.emitter.SetUnicode(true)
-		e.lineWidth = 80
+	e := &encoder{
+		emitter:               emitter,
+		indent:                opts.indent,
+		lineWidth:             opts.lineWidth,
+		explicitStart:         opts.explicitStart,
+		explicitEnd:           opts.explicitEnd,
+		flowSimpleCollections: opts.flowSimpleCollections,
 	}
 
 	if writer != nil {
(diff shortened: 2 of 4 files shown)


options.go
```diff
@@ -440,3 +440,12 @@ func applyOptions(opts ...Option) (*options, error) {
 	}
 	return o, nil
 }
+
+// legacyOptions holds the default options for legacy APIs like
+// Marshal/Unmarshal.
+var legacyOptions = &options{
+	indent:     4,
+	lineWidth:  -1,
```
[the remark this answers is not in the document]
RaduBerinde: `e.lineWidth` is only used in `isSimpleCollection` where -1 and 80 are equivalent. The old code did not actually wrap lines at 80 in the nil options path because it didn't call `emitter.SetWidth()`

ingydotnet (approved):
This looks really good to me. 

Relying on @ccoVeille and/or @colinjlacy or others to look closer.

options.go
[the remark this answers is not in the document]
ingydotnet: Is it possible to use nil or 0 instead of -1? 

YAML uses -1 for no indentation but it leads to subtle issues and I want to change that in go-yaml.
I know that has nothing to do with this, but -1 sets off alarms for me. :)

ingydotnet:
@RaduBerinde I pinged you on the CNCF slack a couple days ago, but you never responded.

options.go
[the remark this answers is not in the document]
RaduBerinde: As far as I can tell -1 and 0 are different, there is at least one check in the emitter for `best_width >= 0`. Setting it to 0 fails a test:
```
    --- FAIL: TestEncodeToYAML/long_string_no_wrapping (0.00s)
        runner.go:54: got a: 'abcdefghijklmnopqrstuvwxyz ABCDEFGHIJKLMNOPQRSTUVWXYZ 1234567890 abcdefghijklmnopqrstuvwxyz
                ABCDEFGHIJKLMNOPQRSTUVWXYZ 1234567890 '
            ; want a: 'abcdefghijklmnopqrstuvwxyz ABCDEFGHIJKLMNOPQRSTUVWXYZ 1234567890 abcdefghijklmnopqrstuvwxyz ABC
```

encode.go
```diff
@@ -60,23 +59,16 @@ func newEncoder(writer io.Writer, opts *options) *encoder {
...
-		e.emitter.SetLineBreak(opts.lineBreak)
-		e.explicitStart = opts.explicitStart
-		e.explicitEnd = opts.explicitEnd
-		e.flowSimpleCollections = opts.flowSimpleCollections
-	} else {
-		// Default values when no options provided
-		e.emitter.SetUnicode(true)
-		e.lineWidth = 80
-	}
+	e.indent = opts.indent
```
ccoVeille: Here and everywhere, I have a problem with ops being *options, so can be possibly nil

So here and at other place it can panic.

Otherwise, I agree with the change. But then maybe calling the Loader, Dump with yaml.V3 option could be considered instead of using legacy option (it might need a refactoring)

I'm busy with Christmas. I'll review in detail in a few days

RaduBerinde: > Loader, Dump with yaml.V3 option

I think it would change the current behavior of Marshal (yaml.V3 has line wrapping).

Added a commit that cleans up the options internal a bit, and we now pass by value.

RaduBerinde:
Added a commit:

#### options can no longer return errors

Having `NewLoader` return an error is a bit clumsy and requires adding
new error checking in code that upgrades to v4.

We change `Options` to be a function that doesn't return an error.
Instead, we panic when an invalid `Option`s is created.

ccoVeille:
> Instead, we panic when an invalid Options is created.

This PR diverted too far. While I was OK initially, the first commit was fixing something.

Then the PR went in a direction to do not use pointer. A choice that can be considered opinionated, some may argue pointers are memory efficient. I get your point about code complexity it came with. But it's open for discussion.

Now the PR has encoder decoder options thing, which *might* be a good thing. But, it went out of scope maybe. At least it needs to be discussed.


But now, we have a third commit that change totally the way option were handled by using panic. The PR replaces something opinioated (do not use panic), to another opinionated position use panic.

I initially didn't use panic in option because I preferred it. My choice. My opinion, no more valuable than any other. It can be changed

My problem is that this PR that could have been approved yesterday as a no brainer as it was fixing something. Now, it became something that would require discussion, iterations.

It doesn't mean it cannot be approved, nor that I wouldn't approve it.

RaduBerinde:
Sure, I can split this up. The second commit was in response to your comment "Here and everywhere, I have a problem with ops being *options, so can be possibly nil" so I am confused :)

RaduBerinde:
Reverted to just the first commit. I can open subsequent PRs to discuss other changes.

ingydotnet:
@RaduBerinde can you amend the commit msg so that the tests pass?

decode.go
```diff
@@ -406,7 +398,9 @@ func (d *decoder) callObsoleteUnmarshaler(n *Node, u obsoleteUnmarshaler) (good
 // its types unmarshalled appropriately.
 //
 // If n holds a null value, prepare returns before doing anything.
-func (d *decoder) prepare(n *Node, out reflect.Value) (newout reflect.Value, unmarshaled, good bool) {
+func (d *decoder) prepare(
+	n *Node, out reflect.Value,
```
ccoVeille: Please revert this change that seems unrelated to the PR.

It looks like something your local IDE may have formatted. With golines maybe?

There is no reason to introduce such a change that would add unnecessary things to the history and git blame

encode.go
```diff
@@ -501,7 +493,9 @@ func (e *encoder) nilv() {
 	e.emitScalar("null", "", "", libyaml.PLAIN_SCALAR_STYLE, nil, nil, nil, nil)
 }
 
-func (e *encoder) emitScalar(value, anchor, tag string, style libyaml.ScalarStyle, head, line, foot, tail []byte) {
```
ccoVeille, line 504: Same as

https://github.com/yaml/go-yaml/pull/224#discussion_r2648433843

encode.go
ccoVeille, line 65: Could you rewrite to use


```go
	emitter := libyaml.NewEmitter()
	emitter.CompactSequenceIndent = opts.compactSeqIndent 
	emitter.SetWidth(opts.lineWidth)
	emitter.SetUnicode(opts.unicode) 
	emitter.SetCanonical(opts.canonical) 
	emitter.SetLineBreak(opts.lineBreak)
	
	e := &encoder{
		emitter: emitter,
		indent: opts.indent,
		lineWidth: opts.lineWidth,
		explicitStart: opts.explicitStart,
		explicitEnd: opts.explicitEnd,
		flowSimpleCollections: opts.flowSimpleCollections,
	}
```

(Pseudocode written on a phone, please adapt)

ccoVeille, line 57: The code can panic when opts is nil

This code is unexported, so this can only occur with an internal refactoring. I feel like keepin a pointer would be better for performance reason. 

I would like to suggest to panic on nil value with a clear message: `panic("opts cannot be nil")`

There are other options, but I feel like, panic might be a good option for now, in the scope of this PR.

A refactoring could be considered later.

decode.go
RaduBerinde: Oops, will revert, didn't notice

encode.go
RaduBerinde: Using a nil will panic with a clear message by definition in Go, adding our own check is unnecessary and non-idiomatic.

The structure is tiny, there is no performance issue with pointer vs value. The peformance concern is allocating options on the heap each time (which leads to extra GC work in hot loops), but unfortunately I don't think we can avoid that with the Option design.

ccoVeille: > Using a nil will panic with a clear message by definition in Go, adding our own check is unnecessary and non-idiomatic.

Indeed.

> The structure is tiny, there is no performance issue with pointer vs value.

Let's keep pointer then. It's a side case that might cause panic on a possible future refactoring. So at least two hypothesis ...
Let's keep changes minimal. We might need pointer later for something else.

> The peformance concern is allocating options on the heap each time (which leads to extra GC work in hot loops), but unfortunately I don't think we can avoid that with the Option design.

True

RaduBerinde:
Updated, thanks for the review.

PR #227, opened 2025-12-25
Support decoding !!binary into []byte

### LINKED ISSUE #226: Support decoding !!binary base64 values into []byte
This is re-report of https://github.com/go-yaml/yaml/issues/324

Currently, go-yaml can only decode `!!binary`-tagged values into a `string`. Although Go doesn't restrict strings to contain valid UTF-8, `[]byte` is still commonly used to represent non-text buffers.

Or, in other words, I'd like the following to work:

```go
var v []byte
yaml.Unmarshal([]byte(`!!binary MTIzCg==`), &v)
```

Such `!!binary`-tagged values can be easily encountered in the wild: PyYAML uses them for `bytes` type:

```
>>> print(yaml.dump(b"foobar"))
!!binary |
  Zm9vYmFy
```

Perhaps we could relax this even further to allow non-tagged string to decode into `[]byte` , but this is somewhat more ambiguous as you can't distinguish `[]byte` and `[]uint8`.

WGH-:
Although Go doesn't restrict strings to contain valid UTF-8, `[]byte` is still commonly used to represent non-text buffers.

PyYAML generates !!binary values for Python's bytes type, so these values are not rare.

This doesn't affect the marshal direction: `[]byte` is still dumped as a list of integers.

Closes #226

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -615,6 +615,14 @@ func (d *decoder) scalar(n *Node, out reflect.Value) bool {
 		}
 		out.SetString(n.Value)
 		return true
+	case reflect.Slice:
+		// allow decoding !!binary-tagged value into []byte specifically
+		if out.Type().Elem().Kind() == reflect.Uint8 {
+			if tag == binaryTag {
+				out.SetBytes([]byte(resolved.(string)))
+				return true
+			}
+		}
 	case reflect.Interface:
 		out.Set(reflect.ValueOf(resolved))
 		return true
--- a/decode_test.go
+++ b/decode_test.go
@@ -121,6 +121,9 @@ func init() {
 	decodeTypes.RegisterFactory("map[string]float64", func() any {
 		return make(map[string]float64)
 	})
+	decodeTypes.RegisterFactory("map[string][]byte", func() any {
+		return make(map[string][]byte)
+	})
 	decodeTypes.RegisterFactory("map[any]any", func() any {
 		return make(map[any]any)
 	})
--- a/testdata/decode.yaml
+++ b/testdata/decode.yaml
@@ -790,6 +790,28 @@
     want:
       v: test
 
+# !!binary tags
+- decode:
+    name: decoding !!binary into any
+    yaml: 'v: !!binary MTIzCg=='
+    type: map[string]any
+    want:
+      v: "123\n"
+
+- decode:
+    name: decoding !!binary into string
+    yaml: 'v: !!binary MTIzCg=='
+    type: map[string]string
+    want:
+      v: "123\n"
+
+- decode:
+    name: decoding !!binary into []byte
+    yaml: 'v: !!binary MTIzCg=='
+    type: map[string][]byte
+    want:
+      v: [0x31, 0x32, 0x33, 0x0A]
+
 # Anchors and aliases
 - decode:
     name: anchor as key with alias


decode_test.go
```diff
@@ -357,6 +357,14 @@ var unmarshalTests = []struct {
 		"a: !!binary |\n  " + strings.Repeat("A", 70) + "\n  ==\n",
 		map[string]string{"a": strings.Repeat("\x00", 52)},
 	},
+	{
```
ingydotnet: Can you add these tests under testdata?

ingydotnet (changes requested):

decode_test.go
WGH-: The last case is gonna look a bit ugly, though. Are you sure?

```yaml
# !!binary tags
- decode:
    name: decoding !!binary into any
    yaml: 'v: !!binary MTIzCg=='
    type: map[string]any
    want:
      v: "123\n"

- decode:
    name: decoding !!binary into string
    yaml: 'v: !!binary MTIzCg=='
    type: map[string]string
    want:
      v: "123\n"

- decode:
    name: decoding !!binary into []byte
    yaml: 'v: !!binary MTIzCg=='
    type: map[string][]byte
    want:
      v: [49, 50, 51, 10]
```

ccoVeille: With a clear comment, I would say it's OK

ccoVeille: What about this ?

```yaml
- decode:
    name: decoding !!binary into []byte
    yaml: 'v: !!binary MTIzCg=='
    type: map[string][]byte
    want:
      v: 
        - 0x31  # '1'
        - 0x32  # '2'
        - 0x33  # '3'
        - 0x0A  # '\n'
```

WGH-: I don't think writing sequence in multiple lines looks good here, but hexadecimal notation certainly helps.

WGH-: I've left both tests for now. I'll leave the final judgement to @ingydotnet, as I don't really have a strong opinion here

testdata/decode.yaml
```diff
@@ -790,6 +790,28 @@
     want:
       v: test
 
+# !!binary tags
+- decode:
+    name: decoding !!binary into any
+    yaml: 'v: !!binary MTIzCg=='
```
ingydotnet, line 796: Instead of key `v` here maybe use keys `any`, `string` and `bytes`.

testdata/decode.yaml
```diff
@@ -790,6 +790,28 @@
...
+    type: map[string]string
+    want:
+      v: "123\n"
+
+- decode:
+    name: decoding !!binary into []byte
+    yaml: 'v: !!binary MTIzCg=='
+    type: map[string][]byte
+    want:
+      v: [0x31, 0x32, 0x33, 0x0A]
```
ingydotnet, line 813: Seems good to me.

testdata/decode.yaml
```diff
@@ -790,6 +790,28 @@
     want:
       v: test
 
+# !!binary tags
```
ingydotnet, line 793: Thanks for making the data tests!

decode_test.go
```diff
@@ -357,6 +360,14 @@ var unmarshalTests = []struct {
...
```
ingydotnet: I think we can drop these. They don't add much value.

Note: I like how the data tests have the `name` description field. Maybe we should have a comment description in front of every test like these in Go.

decode.go
```diff
@@ -615,6 +615,14 @@ func (d *decoder) scalar(n *Node, out reflect.Value) bool {
 		}
 		out.SetString(n.Value)
 		return true
+	case reflect.Slice:
```
ingydotnet, line 618: Is this too general? Does it affect things besides `!!binary`?

WGH-: No.  There's no `default` case, so non-`[]byte` non-`!!binary` will still fall through to `d.terror(n, tag, out)` in the end.

testdata/decode.yaml
WGH-: Not sure about that. We're testing unmarshaling the same YAML document into different Go types. It would look more organic if keys remained the same, too.

ingydotnet (approved):
Looks tight to me.

decode_test.go
ingydotnet: I think it looks great.

testdata/decode.yaml
ingydotnet: ok

PR #228, opened 2025-12-26
Error on characters not allowed in YAML tags

ingydotnet:
I realized that we were allowing commas in tags, which are forbidden by the YAML spec, but then also realized there were a few more characters like this.

### DIFF
--- a/internal/libyaml/scanner_test.go
+++ b/internal/libyaml/scanner_test.go
@@ -5,7 +5,6 @@ package libyaml
 
 import (
 	"bytes"
-	"regexp"
 	"testing"
 
 	"go.yaml.in/yaml/v4/internal/testutil/assert"
@@ -94,10 +93,8 @@ func runScanErrorTest(t *testing.T, tc TestCase) {
 	if wantError {
 		assert.Truef(t, scanErr != nil, "Expected scanner error, but got none")
 		// Check error message against regex pattern if provided
-		if tc.Like != "" && scanErr != nil {
-			matched, err := regexp.MatchString(tc.Like, scanErr.Error())
-			assert.NoErrorf(t, err, "invalid regex pattern: %s", tc.Like)
-			assert.Truef(t, matched, "error %q should match pattern %q", scanErr.Error(), tc.Like)
+		if tc.Like != "" {
+			assert.ErrorMatchesf(t, tc.Like, scanErr, "")
 		}
 	} else {
 		assert.Truef(t, scanErr == nil, "Expected no scanner error, but got %v", scanErr)


perlpunk:
Note that YAML 1.1 allows those characters, so the change is incompatible and could break things.
I remember that when I removed the comma from allowed chars in libyaml [in 2020](https://github.com/yaml/libyaml/pull/179), there was at least one real world case mentioned: https://github.com/yaml/libyaml/issues/196#issuecomment-660639626

ingydotnet:
> Note that YAML 1.1 allows those characters, so the change is incompatible and could break things. I remember that when I removed the comma from allowed chars in libyaml [in 2020](https://github.com/yaml/libyaml/pull/179), there was at least one real world case mentioned: [yaml/libyaml#196 (comment)](https://github.com/yaml/libyaml/issues/196#issuecomment-660639626)

I think we should disallow it, and see who complains. We have the new Options system to help use out if so.

internal/libyaml/scanner.go
```diff
@@ -2040,13 +2040,13 @@ func (parser *Parser) scanTag(token *Token) error {
...
-			if err := parser.scanTagURI(false, nil, start_mark, &suffix); err != nil {
+			// Scan the suffix now (short form - flow indicators not allowed).
+			if err := parser.scanTagURI(false, false, nil, start_mark, &suffix); err != nil {
 				return err
 			}
 		} else {
-			// It wasn't a handle after all.  Scan the rest of the tag.
-			if err := parser.scanTagURI(false, handle, start_mark, &suffix); err != nil {
+			// It wasn't a handle after all.  Scan the rest of the tag (short form).
+			if err := parser.scanTagURI(false, false, handle, start_mark, &suffix); err != nil {
```
ccoVeille, line 2032: The arguments of these method became too numerous. It's complicated to figure what true/false but also others may be about. I would refactor to pass a structure.

This refactoring doesn't need to be done in this PR, it's not blocking, but it should be done at some point. There are linters we could enable to report when su j a complexity is reached

internal/libyaml/scanner_test.go
```diff
@@ -91,6 +92,12 @@ func runScanErrorTest(t *testing.T, tc TestCase) {
 	}
 	if wantError {
 		assert.Truef(t, scanErr != nil, "Expected scanner error, but got none")
+		// Check error message against regex pattern if provided
+		if tc.Like != "" && scanErr != nil {
+			matched, err := regexp.MatchString(tc.Like, scanErr.Error())
```
ccoVeille: You have assert.ErrorMatches (something like that) that is about regexp

yts/known-failing-tests
```diff
@@ -1,5 +1,6 @@
 TestYAMLSuite/27NA/UnmarshalTest
 TestYAMLSuite/27NA/EventComparisonTest
+TestYAMLSuite/229Q/JSONComparisonTest
```
ccoVeille, line 3: Why is this one failing ?

internal/libyaml/scanner.go
ingydotnet: Sounds good. Let's take this on later.

yts/known-failing-tests
ingydotnet: I'm not sure. Open an issue for it?

internal/libyaml/scanner_test.go
ingydotnet: Fixed in bb52fa7

yts/known-failing-tests
ccoVeille: I mean you are the one that adds it to this PR.

Something was working before. And it's no longer works because of the changes you did.

It doesn't sound good.

It's not about opening an issue for fixing it.

If a test from the suite is no working it sounds like we are facing regression

ingydotnet: Sorry, I misunderstood the situation you were talking about. Looking into it.

ingydotnet: @ccoVeille Does https://gist.github.com/ingydotnet/f8d8b7cbd822ee502f3ebf63ca169ca2 make sense to you?

@colinjlacy ?

colinjlacy: This makes sense to me. Parse comparison of int values with JSON should fail by nature of the different ways in which numbers are handled. I think it’s worth exploring if/how some tests aren’t failing to see if more examples like this pop up.

ccoVeille: Let's merge like this and open an issue to normalize such issue, I'm sure we can have a piece of code that will help us to identify when the issue is simply about having a float instead of an int in the test.

ingydotnet: > Let's merge like this and open an issue to normalize such issue, I'm sure we can have a piece of code that will help us to identify when the issue is simply about having a float instead of an int in the test.

@ccoVeille @colinjlacy I'm happy to merge, but I need two approvals! :smile:

colinjlacy (approved):

ccoVeille (approved):

yts/known-failing-tests
ccoVeille: > Let's merge like this and open an issue to normalize such issue, I'm sure we can have a piece of code that will help us to identify when the issue is simply about having a float instead of an int in the test.

I opened https://github.com/yaml/go-yaml/issues/239

ingydotnet: :+1:

PR #230, opened 2025-12-29
Add WithStreamNodes option to expose stream nodes

ingydotnet:
Implements WithStreamNodes() loader option to return YAML stream boundary nodes, exposing stream-level metadata (encoding, version directives, tag directives) that isn't attached to any specific document.

When enabled, the loader returns an interleaved pattern of StreamNode and DocumentNode:
- Empty stream: [StreamNode]
- Single document: [StreamNode, DocumentNode, StreamNode]
- Multi-document: [StreamNode, Doc, StreamNode, Doc, StreamNode]

Stream nodes capture metadata from the YAML stream:
- Encoding (UTF-8, UTF-16LE, UTF-16BE) on all stream nodes
- Version directive (%YAML) on first stream node only
- Tag directives (%TAG) on first stream node only
- Stream position information (line, column)

Without this option, the loader maintains backward compatibility by returning only DocumentNodes as before.

CLI Changes:
- Added -S/--stream flag to display stream nodes in node output mode
- When -S is used, all nodes (including empty stream nodes) are shown with --- separators
- Without -S, empty stream nodes are filtered from output

Implementation:
- Added StreamNode to Kind constants (yaml.go)
- Extended Node struct with stream-specific fields: Encoding, Version, TagDirectives (yaml.go)
- Added WithStreamNodes() option function (options.go)
- Modified parser to track stream state and emit stream nodes at boundaries (decode.go)
- Added accessor methods for event metadata (internal/libyaml/yaml.go)
- Updated CLI to support -S flag and stream node formatting (cmd/go-yaml/main.go, cmd/go-yaml/node.go)

Testing:
- Created testdata/stream.yaml with 7 comprehensive test cases
- Renamed stream_test.go to load_test.go to reflect its role as home for Loader API tests
- Implemented data-driven test infrastructure using datatest pattern consistent with other test suites (node_test.go, decode_test.go)
- Test format uses type-as-key pattern with 4-letter keys (name, yaml, with, want) and negative boolean options defaulting to false

### DIFF
--- a/internal/libyaml/yaml.go
+++ b/internal/libyaml/yaml.go
@@ -342,10 +342,10 @@ type Event struct {
 	encoding Encoding
 
 	// The version directive (for DOCUMENT_START_EVENT).
-	version_directive *VersionDirective
+	versionDirective *VersionDirective
 
 	// The list of tag directives (for DOCUMENT_START_EVENT).
-	tag_directives []TagDirective
+	tagDirectives []TagDirective
 
 	// The comments
 	HeadComment []byte
@@ -381,10 +381,10 @@ func (e *Event) MappingStyle() MappingStyle   { return MappingStyle(e.Style) }
 func (e *Event) GetEncoding() Encoding { return e.encoding }
 
 // GetVersionDirective returns the version directive (for DOCUMENT_START_EVENT).
-func (e *Event) GetVersionDirective() *VersionDirective { return e.version_directive }
+func (e *Event) GetVersionDirective() *VersionDirective { return e.versionDirective }
 
 // GetTagDirectives returns the tag directives (for DOCUMENT_START_EVENT).
-func (e *Event) GetTagDirectives() []TagDirective { return e.tag_directives }
+func (e *Event) GetTagDirectives() []TagDirective { return e.tagDirectives }
 
 // Nodes
 
--- a/options.go
+++ b/options.go
@@ -172,9 +172,9 @@ func WithUniqueKeys(uniqueKeys ...bool) Option {
 //   - Position information (Line, Column)
 //
 // An empty YAML stream returns a single StreamNode.
-// When called without arguments, defaults to true.
+// When called without arguments, defaults to true (enables stream nodes).
 //
-// The default is false.
+// When this option is not used, stream nodes are not returned.
 func WithStreamNodes(enable ...bool) Option {
 	return func(o *options) error {
 		if len(enable) > 1 {
--- a/internal/libyaml/api.go
+++ b/internal/libyaml/api.go
@@ -289,10 +289,10 @@ func NewStreamEndEvent() Event {
 // NewDocumentStartEvent creates a new DOCUMENT-START event.
 func NewDocumentStartEvent(version_directive *VersionDirective, tag_directives []TagDirective, implicit bool) Event {
 	return Event{
-		Type:              DOCUMENT_START_EVENT,
-		version_directive: version_directive,
-		tag_directives:    tag_directives,
-		Implicit:          implicit,
+		Type:             DOCUMENT_START_EVENT,
+		versionDirective: version_directive,
+		tagDirectives:    tag_directives,
+		Implicit:         implicit,
 	}
 }
 
--- a/internal/libyaml/emitter.go
+++ b/internal/libyaml/emitter.go
@@ -367,14 +367,14 @@ func (emitter *Emitter) emitStreamStart(event *Event) error {
 func (emitter *Emitter) emitDocumentStart(event *Event, first bool) error {
 	if event.Type == DOCUMENT_START_EVENT {
 
-		if event.version_directive != nil {
-			if err := emitter.analyzeVersionDirective(event.version_directive); err != nil {
+		if event.versionDirective != nil {
+			if err := emitter.analyzeVersionDirective(event.versionDirective); err != nil {
 				return err
 			}
 		}
 
-		for i := 0; i < len(event.tag_directives); i++ {
-			tag_directive := &event.tag_directives[i]
+		for i := 0; i < len(event.tagDirectives); i++ {
+			tag_directive := &event.tagDirectives[i]
 			if err := emitter.analyzeTagDirective(tag_directive); err != nil {
 				return err
 			}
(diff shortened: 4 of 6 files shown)


decode.go
```diff
@@ -31,12 +31,17 @@ import (
...
-	textless bool
+	parser       libyaml.Parser
+	event        libyaml.Event
+	doc          *Node
+	anchors      map[string]*Node
+	doneInit     bool
+	textless     bool
+	streamNodes  bool             // enable stream node emission
+	returnStream bool             // flag to return stream node next
+	streamNode   *Node            // pending stream node
```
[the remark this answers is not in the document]
ingydotnet: fixed in 947d57e

cmd/go-yaml/node.go
```diff
@@ -53,9 +64,50 @@ func FormatNode(n yaml.Node) *NodeInfo {
...
+	switch e {
+	case yaml.EncodingUTF8:
+		return "UTF-8"
+	case yaml.EncodingUTF16LE:
+		return "UTF-16LE"
+	case yaml.EncodingUTF16BE:
+		return "UTF-16BE"
+	default:
+		return "Any"
+	}
```
ccoVeille, line 102: I would use


```suggestion
func formatEncoding(e yaml.Encoding) string {
	switch e {
	case yaml.EncodingUTF8:
		return "UTF-8"
	case yaml.EncodingUTF16LE:
		return "UTF-16LE"
	case yaml.EncodingUTF16BE:
		return "UTF-16BE"
	case yaml.ANY_ENCODING:
		return "Any"
	default:
		return "Unknown"
	}
```

Also this could be moved as a Stringer of yaml.Encoding type

decode.go
```diff
@@ -63,8 +67,17 @@ func (p *parser) init() {
...
+	if p.peek() == libyaml.STREAM_START_EVENT {
+		p.encoding = p.event.GetEncoding()
+	}
 	p.expect(libyaml.STREAM_START_EVENT)
 	p.doneInit = true
+
+	// If stream nodes are enabled, prepare to return the first stream node
+	if p.streamNodes {
+		p.returnStream = true
+	}
```
ccoVeille: This is OK. But this could also work, no?

```suggestion
	// If stream nodes are enabled, prepare to return the first stream node
	p.returnStream = p.streamNodes
```

decode.go
```diff
@@ -118,8 +131,66 @@ func (p *parser) anchor(n *Node, anchor []byte) {
...
+	if !p.textless && p.event.Type != libyaml.NO_EVENT {
+		n.Line = p.event.StartMark.Line + 1
+		n.Column = p.event.StartMark.Column + 1
+	}
+	return n
+}
+
+// captureDirectives captures version and tag directives from upcoming DOCUMENT_START.
+func (p *parser) captureDirectives(n *Node) {
+	if p.peek() == libyaml.DOCUMENT_START_EVENT {
```
ccoVeille: I tend to avoid branching and nested `if` statement

I would like to suggest this


```suggestion
	if p.peek() != libyaml.DOCUMENT_START_EVENT {
		return
	}
```

decode.go
```diff
@@ -182,6 +253,12 @@ func (p *parser) document() *Node {
 		n.FootComment = string(p.event.FootComment)
 	}
 	p.expect(libyaml.DOCUMENT_END_EVENT)
+
+	// If stream nodes enabled, prepare to return a stream node next
+	if p.streamNodes {
+		p.returnStream = true
+	}
```
ccoVeille: Same here


```suggestion
	// If stream nodes enabled, prepare to return a stream node next
	p.returnStream = p.streamNodes
```

options.go
```diff
@@ -162,6 +163,34 @@ func WithUniqueKeys(uniqueKeys ...bool) Option {
...
+func WithStreamNodes(enable ...bool) Option {
+	if len(enable) > 1 {
+		panic("yaml: WithStreamNodes accepts at most one argument")
+	}
+	val := len(enable) == 0 || enable[0]
+	return func(o *options) error {
+		o.streamNodes = val
+		return nil
+	}
+}
```
ccoVeille, line 186: I would use this to avoid panic

```suggestion
func WithStreamNodes(enable ...bool) Option {
	var err error
	if len(enable) > 1 {
		err = errors.New("yaml: WithStreamNodes accepts at most one argument")
	}
	val := len(enable) == 0 || enable[0]
	return func(o *options) error {
		o.streamNodes = val
		return err
	}
}
```

ccoVeille (approved):
Some minor comments, but we could address them later or never.

cmd/go-yaml/node.go
ingydotnet: @ccoVeille does a7d08f2 address your review issues?

ingydotnet:
@ccoVeille I made a7d08f2 to try to address your review points.

internal/libyaml/yaml.go
```diff
@@ -365,6 +377,15 @@ func (e *Event) ScalarStyle() ScalarStyle     { return ScalarStyle(e.Style) }
 func (e *Event) SequenceStyle() SequenceStyle { return SequenceStyle(e.Style) }
 func (e *Event) MappingStyle() MappingStyle   { return MappingStyle(e.Style) }
 
+// GetEncoding returns the stream encoding (for STREAM_START_EVENT).
+func (e *Event) GetEncoding() Encoding { return e.encoding }
+
+// GetVersionDirective returns the version directive (for DOCUMENT_START_EVENT).
+func (e *Event) GetVersionDirective() *VersionDirective { return e.version_directive }
```
colinjlacy: Should be camel-case. I realize that this is a ref to an existing field, but flagging it now so that we don't forget.

options.go
```diff
@@ -152,12 +149,38 @@ func WithUnicode(unicode ...bool) Option {
...
+// StreamNodes contain metadata about the stream including:
+//   - Encoding (UTF-8, UTF-16LE, UTF-16BE)
+//   - YAML version directive (%YAML)
+//   - Tag directives (%TAG)
+//   - Position information (Line, Column)
+//
+// An empty YAML stream returns a single StreamNode.
+// When called without arguments, defaults to true.
+//
+// The default is false.
```
colinjlacy: This confuses me. Does `The default is false` mean to say that if this option is not included, then stream nodes are not enabled? If so, I would rephrase it that way.

colinjlacy (approved):
One small change suggested for a comment. Code looks good.

ccoVeille (approved):

internal/libyaml/yaml.go
ingydotnet: d6ab85d

options.go
ingydotnet: d6ab85d

PR #236, opened 2026-01-04
Add TypeError.Strings() for v3 to v4 migration

ingydotnet:
In v3, TypeError.Errors was []string. In v4, it changed to []*UnmarshalError to provide structured error information with line and column numbers. This is a breaking change for code that directly accesses the Errors field.

The new Strings() method eases migration by providing a simple way to get the v3-style string slice without changing existing logic:

  // v3 code
  for _, msg := range typeErr.Errors { ... }

  // v4 migration (minimal change)
  for _, msg := range typeErr.Strings() { ... }

This allows users to migrate with a one-line change, deferring the adoption of structured errors until they're ready. New code should still access the Errors field directly to take advantage of the richer error information.

This was added because Claude saw this as the only breaking issue from V3 to V4. 

See https://gist.github.com/ingydotnet/8a2becd45d032655a327fe361b3fe234 for more info.

### DIFF
--- a/yaml.go
+++ b/yaml.go
@@ -677,6 +677,20 @@ func (e *TypeError) As(target any) bool {
 	return false
 }
 
+// Strings returns the error messages as a string slice.
+//
+// This method is provided for compatibility with code migrating from v3,
+// where TypeError.Errors was []string. New code should access the Errors
+// field directly to get structured error information including line and
+// column numbers.
+func (e *TypeError) Strings() []string {
+	result := make([]string, len(e.Errors))
+	for i, err := range e.Errors {
+		result[i] = err.Error()
+	}
+	return result
+}
+
 type Kind uint32
 
 const (
--- a/decode_test.go
+++ b/decode_test.go
@@ -1782,3 +1782,19 @@ func TestParserErrorUnknownAnchorPosition(t *testing.T) {
 		assert.DeepEqual(t, expected, asErr)
 	}
 }
+
+func TestTypeError_Strings(t *testing.T) {
+	// Create a TypeError with multiple errors
+	typeErr := &yaml.TypeError{
+		Errors: []*yaml.UnmarshalError{
+			{Err: errors.New("cannot unmarshal string into int"), Line: 5, Column: 3},
+			{Err: errors.New("cannot unmarshal bool into string"), Line: 10, Column: 7},
+		},
+	}
+
+	strings := typeErr.Strings()
+
+	assert.Equal(t, 2, len(strings))
+	assert.Equal(t, "line 5: cannot unmarshal string into int", strings[0])
+	assert.Equal(t, "line 10: cannot unmarshal bool into string", strings[1])
+}


yaml.go
```diff
@@ -688,6 +688,20 @@ func (e *TypeError) As(target any) bool {
 	return false
 }
 
+// Strings returns the error messages as a string slice.
+//
+// This method is provided for compatibility with code migrating from v3,
+// where TypeError.Errors was []string. New code should access the Errors
+// field directly to get structured error information including line and
```
colinjlacy, line 684: Should this be released as deprecated so as to drive home the idea that this is not the prescribed way to go?

ingydotnet: @colinjlacy Sounds good to me, would you mind pushing a commit to this that does that?

ccoVeille (approved):
LGTM 👍 as long we address the issue of having a deprecated godoc

Another solution for this PR could have been to rename Errors field to Errs and add Errors []string that would be the current Strings.

Also, now we have interfaces that satisfy errors.As bool and errors.Is bool, we could maybe move TypeError.Errors []*UnmarshalError to an unexported field, allowing us to use Errors()

ingydotnet:
> LGTM 👍 as long we address the issue of having a deprecated godoc
> 
> Another solution for this PR could have been to rename Errors field to Errs and add Errors []string that would be the current Strings.
> 
> Also, now we have interfaces that satisfy errors.As bool and errors.Is bool, we could maybe move TypeError.Errors []*UnmarshalError to an unexported field, allowing us to use Errors()

What do you think provides the most painless migration path?

ccoVeille:
One solution would be to add a new struct

LegacyTypeError that would be the exact old code.

And then people could use this type to get the old error. Then we could play with Unwrap in TypeError

Then we could have zero migration changes except asking people to use LegacyTypeError it's Errors field would be []string

ccoVeille:
Here is a short piece of code that should work

```go

// Unwrap converts the TypeError into a LegacyTypeError for compatibility
// with previous versions of the yaml package.
//
// [errors.Unwrap] will call this method when unwrapping errors.
func (e *TypeError) Unwrap() error {
	errLegacy := &LegacyTypeError{
		Errors: make([]string, len(e.Errors)),
	}

	for i, err := range e.Errors {
		errLegacy.Errors[i] = err.Error()
	}

	return errLegacy
}

type LegacyTypeError struct {
	Errors []string
}

func (e *LegacyTypeError) Error() string {
	return fmt.Sprintf("yaml: unmarshal errors:\n  %s", strings.Join(e.Errors, "\n  "))
}
```

My comment is based on the fact, the piece of code that can be found online are pretty dense, it's better to ask people to change one line about the errors.As than the code that use the .Errors field on this struct

https://github.com/search?q=language%3Ago%20%22yaml.TypeError%22%20.Errors%20&type=code

Here are some examples

https://github.com/rhysd/actionlint/blob/a0ce74d8cfe85724a6a47c62bc03a6929d7ad6ef/action_metadata.go#L274

https://github.com/bufbuild/buf/blob/93d4b05ab2be2e380a44776be1b5bb9ac7324f3d/private/pkg/encoding/encoding.go#L200

ccoVeille:
@colinjlacy what do you think about all this? Thanks

I might be overthoverthinking maybe

PR #244, opened 2026-01-11
Replace LoadAll/DumpAll with WithAllDocuments option

ingydotnet:
Remove LoadAll and DumpAll functions from the public API and replace
with a WithAllDocuments option for Load and Dump functions. This simplifies the
API surface while providing better type safety for multi-document
operations.

Changes:
- Add WithAllDocuments(...bool) option to enable multi-document mode
- Modify Load to support both single-doc (strict) and multi-doc modes
  - Single-doc mode (default): requires exactly one document
  - Multi-doc mode (WithAllDocuments): loads all documents into a slice
  - Supports typed slices (e.g., []Config) for type-safe loading
- Modify Dump to support multi-document output with WithAll
- Remove LoadAll and DumpAll functions
- Add comprehensive test coverage for new behavior
- Update all documentation and examples

Benefits:
- Smaller API surface (removes 2 functions, adds 1 option)
- Better type safety with typed slice support
- Fail-fast behavior prevents silent bugs (strict single-doc mode)
- Consistent with existing functional options pattern

Breaking changes:
- Load now requires exactly one document by default (errors on 0 or 2+)
- LoadAll and DumpAll functions removed (use WithAllDocuments option instead)

 Migration:
  Before:
```
docs, err := yaml.LoadAll(data)
```
  After:
```
var docs []any
err := yaml.Load(data, &docs, yaml.WithAllDocuments())
```

  Before: `data, err := yaml.DumpAll(docs)`
  After:  `data, err := yaml.Dump(docs, yaml.WithAllDocuments())`

### DIFF
--- a/loader.go
+++ b/loader.go
@@ -24,12 +24,12 @@ import (
 // If zero documents are found, it returns an error.
 // If multiple documents are found, it returns an error.
 //
-// Use WithAll() to load all documents into a slice:
+// Use WithAllDocuments() to load all documents into a slice:
 //
 //	var configs []Config
-//	yaml.Load(multiDocYAML, &configs, yaml.WithAll())
+//	yaml.Load(multiDocYAML, &configs, yaml.WithAllDocuments())
 //
-// When WithAll is used, out must be a pointer to a slice.
+// When WithAllDocuments is used, out must be a pointer to a slice.
 // Each document is decoded into the slice element type.
 // Zero documents results in an empty slice (no error).
 //
@@ -67,7 +67,7 @@ func Load(in []byte, out any, opts ...Option) error {
 		return err
 	}
 
-	if o.All {
+	if o.AllDocuments {
 		// Multi-document mode: out must be pointer to slice
 		return loadAll(in, out, o)
 	}
@@ -81,14 +81,14 @@ func loadAll(in []byte, out any, opts *libyaml.Options) error {
 	outVal := reflect.ValueOf(out)
 	if outVal.Kind() != reflect.Pointer || outVal.IsNil() {
 		return &TypeError{Errors: []*libyaml.ConstructError{{
-			Err: errors.New("yaml: WithAll requires a non-nil pointer to a slice"),
+			Err: errors.New("yaml: WithAllDocuments requires a non-nil pointer to a slice"),
 		}}}
 	}
 
 	sliceVal := outVal.Elem()
 	if sliceVal.Kind() != reflect.Slice {
 		return &TypeError{Errors: []*libyaml.ConstructError{{
-			Err: errors.New("yaml: WithAll requires a pointer to a slice"),
+			Err: errors.New("yaml: WithAllDocuments requires a pointer to a slice"),
 		}}}
 	}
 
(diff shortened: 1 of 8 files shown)


ingydotnet:
Built over #240

ccoVeille:
All name came from previous implementation.

Having o.All and WithAll is not meaningful.

I would use MultipleDocuments bool and WithMultipleDocuments

ccoVeille:
It's hard to review as only the last commit is relevant, while built on top of others.

Having another target branch could be considered, anyway.

Anyway, the migration guide should be updated maybe.

Also, the fact the yts/knownfailingtest is somehow rearranged in this commit makes me confused.

I cannot see if many things are now broken or not

ccoVeille:
I'm worried about one thing.

Consider this yaml

```yaml
- 10
- 25
- 42
- 101
```

And we use Load with a `any` or []any, I'm unsure about the expect behavior with All: true/false

Especially if we Dump them bacj

Wouldn't become

```yaml
10
---
25
---
42
---
101
```

Or the opposite, this used as a source, dumped as the first example

ingydotnet:
> All name came from previous implementation.
> 
> Having o.All and WithAll is not meaningful.
> 
> I would use MultipleDocuments bool and WithMultipleDocument

I'll change it to `WithAllDocuments`.

ingydotnet:
> I'm worried about one thing.
> 
> Consider this yaml
> 
> ```yaml
> - 10
> - 25
> - 42
> - 101
> ```
> 
> And we use Load with a `any` or []any, I'm unsure about the expect behavior with All: true/false
> 
> Especially if we Dump them bacj
> 
> Wouldn't become
> 
> ```yaml
> 10
> ---
> 25
> ---
> 42
> ---
> 101
> ```
> 
> Or the opposite, this used as a source, dumped as the first example

I think this is easily explainable and expected behavior when using WithAllDocuments.
Don't worry. Be happy! ;)

ingydotnet:
> It's hard to review as only the last commit is relevant, while built on top of others.
> 
> Having another target branch could be considered, anyway.
> 
> Anyway, the migration guide should be updated maybe.
> 
> Also, the fact the yts/knownfailingtest is somehow rearranged in this commit makes me confused.
> 
> I cannot see if many things are now broken or not

It's one commit now that #240 is merged.

ingydotnet:
@ccoVeille I fixed the known failing tests churn. Force pushed. No changes to it. Thanks for pointing it out.

loader.go
```diff
@@ -49,33 +62,101 @@ import (
...
+	var dummy any
+	err = l.Load(&dummy)
+	if err != io.EOF {
+		if err != nil {
+			// Some other error occurred
+			return err
+		}
+		// Successfully loaded a second document - this is an error in strict mode
+		return &TypeError{Errors: []*libyaml.ConstructError{{
+			Err: errors.New("yaml: expected single document, found multiple"),
```
colinjlacy, line 155: Could be worth adding another few words suggesting to consider using WithAllDocuments, e.g. "if loading multiple documents is required, use the `yaml.WithAllDocuments()` option"

colinjlacy (approved):

loader.go
ingydotnet: See #256

PR #246, opened 2026-01-12
Support escaped slash in strings

### LINKED ISSUE #245: bug: Error on optional string escapes, even though valid in JSON
The bug reported in the legacy go-yaml/yaml repo https://github.com/go-yaml/yaml/issues/797 had a PR to fix it (https://github.com/go-yaml/yaml/pull/862), but unfortunately that was never merged and thus didn't make it into this fork.

Here's the original text of the issue:

----

**Describe the bug**
According to the [JSON spec](https://datatracker.ietf.org/doc/html/rfc8259#page-8) (YAML is a superset of JSON), [YAML 1.2.2](https://spec.yaml.io/main/spec/1.2.2/#57-escaped-characters), and [YAML 1.3.0 (draft 2022-01-19)](https://spec.yaml.io/main/spec/1.3.0/#escaped-characters), `"\/"` is valid string syntax, equivalent to `"/"`.

However, when `"\/"` exists on the input to `go-yaml/yaml`, the escaped character is not recognized.

Versions of go-yaml: v2.4.0, both v2 and v3 branches
Operating system: mac, Go Playground

**Example code**
https://go.dev/play/p/UbP6fq9WPbd

```go
package main

import (
	"fmt"
	"strings"

	v2 "gopkg.in/yaml.v2"
	v3 "gopkg.in/yaml.v3"
)

var s = `"\/"`

func main() {
	var out interface{}
	if err := v2.Unmarshal([]byte(s), &out); err != nil {
		fmt.Printf("v2: error: %v\n", err)
	} else {
		fmt.Printf("v2: %#v\n", out)
	}

	reader := strings.NewReader(s)
	var node v3.Node
	decoder := v3.NewDecoder(reader)
	if err := decoder.Decode(&node); err != nil {
		fmt.Printf("v3: error: %v\n", err)
	} else {
		fmt.Printf("v3: %#v\n", node.Content[0].Value)
	}
}
```
**Actual behavior**
```
v2: error: yaml: found unknown escape character
v3: error: yaml: found unknown escape character
```
**Expected behavior**
```
v2: "/"
v3: "/"
```

hairyhenderson:
Fixes #245 by porting forward https://github.com/go-yaml/yaml/pull/862

Original text:

---

The JSON spec lists the `/` character as optionally-escapable, and so YAML v1.2.0 added support for the sequence `\/` to be unescaped to `/`.

This PR adds the minimum necessary support for this, and should solve the issues that a few people have already hit, as documented in https://github.com/go-yaml/yaml/issues/797.

Thanks!

### DIFF
--- a/internal/libyaml/scanner.go
+++ b/internal/libyaml/scanner.go
@@ -2231,6 +2231,10 @@ func (parser *Parser) scanFlowScalar(token *Token, single bool) error {
 					code_length = 4
 				case 'U':
 					code_length = 8
+				case '/':
+					// Support escaped slash ("\/"), as required by YAML 1.2.0+
+					// for strict JSON compatibility
+					s = append(s, '/')
 				default:
 					return formatScannerErrorContext("while scanning a quoted scalar", start_mark,
 						"found unknown escape character", parser.mark)
--- a/internal/libyaml/testdata/constructor.yaml
+++ b/internal/libyaml/testdata/constructor.yaml
@@ -79,6 +79,12 @@
     yaml: '"42"'
     want: '42'
 
+# https://github.com/yaml/go-yaml/issues/245
+- scalar-resolution:
+    name: Escaped slash in double-quoted string
+    yaml: '"\/"'
+    want: '/'
+
 - scalar-resolution:
     name: Mapping with numeric values
     yaml: |


decode_test.go
```diff
@@ -521,6 +521,15 @@ bar: def`,
 			B: "def", // value should be set using whole tag as a name, see issue: <https://github.com/yaml/go-yaml/issues/157>
 		},
 	},
+
+	// Issue #245 - escaped slash
```
ingydotnet: Can you put this test in internal/libyaml/testdata/loader.yaml instead?

ingydotnet: Can this comment be the full url to #245? `https://github.com/yaml/go-yaml/issues/245`

ingydotnet (changes requested):
#240 is merged. It changed a lot, so you need to resolve conflicts. 
Your commit msg isn't formatted to the repo rules. 
Amend to remove the `fix: ` part and it will be fine.
Tests are mostly defined as data now, so move the test as commented.
Thanks for the PR. It's a good one.

ingydotnet:
@hairyhenderson Do you have time or plans to make the required updates here?

hairyhenderson:
@ingydotnet apologies - been swamped. I'll have a look soon.

hairyhenderson:
OK I think I've addressed the comments - I ended up squashing and force-pushing as I was rebasing & amending the commit anyway...

ccoVeille:
Interesting. You are fixing 2 known issues from yaml test suite.

Could you adapt the known test failure file to remove the one that fails ?

You can test locally with `make test`

internal/libyaml/scanner.go
```diff
@@ -2218,6 +2218,10 @@ func (parser *Parser) scanFlowScalar(token *Token, single bool) error {
 					code_length = 4
 				case 'U':
 					code_length = 8
+				case '/':
+					// Support escaped slash ("\/"), as required by YAML 1.2.0+
+					// for strict JSON compatibility
+					s = append(s, '/')
```
ccoVeille, line 2237: I can see this code from an old PR previously opened on a go-yaml/yaml

Here I'm unsure if the parsing of this shouldn't be triggered by an option.

I mean without this it's not accepted.

So now, it is. But the escape if this character is only accepted with YAML spec 1.2

It should remain invalid if we consider the yaml spec 1.1

Or should we consider the lib should by default accept it, and consider adding pure YAML spec 1.1 and YAML spec 1.2 options.

What do you think here @ingydotnet ?

internal/libyaml/testdata/constructor.yaml
```diff
@@ -79,6 +79,12 @@
     yaml: '"42"'
     want: '42'
 
+# https://github.com/yaml/go-yaml/issues/245
+- scalar-resolution:
+    name: Escaped slash in double-quoted string
+    yaml: '"\/"'
+    want: '/'
```
ccoVeille, line 86: Shouldn't we add more tests? This sounds like an optimistic test that only tests what is expected ?

What other variations could we expect?

I'm unsure what to add here, but I raise the point just in case

internal/libyaml/scanner.go
```diff
@@ -2218,6 +2218,10 @@ func (parser *Parser) scanFlowScalar(token *Token, single bool) error {
 					code_length = 4
 				case 'U':
 					code_length = 8
+				case '/':
```
ccoVeille, line 2234: I found back this comment from @ingydotnet in the old issue

https://github.com/go-yaml/yaml/issues/797#issuecomment-1025907699

Could we check if all expected characters is supported ?

internal/libyaml/testdata/constructor.yaml
ccoVeille, line 86: Here a test is only added to the constructor. Don't we have other test files to update?

I mean the constructor is a low level one. I would expect to see code testing it with Load via the testdata suite in testdata/decode.yaml file

hairyhenderson:
Sorry for the slow response. I have very little spare time these days.

I'd appreciate any help to get this merged. Currently I'm using my fork and it's just fine for my needs, so my motivation to spend time on this is basically non-existent...

PR #254, opened 2026-01-17
Add option WithQuotePreference

ingydotnet:
Add WithQuotePreference option for quote style control

Previously, quote style was inconsistent: the representer and serializer
used double quotes for strings requiring quoting (like "on" in YAML 1.1,
or "null"), while the emitter's plain-not-allowed fallback used single
quotes (for leading/trailing whitespace, syntax characters, etc).

This commit introduces WithQuotePreference(QuoteStyle) with three values:

- QuoteSingle: Prefer single quotes everywhere (new v4 default)
  Makes v4 behavior fully consistent with single quotes throughout

- QuoteDouble: Prefer double quotes everywhere
  For users who prefer double quotes consistently

- QuoteLegacy: Preserve v2/v3 inconsistent behavior
  Double quotes in representer/serializer, single in emitter fallback
  Used by V2 and V3 presets for backward compatibility

The option is wired through the entire quote decision pipeline:
representer.stringv(), serializer.serialize(), and emitter's three
plain-not-allowed fallback cases. The emitter field was added to
make v4 fully consistent when using QuoteSingle or QuoteDouble.

### DIFF
--- a/internal/libyaml/options.go
+++ b/internal/libyaml/options.go
@@ -11,6 +11,7 @@ package libyaml
 
 import (
 	"errors"
+	"fmt"
 )
 
 // Options holds configuration for both loading and dumping YAML.
@@ -333,8 +334,13 @@ func WithFlowSimpleCollections(flow ...bool) Option {
 //   - QuoteLegacy: Legacy v2/v3 behavior (mixed quoting)
 func WithQuotePreference(style QuoteStyle) Option {
 	return func(o *Options) error {
-		o.QuotePreference = style
-		return nil
+		switch style {
+		case QuoteSingle, QuoteDouble, QuoteLegacy:
+			o.QuotePreference = style
+			return nil
+		default:
+			return fmt.Errorf("invalid QuoteStyle value: %d", style)
+		}
 	}
 }
 
--- a/internal/libyaml/representer.go
+++ b/internal/libyaml/representer.go
@@ -496,11 +496,7 @@ func (r *Representer) stringv(tag string, in reflect.Value) {
 	case canUsePlain:
 		style = PLAIN_SCALAR_STYLE
 	default:
-		if r.quotePreference == QuoteDouble || r.quotePreference == QuoteLegacy {
-			style = DOUBLE_QUOTED_SCALAR_STYLE
-		} else {
-			style = SINGLE_QUOTED_SCALAR_STYLE
-		}
+		style = r.quotePreference.ScalarStyle()
 	}
 	r.emitScalar(s, "", tag, style, nil, nil, nil, nil)
 }
--- a/internal/libyaml/serializer.go
+++ b/internal/libyaml/serializer.go
@@ -158,11 +158,7 @@ func (r *Representer) node(node *Node, tail string) {
 		case strings.Contains(value, "\n"):
 			style = LITERAL_SCALAR_STYLE
 		case forceQuoting:
-			if r.quotePreference == QuoteDouble || r.quotePreference == QuoteLegacy {
-				style = DOUBLE_QUOTED_SCALAR_STYLE
-			} else {
-				style = SINGLE_QUOTED_SCALAR_STYLE
-			}
+			style = r.quotePreference.ScalarStyle()
 		}
 
 		r.emitScalar(value, node.Anchor, tag, style, []byte(node.HeadComment), []byte(node.LineComment), []byte(node.FootComment), []byte(tail))
--- a/internal/libyaml/yaml.go
+++ b/internal/libyaml/yaml.go
@@ -72,6 +72,16 @@ const (
 	QuoteLegacy                   // Legacy behavior: double in representer, single in emitter.
 )
 
+// ScalarStyle returns the scalar style for this quote preference in the
+// representer/serializer context.
+// In this context, both QuoteDouble and QuoteLegacy use double quotes.
+func (q QuoteStyle) ScalarStyle() ScalarStyle {
+	if q == QuoteDouble || q == QuoteLegacy {
+		return DOUBLE_QUOTED_SCALAR_STYLE
+	}
+	return SINGLE_QUOTED_SCALAR_STYLE
+}
+
 type ErrorType int
 
 // Many bad things could happen with the parser and emitter.
--- a/yaml_test.go
+++ b/yaml_test.go
@@ -2227,8 +2227,16 @@ func runEncodeOptsTest(t *testing.T, tc map[string]any) {
 
 	// Parse options
 	var opts []yaml.Option
-	if optsMap, ok := tc["opts"].(map[string]any); ok {
-		if rq, ok := optsMap["required-quotes"].(string); ok {
+	if optsRaw, exists := tc["opts"]; exists {
+		optsMap, ok := optsRaw.(map[string]any)
+		if !ok {
+			t.Fatalf("opts must be a map, got %T", optsRaw)
+		}
+		if rqRaw, exists := optsMap["required-quotes"]; exists {
+			rq, ok := rqRaw.(string)
+			if !ok {
+				t.Fatalf("required-quotes must be a string, got %T", rqRaw)
+			}
 			switch rq {
 			case "single":
 				opts = append(opts, yaml.WithQuotePreference(yaml.QuoteSingle))


testdata/encode.yaml
```diff
@@ -720,3 +720,248 @@
...
+- encode-opts:
+    name: QuoteLegacy - dash string (single from emitter)
+    data:
+      v: '-'
+    type: map[string]string
+    opts:
+      required-quotes: legacy
+    want: |
+      v: '-'
+
```
ccoVeille, line 967: I would also, add tests about the one that doesn't require a quote


```yaml
- encode-opts: 
     name: QuoteSingle - not required 
     data: 
       a: hello
       b: 'hello'
       c: "hello"
     type: map[string]string 
     opts: 
       required-quotes: single 
     want: | 
       a: hello
       b: hello
       c: hello

- encode-opts: 
     name: QuoteDouble - not required 
     data: 
       a: hello
       b: 'hello'
       c: "hello"
     type: map[string]string 
     opts: 
       required-quotes: double 
     want: | 
       a: hello
       b: hello
       c: hello
```

I assume they won't be quoted.

I'm raising this because I'm a bit worried people may understand the option in something they could use to quote things

yaml_test.go
```diff
@@ -2218,6 +2219,63 @@ func runEncodeTest(t *testing.T, tc map[string]any) {
 	assert.Equal(t, want, string(output))
 }
 
+func runEncodeOptsTest(t *testing.T, tc map[string]any) {
+	t.Helper()
+
+	// Get type and create instance
+	// Note: "type" field in YAML is renamed to "output_type" during normalization
+	// to avoid conflict with the test type ("encode-opts")
+	typeName := tc["output_type"].(string)
```
ccoVeille:
```suggestion
	typeName, ok := tc["output_type"].(string)
```

Then handle OK as an error. Otherwise a poorly written tests would pass

yaml_test.go
```diff
@@ -2218,6 +2219,63 @@ func runEncodeTest(t *testing.T, tc map[string]any) {
...
+	data := tc["data"]
+
+	// Parse options
+	var opts []yaml.Option
+	if optsMap, ok := tc["opts"].(map[string]any); ok {
+		if rq, ok := optsMap["required-quotes"].(string); ok {
```
ccoVeille: This two OK should be checked and t.Fail when false, otherwise a test could be broken without noticing it

yaml_test.go
```diff
@@ -2218,6 +2219,63 @@ func runEncodeTest(t *testing.T, tc map[string]any) {
...
+	}
+
+	// Dump the target to YAML with options
+	output, err := yaml.Dump(targetPtr, opts...)
+	if err != nil {
+		t.Fatalf("Dump failed: %v", err)
+	}
+
+	// Compare with expected output
+	want := tc["want"].(string)
```
ccoVeille:
```suggestion
	want, ok := tc["want"].(string)
```

ccoVeille:
Good thing, added some comments

testdata/encode.yaml
colinjlacy: I agree, it's worth ensuring that quoting functionality matches expected behavior.

yaml.go
```diff
@@ -115,6 +118,9 @@ var (
 	// WithFlowSimpleCollections controls flow style for simple collections.
 	// See internal/libyaml.WithFlowSimpleCollections.
 	WithFlowSimpleCollections = libyaml.WithFlowSimpleCollections
+	// WithRequiredQuotes sets quote style when quoting is required.
+	// See internal/libyaml.WithRequiredQuotes.
+	WithRequiredQuotes = libyaml.WithRequiredQuotes
```
colinjlacy: It's worth adding an explicit documentation note here (or somewhere) that it will not add quotes when they are not required by the YAML spec. The way it's currently written, `when required` could be interpreted as "when required by the user"

ccoVeille: Yes, I have the same concern

ingydotnet: How about WithQuotePreference

I don't like the current name either.

internal/libyaml/emitter.go
```diff
@@ -1051,13 +1051,25 @@ func (emitter *Emitter) selectScalarStyle(event *Event) error {
 	if style == PLAIN_SCALAR_STYLE {
 		if emitter.flow_level > 0 && !emitter.scalar_data.flow_plain_allowed ||
 			emitter.flow_level == 0 && !emitter.scalar_data.block_plain_allowed {
-			style = SINGLE_QUOTED_SCALAR_STYLE
+			if emitter.requiredQuotes == QuoteDouble {
+				style = DOUBLE_QUOTED_SCALAR_STYLE
+			} else {
+				style = SINGLE_QUOTED_SCALAR_STYLE
+			}
```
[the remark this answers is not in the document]
ingydotnet: 5797f2c

testdata/encode.yaml
ingydotnet: c3debed

yaml_test.go
ingydotnet: 1fd1027

yaml_test.go
ingydotnet: 1fd1027

yaml_test.go
ingydotnet: 1fd1027

yaml.go
ingydotnet: 7c9016a

internal/libyaml/options.go
```diff
@@ -315,6 +316,28 @@ func WithFlowSimpleCollections(flow ...bool) Option {
...
+//   - Strings containing special YAML syntax characters
+//   - Empty strings in certain contexts
+//
+// Quote styles:
+//   - QuoteSingle: Use single quotes (v4 default)
+//   - QuoteDouble: Use double quotes
+//   - QuoteLegacy: Legacy v2/v3 behavior (mixed quoting)
+func WithQuotePreference(style QuoteStyle) Option {
+	return func(o *Options) error {
+		o.QuotePreference = style
```
ccoVeille: iota and type aren't restrictivs enough in Go

Someone could pass QuoteStyle(42) as argument.

I would add a switch here to return an error if the value is none of the one expected

yaml_test.go
```diff
@@ -2214,7 +2218,73 @@ func runEncodeTest(t *testing.T, tc map[string]any) {
...
+	typeName, ok := tc["output_type"].(string)
+	if !ok {
+		t.Fatalf("output_type must be a string")
+	}
+	data := tc["data"]
+
+	// Parse options
+	var opts []yaml.Option
+	if optsMap, ok := tc["opts"].(map[string]any); ok {
+		if rq, ok := optsMap["required-quotes"].(string); ok {
```
ccoVeille: I would return an error if required-quotes is something else than a string.

I feel like, someone passing true/false via the yaml data, would lead to a bool here. And then it would report no error, and people may think something works

Same for passing 2 as a value

yaml_test.go
```diff
@@ -2214,7 +2218,73 @@ func runEncodeTest(t *testing.T, tc map[string]any) {
...
+	// to avoid conflict with the test type ("encode-opts")
+	typeName, ok := tc["output_type"].(string)
+	if !ok {
+		t.Fatalf("output_type must be a string")
+	}
+	data := tc["data"]
+
+	// Parse options
+	var opts []yaml.Option
+	if optsMap, ok := tc["opts"].(map[string]any); ok {
```
ccoVeille: Sane for OK being false that should trigger a t.Fatal, even it's less likely to happen

ccoVeille (approved):
Minor remarks

internal/libyaml/options.go
ingydotnet: fixed

yaml_test.go
ingydotnet: fixed

yaml_test.go
ingydotnet: fixed

internal/libyaml/serializer.go
```diff
@@ -158,7 +158,11 @@ func (r *Representer) node(node *Node, tail string) {
 		case strings.Contains(value, "\n"):
 			style = LITERAL_SCALAR_STYLE
 		case forceQuoting:
-			style = DOUBLE_QUOTED_SCALAR_STYLE
+			if r.quotePreference == QuoteDouble || r.quotePreference == QuoteLegacy {
+				style = DOUBLE_QUOTED_SCALAR_STYLE
+			} else {
+				style = SINGLE_QUOTED_SCALAR_STYLE
+			}
 		}
```
ccoVeille: This could become a .ScalarStyle method on QuoteStyle, avoiding to duplicate code

yaml_test.go
```diff
@@ -2202,7 +2206,73 @@ func runEncodeTest(t *testing.T, tc map[string]any) {
...
+		if rq, ok := optsMap["required-quotes"].(string); ok {
```
ccoVeille: These two ok being false are not reported as a failure

internal/libyaml/options.go
ccoVeille: I might have misses something, but I don't see other values being reported as an error

colinjlacy (approved):
Once @ccoVeille 's comments are addressed, rest looks good

internal/libyaml/options.go
ingydotnet: Forgot to push fix. 78a222c

internal/libyaml/serializer.go
ingydotnet: 78a222c

yaml_test.go
ingydotnet: 78a222c

internal/libyaml/yaml.go
```diff
@@ -63,6 +63,25 @@ const (
 	CRLN_BREAK // Use CR LN for line breaks (DOS style).
 )
 
+type QuoteStyle int
+
+// Quote style types for required quoting.
+const (
+	QuoteSingle QuoteStyle = iota // Prefer single quotes when quoting is required.
+	QuoteDouble                   // Prefer double quotes when quoting is required.
+	QuoteLegacy                   // Legacy behavior: double in representer, single in emitter.
```
ccoVeille, line 72: There is something I don't get with this sentence

`double in representer, single in emitter.`

I don't see changes in emitter, and the code in emitter seems to also use `style = r.quotePreference.ScalarStyle()` but was previously using DOUBLE_QUOTED_SCALAR_STYLE

Maybe it wasn't the case in the v2/v3, but here I feel there should be something different in the way it's handled to satisfy the comment here

I mean if we had

```go
// ScalarStyle returns the scalar style for this quote preference in the
// serializer context.
// In this context, both QuoteDouble and QuoteLegacy use double quotes.
func (q QuoteStyle) ScalarStyle() ScalarStyle {
	if q == QuoteDouble || q == QuoteLegacy {
		return DOUBLE_QUOTED_SCALAR_STYLE
	}
	return SINGLE_QUOTED_SCALAR_STYLE
}

// ScalarStyleEmitter returns the scalar style for this quote preference in the
// representer/emitter context.
// In this context, both QuoteSingle and QuoteLegacy use single quotes.
func (q QuoteStyle) ScalarStyleEmitter() ScalarStyle {
	if q == QuoteSingle || q == QuoteLegacy {
		return SINGLE_QUOTED_SCALAR_STYLE
	}
	return DOUBLE_QUOTED_SCALAR_STYLE
}
```

Then `representer.go` should use ScalarStyleEmitter


See https://github.com/yaml/go-yaml/pull/254/changes#r2709977172



@ingydotnet

internal/libyaml/representer.go
```diff
@@ -493,7 +496,7 @@ func (r *Representer) stringv(tag string, in reflect.Value) {
 	case canUsePlain:
 		style = PLAIN_SCALAR_STYLE
 	default:
-		style = DOUBLE_QUOTED_SCALAR_STYLE
+		style = r.quotePreference.ScalarStyle()
```
ccoVeille, line 499: See https://github.com/yaml/go-yaml/pull/254/changes#r2709975986

I would have expected something like this


```suggestion
		style = r.quotePreference.ScalarStyleEmitter()
```

ccoVeille (commented):
I feel like there is a problem @ingydotnet 

I might be wrong but I prefer to raise the point

internal/libyaml/yaml.go
ccoVeille: Note I might be wrong, as it could be opposite

But right now, with code QuoteLegacy always lead to use DoubleQuote

PR #258, opened 2026-01-17
Remove deprecation markers and rename Legacy verbiage

ingydotnet:
This change removes all deprecation warnings from the classic API (Decoder, Encoder, Unmarshal, Marshal) and renames the internal LegacyOptions to DefaultOptions to better reflect its purpose.

I think it's too early to start deprecating things until we know how the community wants to use the APIs. 
I realize the new APIs fully cover the old ones, but it might be an easier transition to leave those as supported for now. 

Changes:
- Removed "Deprecated:" markers from Decoder, Encoder, and related methods in yaml.go
- Removed "Deprecated:" markers from Node.Decode and Node.Encode in internal/libyaml/node.go
- Renamed section header from "Deprecated APIs" to "Classic APIs" in yaml.go
- Renamed LegacyOptions to DefaultOptions in internal/libyaml/options.go
- Updated all references to use DefaultOptions consistently
- Updated documentation in docs/dump-load-api.md and docs/v3-to-v4-migration.md to remove deprecation language

The classic API is no longer marked as deprecated and will remain supported as a stable, simple interface alongside the newer Loader and Dumper APIs. The DefaultOptions variable now clearly indicates it provides default settings for APIs that don't accept explicit options.

All existing tests pass with no regressions.

### DIFF
--- a/internal/libyaml/options.go
+++ b/internal/libyaml/options.go
@@ -351,9 +351,8 @@ func ApplyOptions(opts ...Option) (*Options, error) {
 	return o, nil
 }
 
-// LegacyOptions holds the default options for legacy APIs like
-// Marshal/Unmarshal.
-var LegacyOptions = &Options{
+// DefaultOptions holds the default options for APIs that don't accept options.
+var DefaultOptions = &Options{
 	Indent:     4,
 	LineWidth:  -1,
 	Unicode:    true,
--- a/docs/dump-load-api.md
+++ b/docs/dump-load-api.md
@@ -8,9 +8,9 @@ from simple one-liners to advanced streaming with full configuration control.
 **For new code in v4+, prefer `Dump`/`Load` over `Marshal`/`Unmarshal` and
 `Encode`/`Decode`.**
 
-**IMPORTANT**: `Marshal`, `Unmarshal`, `Encoder`, and `Decoder` are deprecated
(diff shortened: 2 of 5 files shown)


internal/libyaml/options.go
```diff
@@ -351,9 +351,8 @@ func ApplyOptions(opts ...Option) (*Options, error) {
 	return o, nil
 }
 
-// LegacyOptions holds the default options for legacy APIs like
-// Marshal/Unmarshal.
-var LegacyOptions = &Options{
+// DefaultOptions holds the default options for APIs that don't accept options.
+var DefaultOptions = &Options{
```
ccoVeille, line 355: This let me think yaml.V3, yaml.V2 would be better if they were renamed to yaml.DefaultOptionsV3 and yaml.DefaultOptionsV2

It would be more readable and understandable when used

```go
yaml.Dump(foo, yaml.V2)
```

```go
yaml.Dump(foo, yaml.DefaultOptionsV2)
```

I'm also worried it could hit confusion if we provide YAML spec as option to enable/disable support of YAML 1.2 for example.

We should have something clearer than yaml.V3 may sound like a version of the spec (so confused with YAML spec 1.2) and not the lib

ingydotnet: Opened #263

PR #259, opened 2026-01-17
Add a detailed internals doc file with mermaid diagrams

ingydotnet:


docs/go-yaml-internals.md
```diff
@@ -0,0 +1,969 @@
...
+
+Info:
+- File: internal/libyaml/composer.go (320 lines)
+- Main Function: `func (c *Composer) Parse() *Node`
+- Input: `Event` stream from owned Parser (via internal `peek()`/`expect()`)
+- Output: `*Node` tree
+- Called From:
+  * Entry point `yaml.go / unmarshal / 664`
+  * Entry point `yaml.go / Decoder.Decode / 532`
+  * Entry point `constructor.go / Construct / 348`
```
ccoVeille: What are these numbers that pops out randomly in the text?

Here 664, 532, 348

If they are line number or the number of lines for each methods we shouldn't use them, as just one commit can mess up all of them

docs/go-yaml-internals.md
```diff
@@ -0,0 +1,969 @@
...
+Info:
+- File: internal/libyaml/resolver.go (170 lines)
+- Main Function: `func resolve(tag string, in string) (rtag string, out any)`
+- Input: Tag string + scalar value
+- Output: Resolved tag + typed Go value
+- Called From:
+  * Composer (`composer.go:182`)
+  * Constructor (`constructor.go:702`)
+  * Representer (`representer.go:477`)
+  * Serializer (`serializer.go:34`)
```
ccoVeille: Here is explicitly line number, it cannot be like this

If there is a need to spot code, you can use

https://pkg.go.dev/go.yaml.in/yaml/v4/internal/libyaml#Serializer link

That would work when a new tag will be provided

ccoVeille (commented):
I don't see a need for such detailed document, but why not.

My problem with such document is that is state what is true now. So any change in the code would make it obsolete, and would require to keep updating the documentation.

Based on average developer laziness, including me, I predict this document won't be maintained correctly.

For this reason, while I like the diagrams that I would like to keep, I would drastically reduce the level of detail in the markdown file.

There are AI out there, if someone doesn't understand the code, they can ask AI to explain it. And AI would use the code currently in the repository.

Maybe we can add a GHA or Copilot instruction to keep this document up to date 🤔

Another way to say it, is that I would prefer we share a prompt so user could generate such documentation, instead of writting its result in a file.

ingydotnet:
> I don't see a need for such detailed document, but why not.
> 
> My problem with such document is that is state what is true now. So any change in the code would make it obsolete, and would require to keep updating the documentation.
> 
> Based on average developer laziness, including me, I predict this document won't be maintained correctly.
> 
> There are AI out there, if someone doesn't understand the code, they can ask AI to explain it. And AI would use the code currently in the repository.

I worked for several hours with Claude back and forth to create this set of documents. 
I don't want to lose this work because it has the right amount of detail that is necessary for people hacking on this code to know.

I'll do a little bit more work and I think they should go in a separate directory. 
This level of detail is necessary for me to make further refactoring, which is necessary for it, making the proper plugin APIs. 
I also feel like it's necessary for other people making serious contributions to this project. 
From my point of view, these programs are very complex and it's not obvious how everything works. 
If it's hard for us, it must be really hard for newcomers.
That was one of the main inspirations for the CLI.
But even the CLI is limited in what it allows us to see. 

I will take on maintaining this document properly and as the code matures and the right things get put into the right place, the document will become simpler. 

I can look into better ways to reference places than line numbers, but I wouldn't worry about it too much. 
The line numbers are helpful for me now. 

I think moving some of `docs/*.md` under `docs/devel/` will be a good idea.

docs/go-yaml-internals.md
ingydotnet: I moved the dev docs into https://github.com/yaml/go-yaml/tree/go-yaml-internals-docs/docs/dev/
and replaced the line nums with links.

For now I'll use AI to update the docs as necessary. 
We can consider some kind of GHA helper later if it seems like a good idea.

docs/go-yaml-internals.md
ingydotnet: 8046b0c

docs/dev/resolver-problem.mmd
```diff
@@ -0,0 +1,25 @@
+---
+title: resolve Called From 4 Places
+---
+flowchart TB
+    subgraph LOAD ["LOAD STACK calls resolve()"]
+        direction LR
+        L_composer[Composer<br/>composer.go line 182<br/>Sets Node.Tag] --> L_resolve1[resolve - 1st call]
+        L_constructor[Constructor<br/>constructor.go line 702<br/>Gets typed value] --> L_resolve2[resolve - 2nd call<br/>DUPLICATE WORK]
```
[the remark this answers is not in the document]
ingydotnet: 2967208

ingydotnet:
@ccoVeille this is all ready now, With changes from your comments. 
Can you approve it?

PR #262, opened 2026-01-18
Refactor TypeError to LoadErrors and restore v3 TypeError for compatibility

ccoVeille:
This PR refactors the error handling for constructor conflicts, improving code clarity and maintaining compatibility with yaml/v3.

Summary of Changes

   This PR contains 4 commits that clean up and improve the TypeError implementation:

   - Remove TypeError.Unwrap method - Eliminates the Unwrap() method that was added in commit 08d6997 (PR #240). This method couldn't work for old Go 
   versions (<1.20) because Unwrap() []error is not supported there. Since TypeError.Is() and TypeError.As() methods were added in commit 6fae831 (PR #240) to provide compatibility for both old and new Go versions, the Unwrap() method became redundant.

   - Removes the Strings() method that was intended as a migration helper with #236 from []string to []error, which is no longer needed.

   - Rename TypeError to LoadErrors - Renames TypeError to LoadErrors to better reflect its actual purpose (multiple constructor 
   matches) rather than implying a type-related issue. This also prepares for better compatibility with yaml/v3.
   - Restore yaml/v3 TypeError - Restores the TypeError functionality with the signature from yaml/v3.
  https://github.com/yaml/go-yaml/pull/19 changed the signature of the TypeError causing breaking changes. We are now restoring the TypeError signature. LoadErrors struct should be used now, this explains why TypeError is now marked as deprecated.

### DIFF
--- a/docs/v3-to-v4-migration.md
+++ b/docs/v3-to-v4-migration.md
@@ -7,7 +7,6 @@ This guide will help you migrate your code from `go.yaml.in/yaml/v3`
 
 - [ ] Update import path
 - [ ] Optionally migrate to new API (Load/Dump, Loader/Dumper)
-- [ ] Handle TypeError.Errors type change (if you use it directly)
 - [ ] Adjust formatting expectations or use yaml.V3 preset
 - [ ] Update tests
 
@@ -85,55 +84,6 @@ encoder.Close()
 dumper := yaml.NewDumper(writer)
 err := dumper.Dump(&config)
 dumper.Close()
-```
-
-## Breaking Changes
-
-### TypeError.Errors Field Type
-
-This is the **only breaking change** in v4.
-
-**v3:**
-```go
-type TypeError struct {
-    Errors []string  // Simple error strings
-}
-```
-
-**v4:**
-```go
-type TypeError struct {
-    Errors []*UnmarshalError  // Structured errors with location info
-}
-
-type UnmarshalError struct {
-    Line    int
-    Column  int
-    Problem string
-    // ... other fields
-}
-```
-
-**Migration:**
-
-If you directly access `TypeError.Errors` expecting `[]string`,
-you'll need to update your code:
-
-```go
-// v3 code
-if typeErr, ok := err.(*yaml.TypeError); ok {
-    for _, errStr := range typeErr.Errors {
-        fmt.Println(errStr)
-    }
-}
-
-// v4 code
-if typeErr, ok := err.(*yaml.TypeError); ok {
-    for _, unmarshalErr := range typeErr.Errors {
-        fmt.Printf("Line %d, Column %d: %s\n",
-            unmarshalErr.Line, unmarshalErr.Column, unmarshalErr.Problem)
-    }
-}
 ```
 
 ## New Features in v4
@@ -281,8 +231,7 @@ data, err := yaml.Dump(&config, yaml.V3)
 1. Update import path
 2. Replace `Unmarshal` → `Load`, `Marshal` → `Dump`
 3. Replace `NewDecoder` → `NewLoader`, `NewEncoder` → `NewDumper`
-4. Update TypeError.Errors handling
-5. Test with v4 defaults or choose formatting explicitly
+4. Test with v4 defaults or choose formatting explicitly
 
 ### Strategy 3: Feature Adoption (Maximum Value)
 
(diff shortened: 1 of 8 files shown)


ccoVeille:
MultiConstructError is a suggestion as a name, I'm open to any suggestion

ingydotnet:
@ccoVeille It looks like we have conflicts here. Do you want to resolve these or do you want me to?

ccoVeille:
> Thanks. I like this overall.
> 
> I would ask you to change the name `MultiConstructError` to `LoadErrors`.
> 
> I assume `Multi` means plural.
> 
> `Construct` which I assume means from `Constructor` is a current internal state and might change.
> 
> It looks like it contains a set of (multiple) `LoadError` items, so `LoadErrors` is the natural choice.

I'll resolve when I'll rename things.

The reason I used singular is the Go idiom that Error struct ends with -Error singular.

https://go.dev/wiki/Errors#naming

But load Error exists... So here, LoadErrors could be used. I will have to think about it.

ccoVeille:
Also, I considered replacing

```go
type LoaderErrors struct {
	Errors []*libyaml.Error
}
```

By this

```go
type LoaderErrors []*libyaml.Error
```

But this could be done later in another PR, or never.

ingydotnet:
> The reason I used singular is the Go idiom that Error struct ends with -Error singular.
> 
> https://go.dev/wiki/Errors#naming

I had a look at that wiki page and I don't agree with your interpretation there. 
We are not defining an error type, we are defining something that is a collection of error type values.

ccoVeille:
> > The reason I used singular is the Go idiom that Error struct ends with -Error singular.
> > 
> > https://go.dev/wiki/Errors#naming
> 
> I had a look at that wiki page and I don't agree with your interpretation there. 
> We are not defining an error type, we are defining something that is a collection of error type values.

Well, let's consider existing TypeError struct. It's a collection of `string`, that represents the error messages.
But we are returning this collection as a Go `error`, so the struct has a `Error() string` method to satisfy signature, so it's an `error` and Go error TypeError satisfies the Go naming convention for error to end with `-Error` suffix.

Now, consider LoaderErrors that will contain a collection of error. We want to return it as an error, so we could use errors.As to catch the error. So it's an error, and Go error should satisfy Go naming convention for error, and so should end with `-Error` suffix.

But, I don't mind using LoadErrors here in this PR, we can still update anything later before the final 4.0.0. Here the idea is to fix the retro compatibility with YAML/v3

ingydotnet:
> Well, let's consider existing TypeError struct. It's a collection of `string`, that represents the error messages. But we are returning this collection as a Go `error`, so the struct has a `Error() string` method to satisfy signature, so it's an `error` and Go error TypeError satisfies the Go naming convention for error to end with `-Error` suffix.
> 
> Now, consider LoaderErrors that will contain a collection of error. We want to return it as an error, so we could use errors.As to catch the error. So it's an error, and Go error should satisfy Go naming convention for error, and so should end with `-Error` suffix.
> 
> But, I don't mind using LoadErrors here in this PR, we can still update anything later before the final 4.0.0. Here the idea is to fix the retro compatibility with YAML/v3

I love that you took this on for the ease of transition!

I just think that nobody (esp me) would realize that `MultiFooError` really meant `ThePluralFormOfFooError`.
It took me a long time to realize that.
Complicated by "multi" in YAML context almost always means a multi-document stream.

I think `LoadErrors` is perfect here, even if it is bending a Go rule.

ccoVeille:
I'm not very happy about the documentation because we have aliases report.

Considering the new LoaderErrors defined in libyaml as being 

```go
type LoadErrors struct {
   Errors []*ConstructError
}

type ConstructError struct {
   // whatever
}
```

And the aliases we have in  yaml package

```go
type LoadErrors = libyaml.LoadErrors
type LoadError   = libyaml.ConstructError
```

It's not obvious at yaml level that LoadErrors is a list of LoadError

```go
type LoadErrors struct {
   Errors []*LoadError
}
```

renaming libyaml.ConstructError to libyaml.LoadError could ease understanding

andig:
> LoadErrors better reflects that this error is raised when multiple YAML
nodes have issues, rather than indicating a type-related issue.

I beg to disagree with the reasoning. Looking at the readme, you can now get LoadErrors from Load but also from Unmarshal. Load per se is confusing, but I‘ve already commented that. According to reasoning, I would have expected these to be NodeErrors if anything.

PR #264, opened 2026-01-19
Fix godoc for re-exported blocks of comments

ccoVeille:
When re-exporting a block of comments, we added a comment just before the block to indicate that the block is re-exported.

But the godoc will be rendered as being the comment for the next declaration, so the godoc of the first declaration.

This leads to confusion when reading the documentation of the package. The "re-exported comment" shouldn't be visible in the publicly.

### DIFF
--- a/yaml.go
+++ b/yaml.go
@@ -264,6 +264,7 @@ type Unmarshaler interface {
 }
 
 // Re-export stream-related types
+
 type (
 	VersionDirective = libyaml.StreamVersionDirective
 	TagDirective     = libyaml.StreamTagDirective
@@ -271,6 +272,7 @@ type (
 )
 
 // Re-export encoding constants
+
 const (
 	EncodingAny     = libyaml.ANY_ENCODING
 	EncodingUTF8    = libyaml.UTF8_ENCODING
@@ -279,6 +281,7 @@ const (
 )
 
 // Re-export error types
+
 type (
 
 	// LoadError represents an error encountered while decoding a YAML document.
@@ -301,6 +304,7 @@ type (
 )
 
 // Re-export Kind constants
+
 const (
 	DocumentNode = libyaml.DocumentNode
 	SequenceNode = libyaml.SequenceNode
@@ -311,6 +315,7 @@ const (
 )
 
 // Re-export Style constants
+
 const (
 	TaggedStyle       = libyaml.TaggedStyle
 	DoubleQuotedStyle = libyaml.DoubleQuotedStyle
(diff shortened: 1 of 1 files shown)


ccoVeille:
I'm unsure this would be enough. As I'm removing random artifact in the documentation, but there are so much more to do to make it readable and usable by any Gopher.

The fact we re-export internal leads to have a poor documentation. I wish someone could easily see that a LoadError as an Error, Line and Co.

Also, it's not clear that a Load would return a LoadError that is a libyaml.ConstructError.

We should improve things. But the way cyclic dependencies led us to do this doesn't help.

A solution could be to us a package for errors, that can be imported then in yaml (root) and libyaml (internal), but that's not the way to do usually.

Another solution could be to add wrapper and not aliases, but it adds a cost just for improving documentation.

Maybe, I'm the only one worried about the internal being reexported and appearing with poor documentation

ccoVeille:
I worked on reviewing all the documentation and fix issues I found

internal/libyaml/emitter.go
```diff
@@ -72,8 +91,9 @@ type Emitter struct {
 
 	// Emitter stuff
 
-	canonical       bool       // If the output is in the canonical style?
-	BestIndent      int        // The number of indentation spaces.
+	canonical bool // If the output is in the canonical style?
+	// BestIndent is the number of spaces to use for indentation.
```
colinjlacy, line 95: nit: I'd suggest a blank line here to separate the formatting differences, and also to have a line-start before the exported member

internal/libyaml/parser.go
```diff
@@ -79,31 +79,53 @@ type SimpleKey struct {
 // ParserState represents the state of the parser.
 type ParserState int
 
-// Parser state constants define the different states the parser can be in.
+// Possible [ParserState] constants are the different states the parser can be in.
```
colinjlacy, line 82: "Possible"? I think it works below, but I would suggest leaving it off here.

colinjlacy (approved):
Really good stuff. A couple of thoughts, but no blockers.

internal/libyaml/parser.go
ccoVeille: Maybe it's because I'm French 😁

Maybe a better wording Available or simply "list of ParserState"

I'm open to suggestions.

Wording is not fun

colinjlacy: We can blame English. I'd suggest what you had: `[ParserState] constants list the different states the parser can be in.`

internal/libyaml/emitter.go
ccoVeille: I wanted to limit the diff. First to reduce the risk of conflicts, to keep git blame limited, and ease review

internal/libyaml/parser.go
ccoVeille: It should be better, please review and resolve.

I used a dedicated commit to ease the review.

ccoVeille:
Come on ... The rebase feature of GitHub web interface did a merge...

PR #266, opened 2026-01-22
Fix incorrect event formatting

inteon:
Some YTS EventComparisonTest tests were failing because the event formatting did not match.

Fixes:
- TestYAMLSuite/6BFJ/EventComparisonTest
- TestYAMLSuite/C4HZ/EventComparisonTest
- TestYAMLSuite/CN3R/EventComparisonTest
- TestYAMLSuite/EHF6/EventComparisonTest
- TestYAMLSuite/G4RS/EventComparisonTest
- TestYAMLSuite/X38W/EventComparisonTest

### DIFF
--- a/internal/libyaml/parser.go
+++ b/internal/libyaml/parser.go
@@ -1561,6 +1561,30 @@ func ParserGetEvents(in []byte) (string, error) {
 	return events.String(), nil
 }
 
+func escapeScalarValue(b []byte) string {
+	builder := strings.Builder{}
+	builder.Grow(len(b))
+	for _, c := range string(b) {
+		switch c {
+		case '\\':
+			builder.WriteString(`\\`)
+		case 0:
+			builder.WriteString(`\0`)
+		case '\b':
+			builder.WriteString(`\b`)
+		case '\n':
+			builder.WriteString(`\n`)
+		case '\r':
+			builder.WriteString(`\r`)
+		case '\t':
+			builder.WriteString(`\t`)
+		default:
+			builder.WriteRune(c)
+		}
+	}
+	return builder.String()
+}
+
 // formatEvent formats an event as a human-readable string for debugging
 // and testing purposes.
 func formatEvent(e *Event) string {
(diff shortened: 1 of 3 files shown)


internal/libyaml/parser.go
```diff
@@ -1280,6 +1280,30 @@ func ParserGetEvents(in []byte) (string, error) {
 	return events.String(), nil
 }
 
+func escaped(b []byte) string {
```
ccoVeille: Nit: I would prefer a meaningful name here

Naming is hard, I know.

I would prefer like whateverEscape with whatever being about what it's about.

internal/libyaml/parser.go
```diff
@@ -1323,16 +1347,13 @@ func formatEvent(e *Event) string {
...
-		// Escape special characters for consistent event output.
-		val := strings.NewReplacer(
-			`\`, `\\`,
-			"\n", `\n`,
-			"\t", `\t`,
-		).Replace(string(e.Value))
-		b.WriteString(val)
+		b.WriteString(escaped(e.Value))
 
 	case SEQUENCE_START_EVENT:
```
ccoVeille, line 1635: Is there a way to back up all this with YAML data tests that Ingy has set up?

Here it's not clear what the tests are about. I mean, you could fix more than the tests cover. You may change things that the tests don't cover. An existing tests known to be failing, could fail for a different reason with this PR.

I would prefer to double your changes with unit tests in the format we chose

inteon: see https://github.com/yaml/go-yaml/pull/266/commits/ddb3f31828d322d09b50798253dc4d8c0535c689

ccoVeille (approved):

internal/libyaml/parser.go
```diff
@@ -1561,6 +1561,30 @@ func ParserGetEvents(in []byte) (string, error) {
 	return events.String(), nil
 }
 
+func escapeScalarValue(b []byte) string {
+	builder := strings.Builder{}
+	builder.Grow(len(b))
+	for _, c := range string(b) {
```
ingydotnet, line 1567: This is a small subset of YAML escape chars. 
https://yaml.org/spec/1.2.2/#57-escaped-characters
Can we support them all?

ingydotnet (changes requested):

internal/libyaml/parser.go
inteon: does that mean we have to escape spaces too? https://yaml.org/spec/1.2.2/#rule-ns-esc-space

ccoVeille: https://yaml.org/spec/1.2.2/#rule-ns-esc-space

> Escaped ASCII space (x20) character. This is useful at the start or the end of a line to force a leading or trailing space to become part of the content.

There is something similar for tab

My analysis is that there is a need for something like this

```yaml
example: "This line ends with a space\x20
  \x20and this one starts with one."
```

Here you want 2 spaces, if the string was not using \x20 but spaces, they would be trimmed

PR #270, opened 2026-01-22
Add validation for incompatible UnmarshalYAML signatures

ccoVeille:
This commit introduces runtime detection and error reporting for UnmarshalYAML methods that don't match the current v4 interface.

Changes:
- Add ErrIncompatibleUnmarshaler error for clearer error messages
- Add callUnmarshalYAML() to validate method signatures via reflection
- Detect incompatible parameters, return types, and arity mismatches
- Export ObsoleteConstructor interface (was internal-only)
- Add comprehensive tests for invalid UnmarshalYAML signatures
- Provide helpful migration guidance in error messages

This helps users migrating from older YAML libraries identify and fix signature mismatches early with clear, actionable error messages.

### DIFF
--- a/internal/libyaml/constructor.go
+++ b/internal/libyaml/constructor.go
@@ -28,7 +28,10 @@ type constructor interface {
 	UnmarshalYAML(value *Node) error
 }
 
-type obsoleteConstructor interface {
+// ObsoleteConstructor interface was the one used by yaml v2 lib
+//
+// it is still supported for backward compatibility.
+type ObsoleteConstructor interface {
 	UnmarshalYAML(construct func(any) error) error
 }
 
@@ -396,7 +399,10 @@ func (c *Constructor) callConstructor(n *Node, u constructor) (good bool) {
 	}
 }
 
-func (c *Constructor) callObsoleteConstructor(n *Node, u obsoleteConstructor) (good bool) {
+// callObsoleteConstructor calls the go-yaml v2-style UnmarshalYAML method
+//
+// the obsolete UnmarshalYAML signature is: func(func(any) error) error
+func (c *Constructor) callObsoleteConstructor(n *Node, u ObsoleteConstructor) (good bool) {
 	terrlen := len(c.TypeErrors)
 	err := u.UnmarshalYAML(func(v any) (err error) {
 		defer handleErr(&err)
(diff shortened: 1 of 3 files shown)


ccoVeille:
This PR is partial as we need to do the same thing in constructor.go

wwqgtxx:
One suggestion is that this check shouldn't be enabled by default. In reality, there might be code that depends on things outside our control. If we force an error to be returned for these, it will prevent users from upgrading to V4. Another point is the potential performance issue you mentioned in your TODO.

Furthermore, to avoid similar issues in version V5, have you considered defining *yaml.Node as an interface? This is a complex issue and should be discussed in a separate issue.

ccoVeille:
I agree with everything you said @wwqgtxx  😁

I'm not fan of current interface for Unmarshal for the obvious reason we have raised here.

All of these should be options, even the v2 `Unmarshaller(func(any) error) error` should be controlled by an error.

I feel like we may have to deal with flags here maybe.

I feel the v4 shouldn't look for any of these.
The migration guide should provide explanation on how to deal with things like this.

But here, I'm unsure how to add this directly to the v4 branch.

I mean some people are not using the Unmarshaler. Some might without knowing it.

So we should consider things such as:
- having modes: panic, error, log, ignore
- having a tool to analyze code to report how to migrate

The thing that I'm worried about is the UnmarshalYAML(version.Node) I have no idea how to deal with it, I mean right now, the number of method is limited, while almost everything is accessible directly on the struct.

I'm curious to get everyone point of view here

wwqgtxx:
This raises a potentially worthwhile discussion: the `Unmarshaller(func(any) error) error` interface in v2 may still be the optimal choice for compatibility. While it's not the most powerful (it can't access Node's internal structure), it offers the best compatibility, working simultaneously across multiple versions and even multiple YAML libraries from different developers.

I'm not entirely sure why version V3 named it `Obsolete`, but I think for most users, without specific needs, this API should be the preferred choice over relying on `yaml.Node`.

In our own use case, aside from implementing `OrderedMap`, serialization and deserialization of all other custom types work perfectly with the v2 interface. Advanced users might still need the `yaml.Node`-based interface, but for maintenance and migration costs, we shouldn't inform users that the v2 interface is deprecated; instead, it should be the preferred choice under normal circumstances.

PR #272, opened 2026-01-28
Fix 4ABK YTS test

inteon:
`4ABK/in.yaml`
```yaml
{
unquoted : "separate",
http://foo.com,
omitted value:,
}
```

We were not yet checking for `":,", ":]" and ":}" inside a flow content.`.

### DIFF
--- a/internal/libyaml/scanner.go
+++ b/internal/libyaml/scanner.go
@@ -722,11 +722,10 @@ func isEndOfScalarInFlowContentChar(b []byte, i int) bool {
 	// "? "
 	case '?':
 		return isBlankOrZero(b, i+1)
-	// ": ", ":,", ":[", ":]", ":{" and ":}"
+	// ": ", ":,", ":]" and ":}"
 	case ':':
 		return b[i+1] == ' ' || b[i+1] == ',' ||
-			b[i+1] == '[' || b[i+1] == ']' ||
-			b[i+1] == '{' || b[i+1] == '}'
+			b[i+1] == ']' || b[i+1] == '}'
 	default:
 		return false
 	}
--- a/internal/libyaml/testdata/scanner.yaml
+++ b/internal/libyaml/testdata/scanner.yaml
@@ -304,34 +304,34 @@
 - scan-tokens-detailed:
     name: Flow mapping colon followed by comma, bracket, or brace
     yaml: |-
-        {a:,b:[c:],d:{e:}}
+        {a:,b: [c:],d: {e:}}
     want:
     - STREAM_START_TOKEN
     - FLOW_MAPPING_START_TOKEN
     - KEY_TOKEN
     - SCALAR_TOKEN:
-        value: "a"
+        value: a
     - VALUE_TOKEN
     - FLOW_ENTRY_TOKEN
     - KEY_TOKEN
     - SCALAR_TOKEN:
-        value: "b"
+        value: b
     - VALUE_TOKEN
     - FLOW_SEQUENCE_START_TOKEN
     - KEY_TOKEN
     - SCALAR_TOKEN:
-        value: "c"
+        value: c
     - VALUE_TOKEN
     - FLOW_SEQUENCE_END_TOKEN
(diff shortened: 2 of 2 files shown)


internal/libyaml/scanner.go
```diff
@@ -2791,13 +2791,24 @@ func (parser *Parser) scanPlainScalar(token *Token) error {
...
+						parser.buffer[parser.buffer_pos] == ']' ||
+						parser.buffer[parser.buffer_pos] == '{' ||
+						parser.buffer[parser.buffer_pos] == '}')) ||
+				(parser.flow_level > 0 &&
+					(parser.buffer[parser.buffer_pos] == ':') &&
+					(parser.buffer[parser.buffer_pos+1] == ',' ||
+						parser.buffer[parser.buffer_pos+1] == '[' ||
+						parser.buffer[parser.buffer_pos+1] == ']' ||
+						parser.buffer[parser.buffer_pos+1] == '{' ||
+						parser.buffer[parser.buffer_pos+1] == '}')) {
```
ccoVeille: This become unfortunately uneasy to read 

Can we add an helper such isWhateverChar (only an example of possible name), so we could document it and have clearer code.

I know the code was already like this, but you add so much lines (that are needed) to the function that it makes ir uneasy to read and thus maintain.

I'm a bit afraid of possible error if someone changes something later.

internal/libyaml/scanner.go
```diff
@@ -2791,13 +2791,24 @@ func (parser *Parser) scanPlainScalar(token *Token) error {
...
 				break
```
ccoVeille, line 2490: Also same remark as this

https://github.com/yaml/go-yaml/pull/266#discussion_r2737263404

If possible I would appreciate if we could back up the code changes with unit tests to the YAML data tests we added to the package

inteon: Done, added a testcase

internal/libyaml/scanner.go
ccoVeille: @inteon wgat about this?

inteon: I created a helper function called `isEndOfScalarInFlowContentChar`

ccoVeille (approved):
Thank you !

internal/libyaml/testdata/scanner.yaml
```diff
@@ -301,6 +301,42 @@
         value: "hello\nworld"
     - STREAM_END_TOKEN
 
+- scan-tokens-detailed:
+    name: Flow mapping colon followed by comma, bracket, or brace
+    yaml: |-
+        {a:,b:[c:],d:{e:}}
```
ingydotnet: At least according to https://play.yaml.com/?show=refparse,goyaml#e2E6LGI6W2M6XSxkOntlOn19cw==
This is invalid YAML.

`{a:,b: [c:],d: {e:}}` is valid: https://play.yaml.com/?show=refparse,goyaml#e2E6LGI6IFtjOl0sZDoge2U6fX0=

internal/libyaml/testdata/scanner.yaml
```diff
@@ -301,6 +301,42 @@
...
+    want:
+    - STREAM_START_TOKEN
+    - FLOW_MAPPING_START_TOKEN
+    - KEY_TOKEN
+    - SCALAR_TOKEN:
+        value: "a"
```
ingydotnet: Why are these values quoted?

inteon: I removed the quotes

internal/libyaml/testdata/scanner.yaml
inteon: Ok, dropped support for `:[` and `:{`: https://github.com/yaml/go-yaml/pull/272/changes/e7ca8278b52d12acc925123cc9bce39548fc7792

ccoVeille: Funny to read what could appear as sad emojis

PR #273, opened 2026-01-28
Fix HWV9 and QT73 YTS tests

inteon:
Makes these YTS tests pass:
- TestYAMLSuite/HWV9/EventComparisonTest
- TestYAMLSuite/HWV9/LoadTest
- TestYAMLSuite/QT73/EventComparisonTest
- TestYAMLSuite/QT73/LoadTest

Note: I had to add new JSONComparisonTest exceptions in yts/known-failing-tests because these tests are only executed when LoadTest succeeds. These failures were "hidden".

### DIFF
--- a/internal/libyaml/testdata/parser.yaml
+++ b/internal/libyaml/testdata/parser.yaml
@@ -337,6 +337,45 @@
     - DOCUMENT_END_EVENT
     - STREAM_END_EVENT
 
+- parse-events-detailed:
+    name: Leading document-end before explicit document
+    yaml: |
+      ...
+      ...
+      ---
+      key: value
+    want:
+    - STREAM_START_EVENT
+    - DOCUMENT_START_EVENT:
+        implicit: false
+    - MAPPING_START_EVENT
+    - SCALAR_EVENT:
+        value: key
+    - SCALAR_EVENT:
+        value: value
+    - MAPPING_END_EVENT
+    - DOCUMENT_END_EVENT
+    - STREAM_END_EVENT
+
+- parse-events-detailed:
+    name: Leading document-end before implicit document
+    yaml: |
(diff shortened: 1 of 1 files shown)


yts/known-failing-tests
```diff
@@ -122,8 +122,7 @@ TestYAMLSuite/G5U8/LoadTest
 TestYAMLSuite/H2RW/JSONComparisonTest
 TestYAMLSuite/HRE5/EventComparisonTest
 TestYAMLSuite/HRE5/LoadTest
-TestYAMLSuite/HWV9/EventComparisonTest
-TestYAMLSuite/HWV9/LoadTest
+TestYAMLSuite/HWV9/JSONComparisonTest
```
inteon, line 125: JSONComparisonTest is only executed when LoadTest succeeds, so this failure was "hidden"

ccoVeille: Could you add this valuable information to the commit message and PR description

internal/libyaml/parser.go
```diff
@@ -258,12 +258,10 @@ func (parser *Parser) parseDocumentStart(event *Event, implicit bool) error {
...
-		for token.Type == DOCUMENT_END_TOKEN {
-			parser.skipToken()
-			if err := parser.peekToken(&token); err != nil {
-				return err
-			}
+	for token.Type == DOCUMENT_END_TOKEN {
+		parser.skipToken()
+		if err := parser.peekToken(&token); err != nil {
+			return err
 		}
```
ccoVeille, line 469: Same remark as this

https://github.com/yaml/go-yaml/pull/266#discussion_r2737263404

If possible I would appreciate if we could back up the code changes with unit tests to the YAML data tests we added to the package

PR #275, opened 2026-01-29
Export ParserError and ScannerError types

ingydotnet:
Adds public API exports for ParserError, ScannerError, and Mark types to allow users to distinguish between parsing errors (YAML syntax issues) and construction errors (type mismatches) using errors.As().

This enables users migrating to v4 to handle different error phases appropriately without accessing internal packages.

Changes:
- Export ParserError, ScannerError, Mark as type aliases
- Add TestErrorTypeDistinction to verify error type detection
- Fully backward compatible - only adds new exports

### DIFF
--- a/yaml_test.go
+++ b/yaml_test.go
@@ -1101,6 +1101,50 @@ func TestLoadErrors_Unwrapping_Failures(t *testing.T) {
 	assert.ErrorIs(t, errTarget, errSentinel)
 }
 
+func TestErrorTypeDistinction(t *testing.T) {
+	// Parser error - flow sequence syntax issue
+	err := yaml.Unmarshal([]byte("foo: [unclosed"), &struct{}{})
+	if err == nil {
+		t.Fatal("expected error for syntax issue with unclosed bracket")
+	}
+	var pe yaml.ParserError
+	if !errors.As(err, &pe) {
+		t.Errorf("expected ParserError for flow sequence syntax issue, got: %T: %v", err, err)
+	} else if pe.Mark.Line != 1 {
+		t.Errorf("expected error on line 1, got line %d", pe.Mark.Line)
+	}
+
+	// Scanner error - invalid document structure
+	err = yaml.Unmarshal([]byte("a: 1\n=\nb: 2"), &struct{}{})
+	var se yaml.ScannerError
+	if !errors.As(err, &se) {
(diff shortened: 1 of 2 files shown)


yaml_test.go
```diff
@@ -1101,6 +1101,50 @@ func TestLoadErrors_Unwrapping_Failures(t *testing.T) {
...
+	err = yaml.Unmarshal([]byte("foo: [unclosed"), &struct{}{})
+	var pe2 yaml.ParserError
+	if errors.As(err, &pe2) {
+		// Explicitly declare mark as yaml.Mark to verify type is exported
+		//nolint:staticcheck // intentional type declaration to test export
+		var mark yaml.Mark = pe2.Mark
+		if mark.Line < 1 {
+			t.Error("Mark should have valid line number")
+		}
+	}
```
ccoVeille: You should fail if error cannot be converted to ParserError

ccoVeille: Can this be regrouped with the first ParserError at the top.

andig:
How would these new types relate to the previous `yaml.ParserError` and `yaml.UnmarshalError`?

yaml_test.go
```diff
@@ -1101,6 +1101,50 @@ func TestLoadErrors_Unwrapping_Failures(t *testing.T) {
 	assert.ErrorIs(t, errTarget, errSentinel)
 }
 
+func TestErrorTypeDistinction(t *testing.T) {
+	// Parser error - flow sequence syntax issue
+	err := yaml.Unmarshal([]byte("foo: [unclosed"), &struct{}{})
+	if err == nil {
+		t.Fatal("expected error for syntax issue with unclosed bracket")
+	}
+	var pe yaml.ParserError
```
andig, line 1110: Note that we used to have a *long* discussion on using `errors.As` on base or pointer receivers and agreed on pointer receivers before. Any reason to change that?

ingydotnet:
I'm going to close this now and work with the team on a new solution. 
I didn't fully understand the issues when I made this. 
I forgot about ParserError and UnmarshalError being added in https://github.com/yaml/go-yaml/pull/19
and didn't realize they got removed from the public API in https://github.com/yaml/go-yaml/pull/240
now that https://github.com/yaml/go-yaml/pull/279 is merged
and after a closer look at everything, I'd like to take pause to get a solid error solution in place that is clean and also takes account of reporting the stage and position info.

Stay tuned.

PR #277, opened 2026-01-29
Really Big Refactoring

ingydotnet:
coming soon...

### DIFF
--- a/node.go
+++ b/node.go
@@ -1,0 +1,51 @@
+package yaml
+
+import "go.yaml.in/yaml/v4/internal/libyaml"
+
+//-----------------------------------------------------------------------------
+// Node-related type aliases and constants
+//-----------------------------------------------------------------------------
+
+type (
+	// Node represents a YAML node in the document tree.
+	// See internal/libyaml.Node.
+	Node = libyaml.Node
+	// Kind identifies the type of a YAML node.
+	// See internal/libyaml.Kind.
+	Kind = libyaml.Kind
+	// Style controls the presentation of a YAML node.
+	// See internal/libyaml.Style.
+	Style = libyaml.Style
+	// Marshaler is implemented by types with custom YAML marshaling.
+	// See internal/libyaml.Marshaler.
+	Marshaler = libyaml.Marshaler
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -8,3 +8,4 @@
 /go-yaml
 /note/
 /*.yaml
+/*.md
```
ccoVeille, line 11: This one is not logic to me

docs/plugins.md
```diff
@@ -0,0 +1,597 @@
...
+Think of plugins as middleware for YAML.
+
+## Quick Example: Comment Preservation
+
+The simplest plugin use case is preserving comments:
+
+```go
+import (
+    "go.yaml.in/yaml/v4"
+    "go.yaml.in/yaml/v4/plugin/comment/v3"
```
ccoVeille, line 24: I would prefer we could use something like this


```suggestion
    "go.yaml.in/yaml/v4/plugin/commentv3"
```

I'm a bit afraid code importing the plugins will deal with having multiple v2/v3/v4 depending on the plugins one for the comment, another plugin

Any thoughts @colinjlacy

.gitignore
colinjlacy: Agreed. Was this to overcome Claude's indulgent use of `.md` files? If so, it would make for a good case for an `AGENTS.md` file instead.

docs/plugins.md
colinjlacy: I'm definitely not a fan of the `.../v4/.../v3` syntax. Reads as a bit contradictory. That said, I get the need for composability in preferences. In this case, I think @ingydotnet and I talked about something more like `/v4/plugin/comment/legacy`, but it doesn't solve the bigger question.

This may actually be a good opportunity for opening an issue to align on this and address in a downstream PR. I feel like this is a bigger conversation around user experience and real-world applicability. For example - am I overthinking the `.../v4/.../v3` syntax, and only two or three people will ever actually use it that way?

.gitignore
ingydotnet: Yes.

ingydotnet:
Going to try this another way. Closing.

PR #284, opened 2026-02-01
Consolidate stream metadata into Stream struct

ingydotnet:
Refactors the Node type to use a dedicated Stream struct for stream-specific metadata instead of individual fields. This improves the Node API by making stream-only fields explicit and reducing the number of fields that are conditionally valid.

Changes:
- Add Stream struct to hold Encoding, Version, and TagDirectives
- Replace Node.{Encoding,Version,TagDirectives} with Node.Stream
- Update Composer to populate Stream struct for StreamNodes
- Update CLI node formatter to access fields via Stream pointer
- Update tests and test data to reflect new structure
- Re-export Stream type from public yaml package

The Stream field is only non-nil when Kind == StreamNode, making the API clearer and the Node struct more efficient for non-stream nodes.

Clarify nested structure in stream.yaml docs

Update documentation to explicitly show that encoding, version, and tag-directives are nested under the stream: key, not direct properties of Stream nodes. Add colons to field names and "nested structure" label to address Copilot review feedback.

Add explanatory comment in loader_test.go clarifying that Stream is guaranteed non-nil for StreamNode (set by createStreamNode in composer.go), addressing a suppressed Copilot PR review comment.

### DIFF
--- a/docs/v3-to-v4-migration.md
+++ b/docs/v3-to-v4-migration.md
@@ -28,35 +28,6 @@ Update all import statements throughout your codebase.
 
 ## API Changes
 
-### Breaking Change: `Node` Stream Metadata Consolidated
-
-v4 removes the `Encoding`, `Version`, and `TagDirectives` fields from
-`Node` and replaces them with a single `Stream *Stream` field.
-The `Stream` pointer is non-nil only on `StreamNode` nodes.
-
-**v3:**
-```go
-if node.Kind == yaml.DocumentNode {
-    enc := node.Encoding
-    ver := node.Version       // *yaml.VersionDirective
-    tags := node.TagDirectives // []yaml.TagDirective
-}
-```
-
-**v4:**
-```go
-if node.Kind == yaml.StreamNode && node.Stream != nil {
-    enc := node.Stream.Encoding
-    ver := node.Stream.Version       // *yaml.VersionDirective
-    tags := node.Stream.TagDirectives // []yaml.TagDirective
-}
-```
-
-Note that in v4 stream metadata is carried by `StreamNode`, not
-`DocumentNode`.
-Use the `NewLoader` API with `WithStreamNodes()` to receive
-`StreamNode` values.
-
 ### Recommended: Use New API
 
 v4 introduces a cleaner API with better naming.
@@ -116,6 +87,36 @@ dumper.Close()
 ```
 
 ## New Features in v4
+
+### Stream Nodes
+
+v4 adds `StreamNode`, a new node kind that exposes stream-level
+metadata not available in v3: encoding, `%YAML` version directives,
+and `%TAG` directives.
+This is accessed via `Node.Stream`, which is non-nil only on
+`StreamNode` nodes.
+
+Enable stream nodes with `WithStreamNodes()`:
+
+```go
+loader := yaml.NewLoader(reader, yaml.WithStreamNodes())
+for {
+    var node yaml.Node
+    err := loader.Load(&node)
+    if errors.Is(err, io.EOF) {
+        break
+    }
+    if node.Kind == yaml.StreamNode && node.Stream != nil {
+        enc := node.Stream.Encoding
+        ver := node.Stream.Version        // *yaml.VersionDirective
+        tags := node.Stream.TagDirectives // []yaml.TagDirective
+    }
+}
+```
+
+With stream nodes enabled, the loader emits nodes in the pattern
+`[Stream, Doc, Stream, Doc, ..., Stream]` — one opening stream node
+per document boundary, plus a final closing stream node.
 
 ### Functional Options
 


internal/libyaml/node.go
```diff
@@ -107,6 +107,14 @@ type StreamTagDirective struct {
 	Prefix string
 }
 
+// Stream holds stream-level metadata for StreamNode.
+// This includes encoding, version directive, and tag directives.
+type Stream struct {
+	Encoding      Encoding
+	Version       *StreamVersionDirective
```
ccoVeille, line 114: Is there a need for this to be a pointer now the stream is a pointer in Node?

internal/libyaml/loader_test.go
```diff
@@ -109,22 +109,22 @@ func TestStreamNodeDirectives(t *testing.T) {
 
 	// First StreamNode should have encoding
 	assert.Equal(t, StreamNode, nodes[0].Kind)
-	// Encoding should be set (non-zero)
-	if nodes[0].Encoding == 0 {
+	// Stream is guaranteed non-nil for StreamNode (set by createStreamNode)
+	if nodes[0].Stream.Encoding == 0 {
```
ccoVeille: Because of the Murphy's law

You should check that Stream is not nil


```suggestion
	if nodes[0].Stream == nil ||nodes[0].Stream.Encoding == 0 {
```

internal/libyaml/loader_test.go
```diff
@@ -109,22 +109,22 @@ func TestStreamNodeDirectives(t *testing.T) {
...
-	if streamNode.Version != nil {
-		assert.Equal(t, 1, streamNode.Version.Major)
-		assert.Equal(t, 1, streamNode.Version.Minor)
+	if streamNode.Stream.Version != nil {
+		assert.Equal(t, 1, streamNode.Stream.Version.Major)
+		assert.Equal(t, 1, streamNode.Stream.Version.Minor)
 	}
 
-	if len(streamNode.TagDirectives) > 0 {
+	if len(streamNode.Stream.TagDirectives) > 0 {
```
ccoVeille: Same as

https://github.com/yaml/go-yaml/pull/284#discussion_r2758844709

internal/libyaml/loader_test.go
```diff
@@ -109,22 +109,22 @@ func TestStreamNodeDirectives(t *testing.T) {
...
 		t.Fatal("first stream node should have encoding set")
 	}
 
 	// Second node is the StreamNode before the document with directives
 	// Note: directives appear on the StreamNode BEFORE the document
 	streamNode := nodes[0]
-	if streamNode.Version != nil {
-		assert.Equal(t, 1, streamNode.Version.Major)
-		assert.Equal(t, 1, streamNode.Version.Minor)
+	if streamNode.Stream.Version != nil {
```
ccoVeille: Same as

https://github.com/yaml/go-yaml/pull/284#discussion_r2758844709

internal/libyaml/node.go
```diff
@@ -179,26 +187,15 @@ type Node struct {
...
-
-	// Version holds the YAML version directive (%YAML).
-	// Only valid for StreamNode.
-	Version *StreamVersionDirective
-
-	// TagDirectives holds the %TAG directives.
-	// Only valid for StreamNode.
-	TagDirectives []StreamTagDirective
+	// Stream holds stream metadata (non-nil only when Kind == StreamNode).
+	Stream *Stream
```
inteon, line 191: Could we use the new https://pkg.go.dev/unique#Handle here?

ccoVeille: Unique was added with Go 1.23 and the lib supports Go 1.18+ so it doesn't sound an option

internal/libyaml/node.go
ingydotnet: Yes -- Stream being a pointer tells you "this is a stream node,", but within a stream node, a version directive is still optional (most YAML streams don't have %YAML 1.1).
 The pointer on Version distinguishes "no version directive" from "has a version directive,".

internal/libyaml/loader_test.go
ingydotnet: 5b303d9

internal/libyaml/loader_test.go
ingydotnet: 5b303d9

internal/libyaml/loader_test.go
ingydotnet: 5b303d9

docs/v3-to-v4-migration.md
```diff
@@ -28,6 +28,35 @@ Update all import statements throughout your codebase.
 
 ## API Changes
 
+### Breaking Change: `Node` Stream Metadata Consolidated
+
+v4 removes the `Encoding`, `Version`, and `TagDirectives` fields from
+`Node` and replaces them with a single `Stream *Stream` field.
+The `Stream` pointer is non-nil only on `StreamNode` nodes.
+
```
ccoVeille: This somehow rebalance everything I feel

If we are OK to add such breaking changes to the v4 (and I'm OK with it), we should consider changing others we didn't want to change at first such as having Unmarshal being an alias of Decode and other about Dump.

And use v4 defauts, this means having v3 ones would require to use yaml.V3Setting

This would be addressed in another PR tha this one of course.

But this PR brings a real breaking change

ingydotnet: My understanding is that this is not a breaking change with V3. It's only another breaking change with the V4 work that we've done. I'll update the docs to make sure that's clear.

ingydotnet: Fixed in ac86f62

internal/libyaml/node.go
ccoVeille, line 191: If this changes

Consider this file

https://github.com/golang/vulndb/blob/884616812e63fd33ae1077d3ead6477173e5b8f5/internal/report/report.go#L104

If it migrates its import from gopkg.in/yaml.v3 to go.yaml.in/yaml/v4

This line will break as there is no Version field on the yaml.Node that will now be a v4 one.

docs/v3-to-v4-migration.md
ccoVeille: See https://github.com/yaml/go-yaml/pull/284#discussion_r2868034588

internal/libyaml/node.go
ccoVeille: I got it this fields only exists in the v4

So it's a breaking changes between the rc, it should be documented in the changelog.

But we are safe to merge it as we consider our -rc as breakable to for the v4.0.0 we didn't release yet

ccoVeille (approved):
I got it now. Its obky a v4 things.

So it breaks only for people using the v4.0.0-rc

It's acceptable

docs/v3-to-v4-migration.md
```diff
@@ -88,6 +88,36 @@ dumper.Close()
 
 ## New Features in v4
 
+### Stream Nodes
```
colinjlacy, line 91: A use case would probably be a good addition here, can be in a later PR

yaml.go
```diff
@@ -356,10 +356,9 @@ func OptsYAML(yamlStr string) (Option, error) {
 // Type and constant re-exports
 //-----------------------------------------------------------------------------
 
-// Stream-related types for advanced YAML processing
+// Re-export stream-related types
 type (
-	// VersionDirective represents a YAML %YAML version directive for stream
```
colinjlacy, line 361: Looks like getting rid of this comment might have been inadvertent?

testdata/stream.yaml
```diff
@@ -25,29 +25,33 @@
 #   want - List of expected nodes with kind and optional properties
 #
 # Node Format in 'want':
-#   kind     - Node kind: Stream or Document (required)
-#   encoding - Encoding name: UTF-8, UTF-16LE, UTF-16BE (for Stream nodes)
-#   version  - Version string like "1.1" (for Stream nodes)
-#   tag-directives - List of tag directive objects (for Stream nodes)
```
colinjlacy, line 31: Seeing these parenthetical comments - this was. good call 👍

PR #295, opened 2026-02-15
Don't return internal parse errors

### LINKED ISSUE #288: Cannot get source location of parse errors after `yaml.ParseError` was removed
go-yaml v4.0.0-rc.4 removed `yaml.ParseError`. Now `yaml.Unmarshal` returns `internal/libyaml.ScannerError` when parsing fails.

Users can no longer access the details of the error including the source location (line and column) because `internal/libyaml` package is internal.

```console
$ go build                                                                                                                    [Husky.local][02/08 23:32]
package github.com/rhysd/actionlint
        parse.go:12:2: use of internal package go.yaml.in/yaml/v4/internal/libyaml not allowed
```

I'd like to suggest to take one of the following solutions.

- Convert the internal `ScannerError` into `LoadError` and return `LoadErrors` when parsing fails
- Add a public type `ScannerError` to `go.yaml.in/yaml/v4` package and convert the internal error into the public error type
- Expose `internal/libyaml.ScannerError` as part of public API of `go.yaml.in/yaml/v4` package

I can work on this issue. If maintainers are busy, please let me know which one is the good direction.

rhysd:
Fix #288 

This PR fixes #288 by exposing `libyaml.ScannerError` and `libyaml.ParserError`. To access the details of the errors this patch also adds methods for the error positions.

### DIFF
--- a/internal/libyaml/errors.go
+++ b/internal/libyaml/errors.go
@@ -44,12 +44,42 @@ func (e ParserError) Error() string {
 	return MarkedYAMLError(e).Error()
 }
 
+// Line returns the line number (1-indexed) of the error position.
+func (e ParserError) Line() int {
+	return e.Mark.Line
+}
+
+// Column returns the column number (1-indexed) of the error position.
+func (e ParserError) Column() int {
+	return e.Mark.Column + 1
+}
+
+// Offset returns the byte offset of the error position.
+func (e ParserError) Offset() int {
+	return e.Mark.Index
+}
+
 // ScannerError represents an error that occurred during scanning.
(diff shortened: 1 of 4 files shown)


ccoVeille:
I don't think this is the direction we are going.

Your approach is good, but based on previous code we found out we merged too fast earlier with the addition of ParserError and ScannerError

I know @ingydotnet started working in another direction.

Said otherwise, your PR is OK, but I feel another PR that will provide a solution for the issue you are trying to solve

rhysd:
OK, then I close this and wait for the PR. I'll expect the next version will fix #288.

PR #297, opened 2026-02-17
Make the aliasing restrictions configurable

### LINKED ISSUE #296: Allow configuring aliasing levels in documents
The hardcoded aliasing limits are well...hard coded and there's no good reason for them to be.

If I have a particularly large and oddly shaped YAML document - or, as is the case I am programatically editing my YAML tree to insert aliases to take advantage of merge keys - then I should be able to supply options to the YAML decoder to set or disable the aliasing restrictions or modify the ratio or other things.

Crashing with a panic isn't a solution, and more importantly means a production application could sail over the limit and leave users with no configurables to work around an entirely expected situation.

As a very first step, just let a configuration option disable the check.

wrouesnel:
Resolves #296 

This is a fairly straight forward patch to remove the hard coded aliasing restrictions and instead allow them to be customized or disabled at runtime. In testing with a document with about 11,000 aliases I saw no noticeable performance but this document did trip over the aliasing limits. Since the source and nature of it is programmatically controlled, this is a use case where I'd like to disable the restrictions.

This patch takes the concept further by enabling all parameters to be configured, which is more likely to be what DoS style restrictions here are intended for.

### DIFF
--- a/internal/libyaml/options.go
+++ b/internal/libyaml/options.go
@@ -23,6 +23,11 @@ type Options struct {
 	StreamNodes    bool // Enable stream node emission
 	AllDocuments   bool // Load/Dump all documents in multi-document streams
 
+	AliasingRestrictionsNotEnforced bool    // Disable aliasing limits
+	AliasRatioRangeLow              int     // Low range of aliasing expansion limits
+	AliasRatioRangeHigh             int     // High range of aliasing expansion limits
+	AliasRatioRange                 float64 // Range over which allowed alias ratios are scaled
+
 	// Dumping options
 	Indent                int        // Indentation spaces (2-9)
 	CompactSeqIndent      bool       // Whether '- ' counts as indentation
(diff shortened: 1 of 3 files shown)


internal/libyaml/options.go
```diff
@@ -23,6 +23,11 @@ type Options struct {
 	StreamNodes    bool // Enable stream node emission
 	AllDocuments   bool // Load/Dump all documents in multi-document streams
 
+	AliasingRestrictionsNotEnforced bool    // Disable aliasing limits
+	AliasRatioRangeLow              int     // Low range of aliasing expansion limits
+	AliasRatioRangeHigh             int     // High range of aliasing expansion limits
+	AliasRatioRange                 float64 // Range over which allowed alias ratios are scaled
```
ccoVeille, line 29: While I understand why you do this. I'm a bit annoyed we have to create 4 fields for this.

Maybe a struct could be considered here

```go
type AliasRatio struct {
	NoRestrictionsNotEnforced bool    // Disable aliasing limits
	RangeLow              int     // Low range of aliasing expansion limits
	RangeHigh             int     // High range of aliasing expansion limits
	Range                 float64 // Range over which allowed alias ratios are scaled
}
```

And the struct could be used here in the option

Another way could be to use a [2]int for RangeLow and RangeHigh

These are suggestions. I'm not sure we need such complexity but it's what I think about when I see 4 fields being added.

What do you think @colinjlacy @ingydotnet ?

internal/libyaml/options.go
```diff
@@ -347,6 +352,38 @@ func WithQuotePreference(style QuoteStyle) Option {
...
+}
+
+// WithAliasRatioRange sets the range over which allowed alias ratios are scaled
+func WithAliasRatioRange(value float64) Option {
+	return func(o *Options) error {
+		o.AliasRatioRange = value
+		return nil
+	}
+}
+
```
ccoVeille, line 386: This is not tested, so it's a problem

wrouesnel: It is but there's a high risk at the moment the interface has to change heavily - i.e. the above

internal/libyaml/options.go
wrouesnel: The struct is probably the right choice internally, the bigger tension is what ranges should be allowed on aliasing parameters and how - i.e. should these even be `int` since negative values don't really make sense.

Strictly speaking maybe this is just one option - 

```go
func WithAliasingRestrictionFunction( value func(aliasCount int, constructCount int) bool ) {
	return func(o *Options) error {
		o.aliasingRestrictionFunction = value
		return nil
	}
}
```

Then the default behavior is just the existing function, disabling it would be supplying type def - i.e.

```go
func NoRestrictions(aliasCount int, constructCount int) bool {
	return false
}

// Used as something like - 
`WithAliasingRestrictionFunction(NoRestrictions)`
```

Which would substantially simplify the testing burden, an is the least opinionated option.

ccoVeille: Yes. This could be a way to implement it.

I'm curious about everyone feedback here

wrouesnel:
This PR should be superceded by #301 which implements the much simpler design (with an easier testing scheme as well).

ingydotnet:
Closing this as superseded by #301

Issue #299, opened 2026-02-21
Aliases don't work properly with TextUnmarshaler


WGH-:
Given

```go
type testStructAB_TextUnmarshaler struct {
	A simpleTextUnmarshaler
	B simpleTextUnmarshaler
}
```

YAML document

```yaml
a: &foo hello world
b: *foo
```

returns an error `cannot construct  into yaml_test.simpleTextUnmarshaler (TextUnmarshaler)`, but it should be correctly decoded into

```yaml
a:
    value: hello world
b:
    value: hello world
```

This is a problem as of current main (cc10f15886750ec884f374ec92ae1d187e68caba) and v4.0.0-rc.4.

ingydotnet:
@WGH- Fixed by #300?

Closing

PR #300, opened 2026-02-21
Reorder TextUnmarshaler check in Constructor

### LINKED ISSUE #293: [v4] cannot construct !!seq into types.Float64Slice (TextUnmarshaler)
Hi,

I tried the new v4 version, but it seems there is an issue compared to v3.

> cannot construct !!seq into types.Float64Slice (TextUnmarshaler)

Output: https://github.com/jkroepke/access-log-exporter/actions/runs/22008834236/job/63598445134#step:5:102

TypeDef: https://github.com/jkroepke/access-log-exporter/blob/e9d5c64998424c71247db1969143f60cfa5dec0e/internal/config/types/slice.go#L66

Code: https://github.com/jkroepke/access-log-exporter/blob/e9d5c64998424c71247db1969143f60cfa5dec0e/internal/config/types/slice.go#L81-L135

Test Code: https://github.com/jkroepke/access-log-exporter/blob/e9d5c64998424c71247db1969143f60cfa5dec0e/internal/config/types/slice_test.go#L54-L92

Please note: The behavior of JSON is different, it just works. The types have a dedicated `UnmarshalYAML` and should not care about the `UnmarshalText` interface

WGH-:
By moving `TextUnmarshaler` check after alias handling and `c.prepare` call, we resolve two issues:

* When the type implements `yaml.Unmarshaler`, `encoding.TextUnmarshaler` is irrelevant. In particular, we must not fail when decoding non-scalar nodes into such types (#293). `prepare` handles `yaml.Unmarshaler`, and skips the `TextUnmarshaler` check.

* We must resolve aliases before checking for the `TextUnmarshaler`. Otherwise, we would incorrectly fail unmarshaling into `TextUnmarshaler` types when YAML node is a alias (#299).

The only minor problem is that error messages are slightly changed: `prepare` removes pointer indirection, so instead of `*T` just `T` is now printed. This doesn't hurt understanding.

Fixes #293, #299

### DIFF
--- a/internal/libyaml/constructor.go
+++ b/internal/libyaml/constructor.go
@@ -89,6 +89,18 @@ func (c *Constructor) Construct(n *Node, out reflect.Value) (good bool) {
 		return true
 	}
 
+	switch n.Kind {
+	case DocumentNode:
+		return c.document(n, out)
+	case AliasNode:
+		return c.alias(n, out)
+	}
+
+	out, constructed, good := c.prepare(n, out)
+	if constructed {
+		return good
+	}
+
 	// When out type implements [encoding.TextUnmarshaler], ensure the node
 	// is a scalar. Otherwise, for example, constructing a YAML mapping
 	// into a struct having no exported fields, but implementing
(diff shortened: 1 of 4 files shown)


ccoVeille:
I'll review later, but can you fix the commit message to comply the contribution guide

ingydotnet (approved):

PR #301, opened 2026-02-23
Make the aliasing restrictions configurable (version 2)

### LINKED ISSUE #296: Allow configuring aliasing levels in documents

wrouesnel:
Resolves #296 and supercedes #297

This is a fairly straight forward patch to remove the hard coded aliasing restrictions and instead allow them to be customized or disabled at runtime. In testing with a document with about 11,000 aliases I saw no noticeable performance but this document did trip over the aliasing limits. Since the source and nature of it is programmatically controlled, this is a use case where I'd like to disable the restrictions.

This patch simplifies the #297 to allow users to fully replace the aliasing restrictions by replacing the function call which determines restrictions, and adds a built-in function for simply disabling the check. Tests are included to confirm that alternation of the implementation works.

### DIFF
--- a/internal/libyaml/constructor.go
+++ b/internal/libyaml/constructor.go
@@ -48,11 +48,12 @@ type Constructor struct {
 	stringMapType  reflect.Type
 	generalMapType reflect.Type
 
-	KnownFields    bool
-	UniqueKeys     bool
-	constructCount int
-	aliasCount     int
-	aliasDepth     int
+	KnownFields          bool
+	UniqueKeys           bool
+	AliasingExceededFunc AliasingRestrictionFunction
+	constructCount       int
+	aliasCount           int
+	aliasDepth           int
 
 	mergedFields map[any]bool
 }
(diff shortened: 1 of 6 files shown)


ingydotnet:
I like the premise here, but I would rather see this as one option for a Limits plugin.

wrouesnel:
@ingydotnet the problem with rolling it all together into one big limits blob is that it would force people to set a lot of separate parameters simultaneously or do a bunch nil handling IMO. i.e. for my use case here the issue was that under some circumstances only, it was valid to ignore aliasing because I knew why it was there (my code builds an alias heavy document as part of applying defaults at config unmarshal time).

That to me seems like one option, whereas if I was trying to undo traversal depth then that's a rather different one.

From a practical perspective also it would complicate the interface since the encoder/decoder context couldn't be built up by just concatenating `yaml.Option` fields

(EDIT: basically, if this became a common "Limits" option, it feels like you'd build it by having something that looks like yaml.Option but concatenates limits setting instead...which seems slightly redundant, or at least equivalent to just going something like

```go
  var limitOps := []yaml.Option{yaml.WithAliasingRestriction, yaml.WithMaxDepth .. }
```

Also: Also am I missing something about how CI tests run? This checks clean on my system.

ingydotnet:
> the problem with rolling it all together into one big limits blob is that it would force people to set a lot of separate parameters simultaneously or do a bunch nil handling IMO. i.e. for my use case here the issue was that under some circumstances only, it was valid to ignore aliasing because I knew why it was there (my code builds an alias heavy document as part of applying defaults at config unmarshal time).

I don't think the code for setting it as a Limits plugin value will look much different than setting it as an option.
I think you are making assumptions without having seen the plug-in system yet.

ingydotnet:
fwiw, I think a Limits is a great first plugin. Simple, useful and easy to understand. 
I'll strongly consider doing this one first. 
Hopefully in the next few days.

ingydotnet:
#313 was just merged and includes a Limit plugin. Let us know what you think.

ingydotnet:
https://gist.github.com/ingydotnet/03b2d3309093d112cfadcef333991615 
shows Go code examples using the limits plugin.

PR #307, opened 2026-03-06
Replace gopkg.in/check.v1 with the standard testing library

dims:
This ports #52 to the v2 branch, replacing gopkg.in/check.v1 with Go's standard testing package. 

-All 21 tests in decode_test.go, 11 in encode_test.go, and the limits tests in limit_test.go are converted from the 
`func (s *S) TestXxx(c *C)` receiver style to plain `func TestXxx(t *testing.T) `functions. 
- suite_test.go (the check.v1 runner bootstrap) is deleted
- gopkg.in/check.v1 dependency is removed from go.mod.


dims:
cc @ingydotnet @carloslima

decode_test.go
```diff
@@ -747,47 +749,62 @@ type inlineC struct {
...
-	err := yaml.Unmarshal([]byte(str), &t)
-	c.Assert(err, IsNil)
-	c.Assert(t, Equals, time.Date(2015, 2, 24, 18, 19, 39, 123456789, t.Location()))
-	c.Assert(t.In(time.UTC), Equals, time.Date(2015, 2, 24, 21, 19, 39, 123456789, time.UTC))
+	var tm time.Time
+	err := yaml.Unmarshal([]byte(str), &tm)
+	if err != nil {
+		t.Fatalf("Unmarshal() returned error: %v", err)
+	}
+	if !reflect.DeepEqual(tm, time.Date(2015, 2, 24, 18, 19, 39, 123456789, tm.Location())) {
```
dolmen: Don't use reflect.DeepEqual here. Instead, use tm.Equal, but also compare Location.

decode_test.go
```diff
@@ -747,47 +749,62 @@ type inlineC struct {
...
+		t.Fatalf("Unmarshal() returned\n%#v\nbut expected\n%#v", tm, time.Date(2015, 2, 24, 18, 19, 39, 123456789, tm.Location()))
+	}
+	if !reflect.DeepEqual(tm.In(time.UTC), time.Date(2015, 2, 24, 21, 19, 39, 123456789, time.UTC)) {
```
dolmen: No reflect.DeepEqual.

decode_test.go
```diff
@@ -882,27 +910,41 @@ var unmarshalErrorTests = []struct {
...
+			matched, mre := regexp.MatchString(item.error, err.Error())
+			if mre != nil {
+				t.Fatalf("regexp.MatchString() returned error: %v", mre)
+			}
+			if !matched {
+				t.Fatalf("Unmarshal() returned error %q, want match for %q. Partial unmarshal: %#v", err.Error(), item.error, value)
+			}
+		})
 	}
 }
```
dolmen, line 935: The behavior of the test is not respected. Copilot gives some hints, but is not perfect  either.

dolmen (changes requested):
I've stopped reviewing when I noticed that  TestUnmarshalErrors was not straighfully ported.

We don't need AI slop in the test suite.

decode_test.go
dims: Fixed.

decode_test.go
```diff
@@ -747,47 +748,63 @@ type inlineC struct {
...
-	c.Assert(t, Equals, time.Date(2015, 2, 24, 18, 19, 39, 123456789, t.Location()))
-	c.Assert(t.In(time.UTC), Equals, time.Date(2015, 2, 24, 21, 19, 39, 123456789, time.UTC))
+	var tm time.Time
+	err := yaml.Unmarshal([]byte(str), &tm)
+	if err != nil {
+		t.Fatalf("Unmarshal() returned error: %v", err)
+	}
+	expected := time.Date(2015, 2, 24, 18, 19, 39, 123456789, tm.Location())
+	if !tm.Equal(expected) || tm.Location().String() != expected.Location().String() {
+		t.Fatalf("Unmarshal() returned\n%v\nbut expected\n%v", tm, expected)
```
colinjlacy: Is there a reason for these line breaks in the failure output? Here and throughout? Seems unnecessary and also inconsistent.

decode_test.go
```diff
@@ -882,27 +909,54 @@ var unmarshalErrorTests = []struct {
...
+		if err == nil {
+			t.Fatalf("Unmarshal() expected error, got none. Partial unmarshal: %#v", value)
+		}
+		matched, mre := regexp.MatchString(item.error, err.Error())
+		if mre != nil {
+			t.Fatalf("regexp.MatchString() returned error: %v", mre)
+		}
+		if !matched {
+			t.Fatalf("Unmarshal() returned error %q, want match for %q. Partial unmarshal: %#v", err.Error(), item.error, value)
+		}
```
colinjlacy: Why introduce regex functions for simple string matching? Seems like it over-complicates things, especially when later in the PR we see direct string comparison using `==`.

decode_test.go
dims: Fixing

decode_test.go
dims: will remove for things that don't need it!

decode_test.go
dims: (The line breaks were to show clearly both the `tm` and `expected` on separate lines, so you can eyeball things. But i agree, will make it consistent by dropping the extra line breaks everywhere)

decode_test.go
dims: (was using regexp whenever there was `PanicMatches` or `ErrorMatches` in existing code. dropped all of it. there's only one regexp now for a `PanicMatches` scenario)

ingydotnet (approved):
lgtm

encode_test.go
```diff
@@ -486,31 +514,50 @@ var marshalErrorTests = []struct {
...
 	for _, item := range marshalErrorTests {
 		if item.panic != "" {
-			c.Assert(func() { yaml.Marshal(item.value) }, PanicMatches, item.panic)
+			func() {
+				defer func() {
+					r := recover()
+					if r == nil {
+						t.Fatalf("expected panic matching %q", item.panic)
+					}
+					if !regexp.MustCompile(item.panic).MatchString(fmt.Sprintf("%v", r)) {
```
colinjlacy: @dims Still seeing `regexp` string checks here

encode_test.go
```diff
@@ -486,31 +514,50 @@ var marshalErrorTests = []struct {
...
+						t.Fatalf("panic value %q does not match %q", fmt.Sprintf("%v", r), item.panic)
+					}
+				}()
+				yaml.Marshal(item.value)
+			}()
 		} else {
 			_, err := yaml.Marshal(item.value)
-			c.Assert(err, ErrorMatches, item.error)
+			if err == nil || !regexp.MustCompile(item.error).MatchString(err.Error()) {
```
colinjlacy: ...and here

encode_test.go
dims: Fixed!

encode_test.go
dims: fixed

Issue #274, opened 2026-01-28
Error in custom unmarshaller on latest tag `v4.0.0-rc.4`


moukoublen:
Between rc.3 and rc.4 it seems that a bug was introduced when a custom unmarshaller is used. 

Example:

```golang
package yaml_test

import (
	"testing"

	"go.yaml.in/yaml/v4"
	lib "go.yaml.in/yaml/v4/internal/libyaml"
)

type sliceCustomUnmarshaler struct {
	firstKind uint32
}

func (su *sliceCustomUnmarshaler) UnmarshalYAML(node *yaml.Node) error {
	su.firstKind = uint32(node.Kind)
	return nil
}

func TestCustomUnmarshalerNode(t *testing.T) {
	{
		var su sliceCustomUnmarshaler
		_ = yaml.Unmarshal([]byte(`["1", "2", "3"]`), &su)
		if uint32(lib.SequenceNode) != su.firstKind {
			t.Errorf("wanted SequenceNode got %d", su.firstKind)
		}
	}

	{
		var su sliceCustomUnmarshaler
		_ = yaml.Unmarshal([]byte("- name: one \n- name: two"), &su)
		if uint32(lib.SequenceNode) != su.firstKind {
			t.Errorf("wanted SequenceNode got %d", su.firstKind)
		}
	}

	{
		var su sliceCustomUnmarshaler
		_ = yaml.Unmarshal([]byte("1"), &su)
		if uint32(lib.ScalarNode) != su.firstKind {
			t.Errorf("wanted ScalarNode got %d", su.firstKind)
		}
	}
}
```

When a custom unmarshaller is used, and the whole document is being unmarshalled into it, the `node *yaml.Node` that is passed to `UnmarshalYAML` is of type `DocumentNode`, which is the root document if I am not mistaken, and this testcase fails.

In `v4.0.0-rc.3`, this testcase passes, which feels the correct behaviour.

 

It seems the behavior changed because of this: 
https://github.com/yaml/go-yaml/blob/a472fc4f55bb4038ddcd0da05cf1cdc1a83ccfed/yaml.go#L667-L676

ingydotnet:
@moukoublen Thanks for reporting this. We'll take a look. :)

ingydotnet:
@moukoublen Let me know if #276 works for you.

moukoublen:
Hello, yes I used that branch from pr and it [works](https://github.com/ifnotnil/pkgen/pull/46) 

[...]

PR #310, opened 2026-03-09
Simplify Unmarshal for Unmarshaler types

### LINKED ISSUE #306: UnmarshallYAML vs Load
Here is something strange according to me

Load is supposed to the "Unmarshaler for v4", and Unmarshall being kept for compatibility, using Load with v3 default.

It can be verified here

https://github.com/yaml/go-yaml/blob/d604f65988bfb7bbccebf41a7fe2145bd091f4ea/yaml.go#L574-L596

But this piece of code leads to something strange to me

Why do we have this instead of 

```go
func Unmarshal(in []byte, out any) (err error) {
	return Load(in, out, WithV3Defaults(), withFromLegacy())
}
```

I checked and nothing fails in our unit tests, why do we have such a code then ?

ingydotnet:
Unmarshal had a special code path for types implementing the Unmarshaler interface: it manually composed the node, unwrapped the DocumentNode, and called UnmarshalYAML directly — bypassing the normal Load/Constructor pipeline.

This was the original mechanism for supporting yaml.Unmarshaler before tryCallYAMLConstructor was added to the constructor. That later addition handles Unmarshaler via reflection at every level of construction, and is what Decoder.Decode has always relied on (with no special path of its own).

The special path was also bug-prone: it caused issue #274, where custom Unmarshalers received a DocumentNode instead of the content node, requiring an additional fix.

Simplify Unmarshal to a one-liner that delegates to Load, consistent with how Decoder.Decode works. All tests pass.

Closes #306

### DIFF
--- a/yaml_test.go
+++ b/yaml_test.go
@@ -143,9 +143,9 @@ func TestTextUnmarshalerWithYAMLUnmarshaler(t *testing.T) {
 	var target textUnmarshalerWithYAMLUnmarshaler
 	const input = `[foo, bar]`
 
-	// NOTE: can't use [yaml.Unmarshal] here: it has [yaml.Unmarshaler]
-	// shortcut, so it skips Constructor, and this test is a regression
-	// test for [libyaml.Constructor.Construct].
+	// NOTE: also verified with [yaml.Unmarshal] — no shortcut bypasses
+	// Constructor since PR #310, so this is a regression test for
+	// [libyaml.Constructor.Construct] via both paths.
 	err := yaml.NewDecoder(strings.NewReader(input)).Decode(&target)
 
 	assert.NoError(t, err)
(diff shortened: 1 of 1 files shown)


yaml.go
```diff
@@ -572,26 +571,6 @@ func (e *Encoder) Close() error {
...
-			}}}
-		}
-		// Unwrap DocumentNode to get the actual content
-		if node.Kind == libyaml.DocumentNode && len(node.Content) == 1 {
-			node = node.Content[0]
-		}
-		return u.UnmarshalYAML(node)
-	}
-	// Normal path
 	return Load(in, out, WithV3Defaults(), withFromLegacy())
```
ccoVeille, line 574: We need to add tests for this.

It's tested via the internal yaml test suite.

But we didn't see it there was a bug with the public API of the yaml package.

But then it would lead to replicate all tests at the public API level. I'm unsure if it's a good idea.

But maybe add some tests at the yaml API level could be considered.

Like the null value not calling the Unmarshaler, the fact Unmarshaler is called if it's OK

ingydotnet: 0a6843b

PR #311, opened 2026-03-10
Introduce LoadError for all YAML Loader stages

ingydotnet:
Replace scattered error types (`ParserError`, `ScannerError`, `ReaderError`, `MarkedYAMLError`) with a unified `LoadError` type that tracks which processing stage generated the error.

Key changes:
- Add `LoadError` struct with `Stage` field (reader, scanner, parser, composer, resolver, constructor)
- Add `Mark.shortString()` and `rangeString()` for position formatting
- Convert all error formatting functions to return `*LoadError`
- Update `LoadErrors` to use `[]*LoadError` with backward-compatible `As()`
- Export `Stage`, `Mark`, and `LoadError` types in public API
- Remove deprecated `MarkedYAMLError`, `ParserError`, `ScannerError`, `ReaderError` types (`ConstructError` retained for compatibility)
- Unexport `LoadError.Err` field; add `NewLoadError` constructor for external error creation
- Lowercase error prefix to `"go-yaml load error:"` per Go convention
- Include underlying cause in reader input error message
- Populate `Stage`/`Message` on nil-node stream error

Error format example:
```
go-yaml load error: exceeded max depth of 10000
  in scanner (while increasing indent level) at L1,C1-C20001
```

Position format: `L{line},C{col}` with compact ranges (`L2,C6-C7`).

### DIFF
--- a/cmd/go-yaml/event.go
+++ b/cmd/go-yaml/event.go
@@ -408,9 +408,9 @@ func convertLibyamlEvent(ev *libyaml.Event, profuse bool) *Event {
 
 	event := &Event{
 		StartLine:   ev.StartMark.Line,
-		StartColumn: ev.StartMark.Column - 1,
+		StartColumn: ev.StartMark.Column,
 		EndLine:     ev.EndMark.Line,
-		EndColumn:   ev.EndMark.Column - 1,
+		EndColumn:   ev.EndMark.Column,
 		HeadComment: string(ev.HeadComment),
 		LineComment: string(ev.LineComment),
 		FootComment: string(ev.FootComment),
--- a/cmd/go-yaml/parser.go
+++ b/cmd/go-yaml/parser.go
@@ -55,9 +55,9 @@ func (p *Parser) Next() (*Token, error) {
 
 	token := &Token{
 		StartLine:   yamlToken.StartMark.Line,
-		StartColumn: yamlToken.StartMark.Column - 1,
+		StartColumn: yamlToken.StartMark.Column,
 		EndLine:     yamlToken.EndMark.Line,
-		EndColumn:   yamlToken.EndMark.Column - 1,
+		EndColumn:   yamlToken.EndMark.Column,
 	}
 
 	switch yamlToken.Type {
--- a/internal/libyaml/composer.go
+++ b/internal/libyaml/composer.go
@@ -304,13 +304,13 @@ func (c *Composer) expect(e EventType) {
 	if c.event.Type == STREAM_END_EVENT {
 		Fail(formatComposerError(
 			"attempted to go past the end of stream; corrupted value?",
-			Mark{Line: c.event.StartMark.Line + 1, Column: c.event.StartMark.Column},
+			Mark{Line: c.event.StartMark.Line, Column: c.event.StartMark.Column},
 		))
 	}
 	if c.event.Type != e {
 		Fail(formatComposerError(
 			fmt.Sprintf("expected %s event but got %s", e, c.event.Type),
-			Mark{Line: c.event.StartMark.Line + 1, Column: c.event.StartMark.Column},
+			Mark{Line: c.event.StartMark.Line, Column: c.event.StartMark.Column},
 		))
 	}
 	c.event.Delete()
--- a/internal/libyaml/constructor.go
+++ b/internal/libyaml/constructor.go
@@ -113,7 +113,8 @@ func (c *Constructor) Construct(n *Node, out reflect.Value) (good bool) {
 	// encoding/json/v2.
 	if n.Kind != ScalarNode && isTextUnmarshaler(out) {
 		err := fmt.Errorf("cannot construct %s into %s (TextUnmarshaler)", shortTag(n.Tag), out.Type())
-		c.TypeErrors = append(c.TypeErrors, formatConstructorError(err, Mark{Line: n.Line, Column: n.Column}))
+		c.TypeErrors = append(c.TypeErrors,
+			formatConstructorError(err, Mark{Line: n.Line, Column: n.Column}))
 		return false
 	}
 
--- a/cmd/go-yaml/testdata/comments.yaml
+++ b/cmd/go-yaml/testdata/comments.yaml
@@ -14,15 +14,15 @@
     - {token: BLOCK-END}
     - {token: STREAM-END}
   TOKEN: |
-    - {token: STREAM-START, pos: '1:0'}
-    - {token: BLOCK-MAPPING-START, pos: '1:0'}
-    - {token: KEY, pos: '1:0'}
-    - {token: SCALAR, value: a, pos: '1:0-1'}
-    - {token: VALUE, pos: '1:1-2'}
+    - {token: STREAM-START, pos: '1:1'}
+    - {token: BLOCK-MAPPING-START, pos: '1:1'}
+    - {token: KEY, pos: '1:1'}
+    - {token: SCALAR, value: a, pos: '1:1-2'}
+    - {token: VALUE, pos: '1:2-3'}
     - {token: COMMENT, line: '#c', pos: '1:6-8'}
-    - {token: SCALAR, value: b, pos: '1:3-4'}
-    - {token: BLOCK-END, pos: '2:0'}
-    - {token: STREAM-END, pos: '2:0'}
+    - {token: SCALAR, value: b, pos: '1:4-5'}
+    - {token: BLOCK-END, pos: '2:1'}
+    - {token: STREAM-END, pos: '2:1'}
   event: |
     - {event: DOCUMENT-START}
     - {event: MAPPING-START}
--- a/cmd/go-yaml/testdata/document-implicit.yaml
+++ b/cmd/go-yaml/testdata/document-implicit.yaml
@@ -11,12 +11,12 @@
     - {event: MAPPING-END}
     - {event: DOCUMENT-END}
   EVENT: |
-    - {event: DOCUMENT-START, implicit: true, pos: '1:0'}
-    - {event: MAPPING-START, pos: '1:0'}
-    - {event: SCALAR, value: key, style: Plain, pos: '1:0-3'}
-    - {event: SCALAR, value: value, style: Plain, pos: '1:5-10'}
-    - {event: MAPPING-END, pos: '2:0'}
-    - {event: DOCUMENT-END, implicit: true, pos: '2:0'}
+    - {event: DOCUMENT-START, implicit: true, pos: '1:1'}
+    - {event: MAPPING-START, pos: '1:1'}
+    - {event: SCALAR, value: key, style: Plain, pos: '1:1-4'}
+    - {event: SCALAR, value: value, style: Plain, pos: '1:6-11'}
+    - {event: MAPPING-END, pos: '2:1'}
+    - {event: DOCUMENT-END, implicit: true, pos: '2:1'}
 
 - name: Single explicit document
   text: |
@@ -30,12 +30,12 @@
     - {event: MAPPING-END}
     - {event: DOCUMENT-END}
   EVENT: |
-    - {event: DOCUMENT-START, pos: '1:0-3'}
-    - {event: MAPPING-START, pos: '2:0'}
-    - {event: SCALAR, value: key, style: Plain, pos: '2:0-3'}
-    - {event: SCALAR, value: value, style: Plain, pos: '2:5-10'}
-    - {event: MAPPING-END, pos: '3:0'}
-    - {event: DOCUMENT-END, implicit: true, pos: '3:0'}
+    - {event: DOCUMENT-START, pos: '1:1-4'}
+    - {event: MAPPING-START, pos: '2:1'}
+    - {event: SCALAR, value: key, style: Plain, pos: '2:1-4'}
+    - {event: SCALAR, value: value, style: Plain, pos: '2:6-11'}
+    - {event: MAPPING-END, pos: '3:1'}
+    - {event: DOCUMENT-END, implicit: true, pos: '3:1'}
 
 - name: Implicit then explicit document
   text: |
@@ -50,12 +50,12 @@
     - {event: SCALAR, value: second}
     - {event: DOCUMENT-END}
   EVENT: |
-    - {event: DOCUMENT-START, implicit: true, pos: '1:0-5'}
-    - {event: SCALAR, value: first, style: Plain, pos: '1:0-5'}
-    - {event: DOCUMENT-END, implicit: true, pos: '2:0'}
-    - {event: DOCUMENT-START, pos: '2:0-3'}
-    - {event: SCALAR, value: second, style: Plain, pos: '3:0-6'}
-    - {event: DOCUMENT-END, implicit: true, pos: '4:0'}
+    - {event: DOCUMENT-START, implicit: true, pos: '1:1-6'}
+    - {event: SCALAR, value: first, style: Plain, pos: '1:1-6'}
+    - {event: DOCUMENT-END, implicit: true, pos: '2:1'}
+    - {event: DOCUMENT-START, pos: '2:1-4'}
+    - {event: SCALAR, value: second, style: Plain, pos: '3:1-7'}
+    - {event: DOCUMENT-END, implicit: true, pos: '4:1'}
 
 - name: Two explicit documents
   text: |
@@ -71,12 +71,12 @@
     - {event: SCALAR, value: second}
     - {event: DOCUMENT-END}
   EVENT: |
-    - {event: DOCUMENT-START, pos: '1:0-3'}
-    - {event: SCALAR, value: first, style: Plain, pos: '2:0-5'}
-    - {event: DOCUMENT-END, implicit: true, pos: '3:0'}
-    - {event: DOCUMENT-START, pos: '3:0-3'}
-    - {event: SCALAR, value: second, style: Plain, pos: '4:0-6'}
-    - {event: DOCUMENT-END, implicit: true, pos: '5:0'}
+    - {event: DOCUMENT-START, pos: '1:1-4'}
+    - {event: SCALAR, value: first, style: Plain, pos: '2:1-6'}
+    - {event: DOCUMENT-END, implicit: true, pos: '3:1'}
+    - {event: DOCUMENT-START, pos: '3:1-4'}
+    - {event: SCALAR, value: second, style: Plain, pos: '4:1-7'}
+    - {event: DOCUMENT-END, implicit: true, pos: '5:1'}
 
 - name: Document with explicit end marker
   text: |
@@ -91,12 +91,12 @@
     - {event: MAPPING-END}
     - {event: DOCUMENT-END, explicit: true}
   EVENT: |
-    - {event: DOCUMENT-START, pos: '1:0-3'}
-    - {event: MAPPING-START, pos: '2:0'}
-    - {event: SCALAR, value: key, style: Plain, pos: '2:0-3'}
-    - {event: SCALAR, value: value, style: Plain, pos: '2:5-10'}
-    - {event: MAPPING-END, pos: '3:0'}
-    - {event: DOCUMENT-END, pos: '3:0-3'}
+    - {event: DOCUMENT-START, pos: '1:1-4'}
+    - {event: MAPPING-START, pos: '2:1'}
+    - {event: SCALAR, value: key, style: Plain, pos: '2:1-4'}
+    - {event: SCALAR, value: value, style: Plain, pos: '2:6-11'}
+    - {event: MAPPING-END, pos: '3:1'}
+    - {event: DOCUMENT-END, pos: '3:1-4'}
 
 - name: Implicit doc with explicit end then explicit doc
   text: |
@@ -112,12 +112,12 @@
     - {event: SCALAR, value: second}
     - {event: DOCUMENT-END}
   EVENT: |
-    - {event: DOCUMENT-START, implicit: true, pos: '1:0-5'}
-    - {event: SCALAR, value: first, style: Plain, pos: '1:0-5'}
-    - {event: DOCUMENT-END, pos: '2:0-3'}
-    - {event: DOCUMENT-START, pos: '3:0-3'}
-    - {event: SCALAR, value: second, style: Plain, pos: '4:0-6'}
-    - {event: DOCUMENT-END, implicit: true, pos: '5:0'}
+    - {event: DOCUMENT-START, implicit: true, pos: '1:1-6'}
+    - {event: SCALAR, value: first, style: Plain, pos: '1:1-6'}
+    - {event: DOCUMENT-END, pos: '2:1-4'}
+    - {event: DOCUMENT-START, pos: '3:1-4'}
+    - {event: SCALAR, value: second, style: Plain, pos: '4:1-7'}
+    - {event: DOCUMENT-END, implicit: true, pos: '5:1'}
 
 - name: Three documents with mixed markers
   text: |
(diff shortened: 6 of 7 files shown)


internal/libyaml/composer.go
```diff
@@ -381,3 +384,24 @@ func Fail(err error) {
...
+		Message: message,
+	}
+}
+
+// formatComposerErrorContext creates a LoadError with both context and
+// problem information for composer-stage errors.
+func formatComposerErrorContext(context string, context_mark Mark, message string, mark Mark) *LoadError {
+	return &LoadError{
+		Stage:       ComposerStage,
+		ContextMark: context_mark,
```
ccoVeille: This is new code, we should avoid all non standard variable names


```suggestion
func formatComposerErrorContext(context string, contextMark Mark, message string, mark Mark) *LoadError {
	return &LoadError{
		Stage:       ComposerStage,
		ContextMark: contextMark,
```

internal/libyaml/errors.go
```diff
@@ -123,26 +144,39 @@ func (e *LoadErrors) Error() string {
...
+		if len(e.Errors) == 0 {
+			return false
+		}
+		first := e.Errors[0]
+		*t = &ConstructError{
+			Err:    first.err,
+			Line:   first.Mark.Line,
+			Column: first.Mark.Column,
+		}
+		return true
```
ccoVeille: This code should be considered to be temporary. Only people using rc might have ConstructorError

internal/libyaml/reader.go
```diff
@@ -425,21 +425,22 @@ func (parser *Parser) updateRawBuffer() error {
...
-		return ReaderError{
-			Offset: parser.offset,
-			Value:  -1,
-			Err:    fmt.Errorf("input error: %w", err),
+		return &LoadError{
+			Stage:   ReaderStage,
+			Message: fmt.Sprintf("input error: %v", err),
+			Mark:    Mark{Index: parser.offset},
+			err:     err,
 		}
```
ccoVeille, line 433: This remark is out of the scope of this PR, but I wonder if there is a need for a Message  field being a string in this struct. I would have used err field everywhere if I had written this today.

What do you think @colinjlacy ?

I mean the err field is available for Unwraping things such as io.EOF I think.

But having a struct about error handling that may have an error or not in an err field, and where message is a string that cannot be unwrapped sounds a bad thing

What do you guy think?

internal/libyaml/resolver.go
```diff
@@ -112,7 +113,10 @@ func resolve(tag string, in string) (rtag string, out any) {
 				}
 			}
 		}
-		failf("cannot construct %s `%s` as a %s", shortTag(rtag), in, shortTag(tag))
+		Fail(formatResolverError(
+			fmt.Sprintf("cannot construct %s `%s` as a %s", shortTag(rtag), in, shortTag(tag)),
+			Mark{},
+		))
```
ccoVeille, line 119: The pattern of panicking here should be reconsidered.

Fail call panic, so the error cannot be caught without a recover

What ro you think @colinjlacy

internal/libyaml/scanner.go
```diff
@@ -3393,28 +3393,29 @@ func (parser *Parser) readLine(s []byte) []byte {
...
 		Mark:    problem_mark,
 		Message: problem,
 	}
 }
 
-// formatScannerErrorContext creates a ScannerError with both context and
+// formatScannerErrorContext creates a LoadError with both context and
 // problem information, each with their own mark positions.
-func formatScannerErrorContext(context string, context_mark Mark, problem string, problem_mark Mark) error {
+func formatScannerErrorContext(context string, context_mark Mark, problem string, problem_mark Mark) *LoadError {
```
ccoVeille:
```suggestion
func formatScannerErrorContext(context string, contextMark Mark, problem string, problemMark Mark) *LoadError {
```

internal/libyaml/testdata/errors.yaml
```diff
@@ -10,305 +10,263 @@
...
+# EmitterError tests (dump stack)
 
-# EmitterError tests
 - emitter-error:
     name: Basic emitter error
     from:
-      message: could not write to output
-    want: 'yaml: could not write to output'
+      message: expected STREAM-START
+    want: 'yaml: expected STREAM-START'
```
ccoVeille, line 143: This message may sound cryptic to end user.

internal/libyaml/testdata/errors.yaml
```diff
@@ -10,305 +10,263 @@
...
-      message: cannot unmarshal string into int
-    want: 'line 10: cannot unmarshal string into int'
+      line: 5
+      message: type mismatch
+    want: 'line 5: type mismatch'
     also: unwrap
 
 - construct-error:
-    name: Construct error line 1
+    name: Construct error line 0
```
ccoVeille: There is a problem here. We cannot report error on line 0

node_test.go
```diff
@@ -282,7 +282,7 @@ func TestNodeZeroEncodeDecode(t *testing.T) {
 	_, err = yaml.Marshal(&n)
 	assert.ErrorMatches(t, "yaml: cannot represent node with unknown kind 0", err)
 	err = n.Load(&v)
-	assert.ErrorMatches(t, "yaml: cannot construct node with unknown kind 0", err)
+	assert.ErrorMatches(t, `go-yaml load error: cannot construct node with unknown kind 0\n  in constructor at L1`, err)
```
ccoVeille: This error message could be improved into something like 


```suggestion
	assert.ErrorMatches(t, `go-yaml load error: cannot construct node with unknown kind (kind=0)\n  in constructor at L1`, err)
```

Because the 0 sounds awkward otherwise

internal/libyaml/errors.go
```diff
@@ -123,26 +144,39 @@ func (e *LoadErrors) Error() string {
 	b.WriteString("yaml: construct errors:")
 	for _, err := range e.Errors {
 		b.WriteString("\n  ")
```
ccoVeille: Go spec discourages having new lines in error. Errors usually end in logs and newline  coukd lead to strange things, prevent easy grep

I notice it now because the error message in test suite sounded weird to me

@colinjlacy ?

Maybe reporting the first error is enough, andor reporting that multiple errors occurred.

internal/libyaml/errors.go
```diff
@@ -96,6 +115,8 @@ func (e WriterError) Unwrap() error {
 
 // ConstructError represents a single, non-fatal error that occurred during
 // the constructing of a YAML document into a Go value.
+//
+// Deprecated: Use LoadError instead.
```
ccoVeille:
```suggestion
// Deprecated: Use [LoadError] instead.
```

internal/libyaml/composer.go
ingydotnet: 9296135

internal/libyaml/errors.go
ingydotnet: Removed in c841449

node_test.go
ingydotnet: 2696ae4

internal/libyaml/testdata/errors.yaml
ingydotnet: ok for now

internal/libyaml/testdata/errors.yaml
ingydotnet: 3a04b4c

internal/libyaml/reader.go
colinjlacy: @ccoVeille @ingydotnet if we're calling it `Message` then I think it should be a string. I expect a `message` to be something I can quickly print or append to a log or downstream error. Maybe exporting the `err` field so that the consumer of this struct has access to the original `error` struct?

internal/libyaml/resolver.go
colinjlacy: Agreed, I'd like to see this error instead of panic.

internal/libyaml/errors.go
colinjlacy: I'm never a fan of line breaks in messages. In observability tools they can show up as different entries, thus breaking the full message across multiple records in e.g. Kibana. Kubernetes and ArgoCD separate multiple errors by `; `, and that's what I'd suggest we do here.

internal/libyaml/errors.go
```diff
@@ -12,61 +12,80 @@ import (
...
+	// Error chaining
+	err error // Underlying error (for Unwrap support)
+}
+
+// Error returns the error message with stage and position information.
+// Format: "go-yaml load error: <message>\n  in <stage> at L:C"
+// Or with context: "go-yaml load error: <message>\n  in <stage> (while <ctx>) at L:C-L:C"
+func (e *LoadError) Error() string {
+	if len(e.ContextMsg) > 0 {
+		return fmt.Sprintf("go-yaml load error: %s\n  in %s (%s) at %s",
```
colinjlacy: Seem my response to @ccoVeille 's comment - I'm never a fan of line breaks in these cases. Is there a reason we need this line break?

internal/libyaml/errors.go
```diff
@@ -12,61 +12,80 @@ import (
...
+			e.Message, e.Stage, e.ContextMsg, e.ContextMark.rangeString(e.Mark))
+	}
+	return fmt.Sprintf("go-yaml load error: %s\n  in %s at %s",
```
colinjlacy: Same here re: line break

internal/libyaml/resolver.go
ingydotnet: Are we just talking about this one specific place?

internal/libyaml/errors.go
ingydotnet: d28a2bd

internal/libyaml/errors.go
ingydotnet: d28a2bd

internal/libyaml/errors.go
ingydotnet: d28a2bd

internal/libyaml/resolver.go
colinjlacy: I didn't notice other places in the additions, but it's been a minute since I walked through this PR. If there are other places in the diff, then I'd suggest changing there too.

ccoVeille: I feel this must to be addressed, but I would do it in another PR.

The scope might be one too big.

internal/libyaml/errors.go
```diff
@@ -12,61 +12,80 @@ import (
...
+// Format: "go-yaml load error: <message>; in <stage> at L:C"
+// Or with context: "go-yaml load error: <message>; in <stage> (while <ctx>) at L:C-L:C"
+func (e *LoadError) Error() string {
+	if len(e.ContextMsg) > 0 {
+		return fmt.Sprintf("go-yaml load error: %s; in %s (%s) at %s",
+			e.Message, e.Stage, e.ContextMsg, e.ContextMark.rangeString(e.Mark))
+	}
+	return fmt.Sprintf("go-yaml load error: %s; in %s at %s",
+		e.Message, e.Stage, e.Mark.shortString())
 }
```
ccoVeille: While I find the semi colon OK for when multiple errors are present.

I wouldn't use it here. I'm not a native English speaker but the semi colon sounds strange in the flow of the text of error message.

I noticed this by reading the error messages in the tests.

Also the usage of semi colon in error message is pretty rare I think.


So, I would like to suggest this.

```suggestion
// Format: "go-yaml load error: <message>; in <stage> at L:C"
// Or with context: "go-yaml load error <message>; in <stage> (while <ctx>) at L:C-L:C"
func (e *LoadError) Error() string {
	if len(e.ContextMsg) > 0 {
		return fmt.Sprintf("go-yaml load error: %s in %s (%s) at %s",
			e.Message, e.Stage, e.ContextMsg, e.ContextMark.rangeString(e.Mark))
	}
	return fmt.Sprintf("go-yaml load error: %s in %s at %s",
		e.Message, e.Stage, e.Mark.shortString())
 }
```

ingydotnet: d28a2bd

ccoVeille: I don't see anything pushed since I reviewed, and the commit you mention is 2 days old.

So I'm unsure why this thread was marked as resolved.

I'm unsure what was changed. Because the semi colon is still there when there is only one error.

For these reasons I unresolved this thread

internal/libyaml/constructor.go
```diff
@@ -82,7 +82,10 @@ func (c *Constructor) Construct(n *Node, out reflect.Value) (good bool) {
 		c.aliasCount++
 	}
 	if c.aliasCount > 100 && c.constructCount > 1000 && float64(c.aliasCount)/float64(c.constructCount) > allowedAliasRatio(c.constructCount) {
-		failf("document contains excessive aliasing")
+		Fail(formatConstructorError(
+			fmt.Errorf("document contains excessive aliasing"),
+			Mark{Line: n.Line, Column: n.Column},
```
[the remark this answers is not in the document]
ccoVeille: These comments about 0/1 based index are pretty relevant

internal/libyaml/errors.go
ingydotnet: Sorry. I misunderstood you here.

ingydotnet: 0d6f604

internal/libyaml/constructor.go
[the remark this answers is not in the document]
ingydotnet: 0e2663f

internal/libyaml/composer.go
```diff
@@ -167,13 +167,10 @@ func (c *Composer) alias() *Node {
...
-		Fail(&ParserError{
-			Message: msg,
-			Mark: Mark{
-				Line:   n.Line,
-				Column: n.Column,
-			},
-		})
+		Fail(formatComposerError(msg, Mark{
+			Line:   n.Line,
+			Column: n.Column,
```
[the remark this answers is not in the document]
ingydotnet: 0e2663f

internal/libyaml/parser.go
```diff
@@ -1490,24 +1490,25 @@ func (parser *Parser) skipToken() {
...
-		Mark:    problem_mark,
-		Message: problem,
+func formatParserErrorContext(context string, contextMark Mark, problem string, problemMark Mark) *LoadError {
+	return &LoadError{
+		Stage:       ParserStage,
+		ContextMark: contextMark,
+		ContextMsg:  context,
+		Mark:        problemMark,
+		Message:     problem,
 	}
```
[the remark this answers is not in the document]
ingydotnet: 0e2663f

internal/libyaml/yaml.go
```diff
@@ -123,6 +123,43 @@ func (m Mark) String() string {
...
+// Returns "L{line}" when Column is 0, or "L{line},C{col}" otherwise.
+// Line and column values are stored 1-based and displayed as-is.
+func (m Mark) shortString() string {
+	if m.Line == 0 {
+		return "<unknown position>"
+	}
+	if m.Column != 0 {
+		return fmt.Sprintf("L%d,C%d", m.Line, m.Column+1)
+	}
+	return fmt.Sprintf("L%d", m.Line)
```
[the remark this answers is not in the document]
ingydotnet: 0e2663f

internal/libyaml/errors.go
```diff
@@ -12,61 +12,80 @@ import (
...
+// Format: "go-yaml load error in <stage> at L:C: <message>"
+// Or with context: "go-yaml load error in <stage> (<ctx>) at L:C-L:C: <message>"
+func (e *LoadError) Error() string {
+	if len(e.ContextMsg) > 0 {
+		return fmt.Sprintf("go-yaml load error in %s (%s) at %s: %s",
+			e.Stage, e.ContextMsg, e.ContextMark.rangeString(e.Mark), e.Message)
+	}
+	return fmt.Sprintf("go-yaml load error in %s at %s: %s",
+		e.Stage, e.Mark.shortString(), e.Message)
 }
```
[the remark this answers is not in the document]
ingydotnet: e5e775a

internal/libyaml/errors.go
```diff
@@ -94,46 +113,30 @@ func (e WriterError) Unwrap() error {
...
-		b.WriteString("\n  ")
-		b.WriteString(err.Error())
+	b.WriteString("yaml: construct errors: ")
+	for i, err := range e.Errors {
+		if i > 0 {
+			b.WriteString("; ")
+		}
+		b.WriteString(err.simpleError())
 	}
 	return b.String()
```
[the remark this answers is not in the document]
ingydotnet: fixed

internal/libyaml/errors.go
```diff
@@ -176,7 +179,7 @@ type TypeError struct {
 
 // Error returns a formatted error message listing all unmarshal errors.
 func (e *TypeError) Error() string {
```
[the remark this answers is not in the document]
ingydotnet: fixed

internal/libyaml/constructor.go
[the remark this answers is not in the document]
ccoVeille: I love the way you coded it,but I feel like the error messages shouldn't have been updated.

I mean the error message being C5 becoming C4 and all others

Was the index wrong?

I feel like the code dealing with the Error should do a kind of +1 no ?

Maybe I'm wrong, but at least I doubt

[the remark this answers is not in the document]
ingydotnet: I feel like in error messages we want to use 1-based numbering for lines and columns. 
I feel like also anytime we use the words line and column, this should be 1-based. 
But the Mark struct uses 0-based and that's legacy.
I don't know how much code out there depends on it, but imagine that's quite a bit might.
If people need their error messages formatted a specific way for legacy reasons, then we'll have a plugin API to allow them to make those adjustments.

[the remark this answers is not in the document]
ingydotnet: I had Claude do a deep dive on line and column numbers, both in the current branch and in v3: https://gist.github.com/ingydotnet/79e8c3888eaf7e5be3934382a98b51b3

It appears to me that the best course of action after reading all of this is that we can make Everything 1-based from the get go. Using `0` for line or for column will be a sentinel for unknown. 

This should be easy to understand and explain and should get rid of the current bugs. 
I'm going to try to do that change now.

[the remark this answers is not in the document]
ingydotnet: d0ee6b4c

Now we use 1-based integers for line and columns everywhere consistently!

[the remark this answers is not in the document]
ingydotnet: @ccoVeille Thanks for the push. It ended up being a big win imho.

ccoVeille (approved):

internal/libyaml/constructor.go
```diff
@@ -110,11 +113,7 @@ func (c *Constructor) Construct(n *Node, out reflect.Value) (good bool) {
 	// encoding/json/v2.
 	if n.Kind != ScalarNode && isTextUnmarshaler(out) {
 		err := fmt.Errorf("cannot construct %s into %s (TextUnmarshaler)", shortTag(n.Tag), out.Type())
-		c.TypeErrors = append(c.TypeErrors, &ConstructError{
-			Err:    err,
-			Line:   n.Line,
-			Column: n.Column,
-		})
+		c.TypeErrors = append(c.TypeErrors, formatConstructorError(err, Mark{Line: n.Line, Column: n.Column}))
```
colinjlacy: nit: this was more readable as multiple lines

internal/libyaml/errors.go
```diff
@@ -94,46 +113,30 @@ func (e WriterError) Unwrap() error {
...
-
-// Unwrap returns the underlying error.
-func (e *ConstructError) Unwrap() error {
-	return e.Err
-}
-
 // LoadErrors is returned when one or more fields cannot be properly decoded.
 type LoadErrors struct {
-	Errors []*ConstructError
+	Errors []*LoadError
```
colinjlacy, line 118: Looks like this was already here so could certainly be a problem for another day. But this strikes me as an odd construct for a plural form of an existing struct. I would have expected `LoadErrors` to be a slice of `LoadError`, not a struct with a single nested field.

colinjlacy (approved):
A couple of nits, but it looks good. Seeing how many structs were unified into a single error format shows the clear value of this change. Nicely done.

internal/libyaml/errors.go
ccoVeille: I hesitated when I converted -rc2 TypeError to this.

I also dislike the name of LoadErrors, as an error, is supposed to end with Error, no plural.

This can still be changed in another PR before v4.0.0 is released.

internal/libyaml/composer.go
```diff
@@ -305,10 +302,16 @@ func (c *Composer) expect(e EventType) {
...
+			"attempted to go past the end of stream; corrupted value?",
+			Mark{Line: c.event.StartMark.Line + 1, Column: c.event.StartMark.Column},
+		))
 	}
 	if c.event.Type != e {
-		c.fail(fmt.Errorf("expected %s event but got %s", e, c.event.Type))
+		Fail(formatComposerError(
+			fmt.Sprintf("expected %s event but got %s", e, c.event.Type),
+			Mark{Line: c.event.StartMark.Line + 1, Column: c.event.StartMark.Column},
+		))
```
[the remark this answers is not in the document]
ingydotnet: 1b9d64b

cmd/go-yaml/parser.go
```diff
@@ -54,10 +54,10 @@ func (p *Parser) Next() (*Token, error) {
...
 
 	token := &Token{
-		StartLine:   yamlToken.StartMark.Line + 1,
-		StartColumn: yamlToken.StartMark.Column,
-		EndLine:     yamlToken.EndMark.Line + 1,
-		EndColumn:   yamlToken.EndMark.Column,
+		StartLine:   yamlToken.StartMark.Line,
+		StartColumn: yamlToken.StartMark.Column - 1,
+		EndLine:     yamlToken.EndMark.Line,
+		EndColumn:   yamlToken.EndMark.Column - 1,
```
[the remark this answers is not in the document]
ingydotnet: 1b9d64b

cmd/go-yaml/parser.go
```diff
@@ -159,10 +159,10 @@ func (p *Parser) appendCommentTokenIfNotEmpty(value []byte, commentType string,
...
 			CommentType: commentType,
-			StartLine:   comment.StartMark.Line + 1,
-			StartColumn: comment.StartMark.Column + 1,
-			EndLine:     comment.EndMark.Line + 1,
-			EndColumn:   comment.EndMark.Column + 1,
+			StartLine:   comment.StartMark.Line,
+			StartColumn: comment.StartMark.Column,
+			EndLine:     comment.EndMark.Line,
+			EndColumn:   comment.EndMark.Column,
 		}
```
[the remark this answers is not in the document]
ingydotnet: 1b9d64b

cmd/go-yaml/event.go
```diff
@@ -407,10 +407,10 @@ func convertLibyamlEvent(ev *libyaml.Event, profuse bool) *Event {
...
 
 	event := &Event{
-		StartLine:   ev.StartMark.Line + 1, // libyaml uses 0-based lines
-		StartColumn: ev.StartMark.Column,
-		EndLine:     ev.EndMark.Line + 1,
-		EndColumn:   ev.EndMark.Column,
+		StartLine:   ev.StartMark.Line,
+		StartColumn: ev.StartMark.Column - 1,
+		EndLine:     ev.EndMark.Line,
+		EndColumn:   ev.EndMark.Column - 1,
```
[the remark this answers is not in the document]
ingydotnet: 1b9d64b

internal/libyaml/constructor.go
ingydotnet: 1b9d64b

PR #313, opened 2026-03-10
Plugin system with Limit plugin API

ingydotnet:
Add plugin system with Limit plugin API

Introduce a plugin architecture for extending YAML processing.
The first plugin, "limit", provides configurable depth and alias
expansion checks that replace the previously hardcoded limits.

New public API:
- LimitPlugin interface (CheckDepth, CheckAlias)
- WithPlugin() / WithoutPlugin() options
- limit.New() with DepthValue/DepthNone/DepthFunc and 
  AliasValue/AliasNone/AliasFunc options

Refactor internal depth/alias checking from inline logic in
the composer and constructor into pluggable callback functions
on the Options struct. Version presets (V2/V3/V4) default to
limit.New() so existing behavior is preserved.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -1,0 +1,128 @@
+# Plugin System
+
+The go-yaml v4 plugin system extends YAML processing with custom logic while
+maintaining performance, safety and backward compatibility.
+
+## Overview
+
+Plugins allow you to customize certain internal processing during loading and
+dumping.
+Plugin interfaces use public types and can be implemented by external packages.
+
+## Available Plugins
+
+### Limit Plugin
+
+The limit plugin controls the maximum nesting depth and alias expansion
+allowed during parsing.
+By default, go-yaml enforces conservative limits to prevent DoS attacks.
+Use the limit plugin to relax or tighten those limits.
+
+```go
+import "go.yaml.in/yaml/v4/plugin/limit"
+
+// Default limits (same as library defaults)
+loader := yaml.NewLoader(data, yaml.WithPlugin(limit.New()))
+
+// Disable alias checking (e.g. for documents with many programmatic aliases)
+loader := yaml.NewLoader(data, yaml.WithPlugin(limit.New(limit.AliasNone())))
+
+// Custom depth limit
+loader := yaml.NewLoader(data, yaml.WithPlugin(limit.New(limit.DepthValue(50))))
+```
+
+#### Limit Options
+
+| Option | Effect |
+|---|---|
+| `DepthValue(n)` | Max nesting depth (both flow and block) |
+| `DepthNone()` | Disable depth checking |
+| `DepthFunc(fn)` | Custom `func(depth int, ctx *yaml.DepthContext) error` |
+| `AliasValue(n)` | Max alias expansion count (simple threshold) |
+| `AliasNone()` | Disable alias ratio checking |
+| `AliasFunc(fn)` | Custom `func(aliasCount, constructCount int) error` |
+
+## Using Plugins
+
+### Basic Usage
+
+Register plugins with `WithPlugin()`:
+
+```go
+import (
+    "go.yaml.in/yaml/v4"
+    "go.yaml.in/yaml/v4/plugin/limit"
+)
+
+loader := yaml.NewLoader(data, yaml.WithPlugin(limit.New(limit.AliasNone())))
+var result any
+loader.Load(&result)
+```
+
+## Default Behavior
+
+Both bare `NewLoader(data)` and version presets (`WithV4Defaults()`, etc.)
+include default limits equivalent to `limit.New()`.
+
+## YAML Configuration
+
+Plugins can be configured from YAML strings using `OptsYAML`:
+
+```go
+opts, err := yaml.OptsYAML(`
+  plugin:
+    limit:
+      depth: 50
+      alias: 1000
+`)
+```
+
+Each plugin key maps to a configuration object. For the limit plugin:
+- `depth` (int) — max nesting depth; `null` disables depth checking
+- `alias` (int) — max alias count; `null` disables alias checking
+- Omitted keys keep defaults
+- Bare `limit:` (null value) uses all defaults
+
+```yaml
+# Disable depth checking, keep default alias limits
+plugin:
+  limit:
+    depth: null
+```
+
+## Third-Party Plugins
+
+To write a third-party plugin, implement the `yaml.LimitPlugin`
+interface:
+
+```go
+type LimitPlugin interface {
+    CheckDepth(depth int, ctx *DepthContext) error
+    CheckAlias(aliasCount, constructCount int) error
+}
+```
+
+Pass an instance to `yaml.WithPlugin()` — no import of
+`plugin/limit` is needed.
+
+Example:
+
+```go
+type StrictLimit struct{}
+
+func (s *StrictLimit) CheckDepth(depth int, ctx *yaml.DepthContext) error {
+    if depth > 100 {
+        return fmt.Errorf("depth %d exceeds policy limit of 100", depth)
+    }
+    return nil
+}
+
+func (s *StrictLimit) CheckAlias(aliasCount, constructCount int) error {
+    if aliasCount > 1000 {
+        return fmt.Errorf("alias count %d exceeds policy limit", aliasCount)
+    }
+    return nil
+}
+
+yaml.NewLoader(data, yaml.WithPlugin(&StrictLimit{}))
+```
--- a/internal/libyaml/composer.go
+++ b/internal/libyaml/composer.go
@@ -37,6 +37,9 @@ func NewComposer(b []byte, opts *Options) *Composer {
 		b = []byte{'\n'}
 	}
 	p.Parser.SetInputString(b)
+	if opts != nil {
+		p.Parser.depthCheck = opts.DepthCheck
+	}
 	return &p
 }
 
@@ -47,6 +50,9 @@ func NewComposerFromReader(r io.Reader, opts *Options) *Composer {
 		opts:   opts,
 	}
 	p.Parser.SetInputReader(r)
+	if opts != nil {
+		p.Parser.depthCheck = opts.DepthCheck
+	}
 	return &p
 }
 
--- a/internal/libyaml/constructor.go
+++ b/internal/libyaml/constructor.go
@@ -53,6 +53,7 @@ type Constructor struct {
 	constructCount int
 	aliasCount     int
 	aliasDepth     int
+	aliasCheck     func(aliasCount, constructCount int) error
 
 	mergedFields map[any]bool
 }
@@ -66,6 +67,7 @@ func NewConstructor(opts *Options) *Constructor {
 		KnownFields:    opts.KnownFields,
 		UniqueKeys:     opts.UniqueKeys,
 		aliases:        make(map[*Node]bool),
+		aliasCheck:     opts.AliasCheck,
 	}
 }
 
@@ -81,11 +83,10 @@ func (c *Constructor) Construct(n *Node, out reflect.Value) (good bool) {
 	if c.aliasDepth > 0 {
 		c.aliasCount++
 	}
-	if c.aliasCount > 100 && c.constructCount > 1000 && float64(c.aliasCount)/float64(c.constructCount) > allowedAliasRatio(c.constructCount) {
-		Fail(formatConstructorError(
-			fmt.Errorf("document contains excessive aliasing"),
-			Mark{Line: n.Line, Column: n.Column},
-		))
+	if c.aliasCheck != nil {
+		if err := c.aliasCheck(c.aliasCount, c.constructCount); err != nil {
+			Fail(formatConstructorError(err, Mark{Line: n.Line, Column: n.Column}))
+		}
 	}
 	if out.Type() == nodeType {
 		out.Set(reflect.ValueOf(n).Elem())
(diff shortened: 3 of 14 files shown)


docs/plugins.md
```diff
@@ -0,0 +1,124 @@
...
+Register plugins with `WithPlugin()`:
+
+```go
+import (
+    "go.yaml.in/yaml/v4"
+    "go.yaml.in/yaml/v4/plugin/limits"
+)
+
+loader := yaml.NewLoader(data, yaml.WithPlugin(limits.New(limits.AliasNone())))
+var result interface{}
```
ccoVeille: This would be better, as we target Go 1.18+

```suggestion
var result any
```

docs/plugins.md
```diff
@@ -0,0 +1,124 @@
...
+Both bare `NewLoader(data)` and version presets (`WithV4Defaults()`, etc.)
+include default limits equivalent to `limits.New()`.
+
+## YAML Configuration
+
+Plugins can be configured from YAML strings using `OptsYAML`:
+
+```go
+opts, err := yaml.OptsYAML(`
+  plugin:
```
ccoVeille, line 73: Shouldn't it be `plugins` ? As it's an object with keys

docs/plugins.md
```diff
@@ -0,0 +1,124 @@
...
+- `depth` (int) — max nesting depth; `null` disables depth checking
+- `alias` (int) — max alias count; `null` disables alias checking
+- Omitted keys keep defaults
+- Bare `limits:` (null value) uses all defaults
+
+```yaml
+# Disable depth checking, keep default alias limits
+plugin:
+  limits:
+    depth:
```
ccoVeille: This seems weird.

Having 0, -1 would be more logic. Maybe an explicit null here. People are not use to `depth: ` meaning nil value like having `~` or `null`

plugin/limit/plugin.go
```diff
@@ -0,0 +1,187 @@
...
+	return func(p *Plugin) {
+		p.depthDisabled = true
+		p.depthLimit = nil
+		p.depthFn = nil
+	}
+}
+
+// DepthFunc sets a custom depth check function.
+func DepthFunc(fn func(depth int, ctx *libyaml.DepthContext) error) Option {
+	return func(p *Plugin) {
```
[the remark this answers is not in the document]
ccoVeille: That's true

testdata/plugin.yaml
```diff
@@ -0,0 +1,107 @@
+# Copyright 2025 The go-yaml Project Contributors
```
ccoVeille: I would use 2026 for new files

yaml.go
```diff
@@ -254,6 +259,64 @@ func Options(opts ...Option) Option {
...
+// WithoutPlugin resets plugins of the specified kinds to their defaults.
+//
+// This can be used to override plugin settings from previous options.
+// Valid kinds: "limits"
+//
+// Example:
+//
+//	// Reset limits to library defaults
+//	loader := yaml.NewLoader(data, yaml.WithoutPlugin("limits"))
+func WithoutPlugin(kinds ...string) Option {
```
ccoVeille: I would use another name than WithoutPlugin

I mean if it resets the anchor limit to their default value, there are still limits for the number of anchors then. So I wouldn't say Without if still present

Maybe 


```suggestion
func ResetPlugin(kinds ...string) Option {
```

docs/plugins.md
```diff
@@ -0,0 +1,124 @@
...
+loader.Load(&result)
+```
+
+### Resetting to Defaults
+
+Use `WithoutPlugin()` to reset a plugin kind to library defaults:
+
+```go
+// Reset limits to defaults (overrides any previous WithPlugin)
+loader := yaml.NewLoader(data, yaml.WithoutPlugin("limits"))
```
colinjlacy: I find this confusing. If it's a `NewLoader`, then wouldn't it already be using defaults? How would previous `WithPlugin` functions have been called?

docs/plugins.md
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func (s *StrictLimits) CheckAlias(aliasCount, constructCount int) error {
+    if aliasCount > 1000 {
+        return fmt.Errorf("alias count %d exceeds policy limit", aliasCount)
+    }
+    return nil
+}
+
+yaml.NewLoader(data, yaml.WithPlugin(&StrictLimits{}))
```
colinjlacy: It's not clear how this works. Is there an interface that should be documented here?

internal/libyaml/constructor.go
```diff
@@ -54,6 +54,8 @@ type Constructor struct {
 	aliasCount     int
 	aliasDepth     int
 
+	aliasCheck func(aliasCount, constructCount int) error
```
colinjlacy: Totally nit-picking here, but is there a reason for the blank lines?

colinjlacy: Feel free to ignore. I just...can't not notice the blank lines

internal/libyaml/options.go
```diff
@@ -35,13 +33,58 @@ type Options struct {
...
 	// Private options (not exported, used internally)
 	FromLegacy bool // Indicates legacy Unmarshal()/Decoder path (check Unmarshaler, allow trailing content)
 }
 
 // Option allows configuring YAML loading and dumping operations.
 type Option func(*Options) error
 
+// DepthContext holds context about a nesting depth check.
+type DepthContext struct {
+	Kind string // "flow" or "block"
```
colinjlacy: Should this be an enum?

plugin/limits/plugin.go
```diff
@@ -0,0 +1,187 @@
...
+		return nil
+	}
+	if p.aliasLimit != nil {
+		if aliasCount > *p.aliasLimit {
+			return fmt.Errorf("exceeded max alias count of %d", *p.aliasLimit)
+		}
+		return nil
+	}
+	return libyaml.DefaultAliasCheck(aliasCount, constructCount)
+}
```
colinjlacy: As far as file formatting goes, it's weird to separate this from `CheckDepth` and have `NewFromYAML` sitting in between the two.

plugin.go
colinjlacy, line 1: Do we want this file sitting at the root of the project? Seems like there should be a folder with additional files and exported interfaces as new plugins are added.

plugin_test.go
colinjlacy, line 1: Same here, strikes me as something that should be in a folder, not at the root.

yaml.go
```diff
@@ -348,6 +421,30 @@ func OptsYAML(yamlStr string) (Option, error) {
 		}
 	}
 
+	if cfg.Plugin != nil {
+		for name, val := range *cfg.Plugin {
```
colinjlacy: Looks like we're starting with strictly limited plugin support. Will this be loosened in the future?

docs/plugins.md
ingydotnet: I'm not a fan of plurals. It's the `plugin` section of the config.

plugin.go
ingydotnet: `plugin.go` at the root defines the yaml.LimitsPlugin interface, which is intentionally part of the root yaml package's public API. 
Keeping it here means users write `yaml.LimitsPlugin` and `yaml.WithPlugin(...)` without any extra import. Moving it to a subfolder would require users to import a separate package just to reference the interface type, which would be an unnecessary friction for the common case of implementing a custom plugin.

yaml.go
ingydotnet: I'm not certain what you mean here. We define the plugin APIs. People can't just make up new ones. Keep hooking right third party plugins that work with our APIs. But plugins each hook deep into the go-yaml system. So that's what we define.

ingydotnet: If you mean, will we be adding more plugin APIs...
Yes, of course. Dozens of them.

internal/libyaml/composer.go
```diff
@@ -37,6 +37,9 @@ func NewComposer(b []byte, opts *Options) *Composer {
 		b = []byte{'\n'}
 	}
 	p.Parser.SetInputString(b)
+	if opts != nil {
+		p.Parser.depthCheck = opts.DepthCheck
+	}
```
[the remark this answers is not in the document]
ingydotnet: 8d8368c

internal/libyaml/composer.go
```diff
@@ -47,6 +50,9 @@ func NewComposerFromReader(r io.Reader, opts *Options) *Composer {
 		opts:   opts,
 	}
 	p.Parser.SetInputReader(r)
+	if opts != nil {
+		p.Parser.depthCheck = opts.DepthCheck
+	}
```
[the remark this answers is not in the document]
ingydotnet: 8d8368c

plugin/limit/plugin.go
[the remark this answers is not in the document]
ingydotnet: 5f64fcd

docs/plugins.md
ingydotnet: 9f244e9

docs/plugins.md
ingydotnet: 9f244e9

yaml.go
ingydotnet: c621c05

docs/plugins.md
ingydotnet: c621c05

testdata/plugin.yaml
ingydotnet: b547dbb

docs/plugins.md
ingydotnet: I don't see a need for plural keys with collection values

internal/libyaml/constructor.go
ingydotnet: f19a653

internal/libyaml/options.go
ingydotnet: 9b569fe

docs/plugins.md
ingydotnet: 9f244e9

plugin/limits/plugin.go
ingydotnet: 5f64fcd

yaml.go
ccoVeille: I was suggesting to rename WithoutPlugin to ResetPlugin because using it with limit plugin wouldn't remove it, but reset the limit plugin to its default value.

But then I feel you remove it from the documentation only 🤔

plugin.go
ccoVeille: Go interfaces are not meant to be imported.
They can but they don't need it to be referenced in the code implenting them.

Anything that satisfies this would match

        CheckDepth(depth int, ctx *DepthContext) error 
        CheckAlias(aliasCount, constructCount int) error

BTW a Go interface doesn't even need to be exported to be satisfied/implemented, they can be exported but they don't need it.

So the argument of having a user who needs this at the package level doesn't sound valid to me.

I'm fine with the `yaml.WithPlugin` being at the root level.

But someone who wants to use the limit plugin that you implement in YAML/plugin/limit would have to import it.

I won't fight against it, but it doesn't make sense for to have a bunch of interface at the root level.

internal/libyaml/options.go
ccoVeille: I might be wrong, but I feel like Colin was expecting a `type DepthKind string` and not two const Being `string`

I'm raising this, because it's what I was expecting.

yaml.go
```diff
@@ -254,6 +259,64 @@ func Options(opts ...Option) Option {
...
+		for _, p := range plugins {
+			registered := false
+			if lp, ok := p.(LimitPlugin); ok {
+				o.DepthCheck = lp.CheckDepth
+				o.AliasCheck = lp.CheckAlias
+				registered = true
+			}
+			// Future plugin types add cases here (non-exclusive if)
+			if !registered {
+				return errors.New("yaml: unsupported plugin type")
```
ccoVeille:
```suggestion
				return fmt.Errorf("yaml: unsupported plugin type: %T", p)
```

yaml.go
```diff
@@ -254,6 +259,64 @@ func Options(opts ...Option) Option {
...
+// Plugins use public types and can be implemented by external packages.
+func WithPlugin(plugins ...any) Option {
+	return func(o *libyaml.Options) error {
+		for _, p := range plugins {
+			registered := false
+			if lp, ok := p.(LimitPlugin); ok {
+				o.DepthCheck = lp.CheckDepth
+				o.AliasCheck = lp.CheckAlias
+				registered = true
+			}
```
ccoVeille, line 295: I feel like there's something strange here, but I'm unsure what is the solution.

Having something like this code differs from the Option patterns that was pretty good. Here, it would require to write here what the yaml/plugin/limit supports in terms of overload in the libyaml.Option

I feel like this should be in the yaml/plugin package, or better in the yaml/plugin/limit one.

But I'm unsure how to write it.

My problem is that if we write 10+ plugins, this file would become way to complex and coupled

I'm OK with current code being merged like this. But I wanted to share my concern

yaml.go
ccoVeille: I don't think a need to dereference the pointer 


```suggestion
		for name, val := range cfg.Plugin {
```

plugin.go
ingydotnet: @ccoVeille, Claude says:
You're right that Go interfaces are implicitly satisfied — a third-party plugin never needs to import LimitPlugin to implement it. The original rationale was imprecise.      
The actual reason LimitPlugin lives at the root is that WithPlugin (in yaml.go) does a type assertion: p.(LimitPlugin). The interface must be in the same package as the dispatch logic. Since WithPlugin is part of the yaml public API, LimitPlugin must be too. Moving the interface to a subpackage would create a circular dependency.

yaml.go
ingydotnet: I'm thinking now that we don't need this function at all. It's a premature optimization and I think there are other ways to accomplish what people want. I'm going to get rid of it and then we can add it if people ask for it.

plugin.go
ccoVeille: Yes, and no.

The fact we have to deal with the interface in the with plugin let me think there is a problem. Here we reach the issue I raised here

https://github.com/yaml/go-yaml/pull/313#discussion_r2934014987

There argument of cyclic dependency doesn't make sense either. Yes, it would create a problem if the code stays like that.

The interface you created here you could in a internal/plugin that would only define this interface. And import nothing.

Then yaml (root package) and limit/plugin package could import it.

But once again, to refer the other comment I quoted. I'm not even sure there is a need for this strange code in the yaml package. Listing all plugins and handling their implementation sounds strange to me. I mean there is something that could be better. But I didn't figure out what for now

yaml.go
ingydotnet: 627e83f

internal/libyaml/options.go
ingydotnet: I think 627e83f3 addresses this. Unresolve if I'm wrong.

ccoVeille: It does. Thanks

yaml.go
ingydotnet: c6269ff

yaml.go
ingydotnet: 2978947

yaml.go
ingydotnet: Noted. We probably want to have a registration system for plugins, but we can add that later, I think.

plugin.go
ingydotnet: I think things might become clearer once we have two or three plug-ins. It might be hard to understand where I'm going with it. Let's table this convo for when we have a couple more plugins...

internal/libyaml/options.go
```diff
@@ -1,7 +1,5 @@
-//
-// Copyright (c) 2025 The go-yaml Project Contributors
+// Copyright 2025 The go-yaml Project Contributors
```
colinjlacy, line 1: Note for the future - probably worth having a chore PR to go through existing files and update to current year

plugin/limit/plugin.go
```diff
@@ -0,0 +1,191 @@
+// Copyright 2025 The go-yaml Project Contributors
```
colinjlacy: Since these are new files, I think they should have been written with the current year

ccoVeille: I do agree. I reported it on another file, maybe in another PR 🤔

internal/libyaml/options.go
ccoVeille: There is no need to keep it up to date, I think. It doesn't sound logic but it's OK

ingydotnet: 4926215

plugin/limit/plugin.go
ingydotnet: 4926215

PR #319, opened 2026-03-15
Add DumpError for all dump stack errors

ingydotnet:
Introduce DumpError as the unified error type for the dump pipeline, mirroring the LoadError pattern added in PR #311 for the load stack.

## Motivation

Before this change the dump stack had no structured error type: errors escaped as a mix of bare panics (raw strings or raw error values), untyped fmt.Errorf strings from failf(), EmitterError, WriterError, and — incorrectly — LoadErrors/LoadError in dumper.go.  None of these were introspectable by callers: you couldn't tell which stage failed or extract the underlying cause without string-matching.

### New types and helpers (internal/libyaml/errors.go)

* Four new Stage constants covering the dump pipeline: RepresenterStage, SerializerStage, EmitterStage, WriterStage.
* DumpError struct: Stage + Message + unexported err for Unwrap(). Error() format: "go-yaml dump error in <stage>: <message>".
* NewDumpError(stage, message, cause) public constructor.
* failDump(stage, err)  — wraps an existing error as DumpError.
* failDumpf(stage, fmt, args...) — formats a new DumpError message. Both helpers panic with *YAMLError so handleErr catches them and returns them as ordinary Go errors.

### representer.go — 7 error sites converted

| was | now |
|-----|-----|
| Fail(err) (MarshalYAML) | failDump(RepresenterStage, err) | | Fail(err) (MarshalText) | failDump(RepresenterStage, err) | | panic("cannot represent type: …") | failDumpf(RepresenterStage, …) | | panic(err) (getStructInfo) | failDump(RepresenterStage, err) | | panic(fmt.Sprintf("cannot have key…")) | failDumpf(RepresenterStage, …) | | failf("explicitly tagged !!binary…") | failDumpf(RepresenterStage, …) | | failf("cannot represent invalid UTF-8…") | failDumpf(RepresenterStage, …) |

The three bare panics (type, getStructInfo, inline-map key) previously escaped handleErr; they now become returned *DumpError values.  Unused fmt import removed.

### serializer.go — 3 error sites + must() rewritten

The three failf() calls in node() are now failDumpf(SerializerStage,…).

must() previously collapsed all emitter/writer errors into a single untyped failf string.  It now type-switches on the error:
* EmitterError  → failDumpf(EmitterStage, e.Message)
* WriterError   → failDump(WriterStage, unwrapped-cause)
  (errors.Unwrap strips the "write error: " prefix that writer.go adds
  so that WriterStage in the DumpError already communicates that context)
* default       → failDumpf(SerializerStage, msg)

### dumper.go — 2 fixes

* Lines 77-86: the AllDocuments slice-check incorrectly returned a LoadErrors/LoadError (load-stack types).  Replaced with a plain DumpError{Stage: RepresenterStage}.
* SetIndent: bare panic("yaml: cannot indent…") replaced with failDumpf(SerializerStage, …) for consistent error type. (SetIndent has no handleErr so the panic still propagates, but it is now a *YAMLError wrapping *DumpError instead of a raw string.) Unused errors import removed.

### yaml.go — public API additions

* Stage doc comment updated to cover load and dump.
* Four new Stage constants re-exported (RepresenterStage …WriterStage).
* DumpError type alias added to the error-types block.
* NewDumpError var added alongside NewLoadError.

## Tests

* errors_test.go: runDumpErrorTest handler + buildDumpError helper, registered under the "dump-error" key.
* testdata/errors.yaml: 5 new dump-error test cases covering all four dump stages and Unwrap behaviour.
* yaml_test.go TestMarshalErrors: changed from PanicMatches to ErrorMatches + errors.As checks now that the bare panics are caught. Test struct field renamed panic→stage to reflect the new approach.
* yaml_test.go TestEncoderWriteError: updated expected message from "yaml: write error: …" to "go-yaml dump error in writer: …" and added errors.As / Stage assertions.
* yaml_test.go TestMarshalerError: fixed argument order in ErrorIs (was assert.ErrorIs(t, errFailing, err); now assert.ErrorIs(t, err, errFailing)) — the old order was accidentally correct only because the returned error was the raw errFailing; now it is a DumpError that wraps errFailing via Unwrap().
* node_test.go: two ErrorMatches expectations updated to the new "go-yaml dump error in serializer: …" format.

### DIFF
--- a/internal/libyaml/errors.go
+++ b/internal/libyaml/errors.go
@@ -12,16 +12,24 @@ import (
 	"strings"
 )
 
-// Stage identifies the processing stage where an error occurred during YAML loading.
+// Stage identifies the processing stage where an error occurred during YAML
+// loading or dumping.
 type Stage string
 
 const (
+	// Load stages
 	ReaderStage      Stage = "reader"      // Input reading and encoding
 	ScannerStage     Stage = "scanner"     // Tokenization
 	ParserStage      Stage = "parser"      // Event stream parsing
 	ComposerStage    Stage = "composer"    // Node tree construction
 	ResolverStage    Stage = "resolver"    // Tag resolution
 	ConstructorStage Stage = "constructor" // Go value construction
+
+	// Dump stages
+	RepresenterStage Stage = "representer" // Go value to Node tree
+	SerializerStage  Stage = "serializer"  // Node tree to events
+	EmitterStage     Stage = "emitter"     // Events to YAML bytes
+	WriterStage      Stage = "writer"      // Output writing
 )
 
 // LoadError represents an error that occurred while loading a YAML document.
@@ -86,6 +94,45 @@ func NewLoadError(stage Stage, message string, mark Mark, cause error) *LoadErro
 		Mark:    mark,
 		err:     cause,
 	}
+}
+
+// DumpError represents an error that occurred while dumping a YAML document.
+//
+// It identifies the processing stage where the error occurred and provides
+// an optional underlying cause via Unwrap.
+type DumpError struct {
+	Stage   Stage  // Processing stage where error occurred
+	Message string // Error description
+
+	// Error chaining
+	err error // Underlying error (for Unwrap support)
+}
+
+// Error returns the error message with stage information.
+// Format: "go-yaml dump error in <stage>: <message>"
+func (e *DumpError) Error() string {
+	return fmt.Sprintf("go-yaml dump error in %s: %s", e.Stage, e.Message)
+}
+
+// Unwrap returns the underlying error.
+func (e *DumpError) Unwrap() error {
+	return e.err
+}
+
+// NewDumpError creates a DumpError with an underlying cause.
+// The cause is accessible via Unwrap for use with [errors.Is] and [errors.As].
+func NewDumpError(stage Stage, message string, cause error) *DumpError {
+	return &DumpError{Stage: stage, Message: message, err: cause}
+}
+
+// failDump panics with a YAMLError wrapping a DumpError for the given stage.
+func failDump(stage Stage, err error) {
+	panic(&YAMLError{&DumpError{Stage: stage, Message: err.Error(), err: err}})
+}
+
+// failDumpf panics with a YAMLError wrapping a formatted DumpError.
+func failDumpf(stage Stage, format string, args ...any) {
+	panic(&YAMLError{&DumpError{Stage: stage, Message: fmt.Sprintf(format, args...)}})
 }
 
 // EmitterError represents an error that occurred during emitting.
--- a/internal/libyaml/errors.go
+++ b/internal/libyaml/errors.go
@@ -126,11 +126,12 @@ func NewDumpError(stage Stage, message string, cause error) *DumpError {
 }
 
 // failDump panics with a YAMLError wrapping a DumpError for the given stage.
-// If err is already a *DumpError it is passed through unchanged to avoid
+// If err is exactly a *DumpError it is passed through unchanged to avoid
 // double-wrapping (e.g. a user MarshalYAML that returns yaml.NewDumpError).
+// Errors that merely wrap a *DumpError are treated as ordinary errors so that
+// the outer wrapper's message and context are preserved.
 func failDump(stage Stage, err error) {
-	var de *DumpError
-	if errors.As(err, &de) {
+	if de, ok := err.(*DumpError); ok {
 		panic(&YAMLError{de})
 	}
 	panic(&YAMLError{&DumpError{Stage: stage, Message: err.Error(), err: err}})
--- a/internal/libyaml/errors.go
+++ b/internal/libyaml/errors.go
@@ -126,7 +126,13 @@ func NewDumpError(stage Stage, message string, cause error) *DumpError {
 }
 
 // failDump panics with a YAMLError wrapping a DumpError for the given stage.
+// If err is already a *DumpError it is passed through unchanged to avoid
+// double-wrapping (e.g. a user MarshalYAML that returns yaml.NewDumpError).
 func failDump(stage Stage, err error) {
+	var de *DumpError
+	if errors.As(err, &de) {
+		panic(&YAMLError{de})
+	}
 	panic(&YAMLError{&DumpError{Stage: stage, Message: err.Error(), err: err}})
 }
 
--- a/internal/libyaml/errors_test.go
+++ b/internal/libyaml/errors_test.go
@@ -66,16 +66,22 @@ func runDumpErrorTest(t *testing.T, tc TestCase) {
 	assert.Truef(t, ok, "from should be map[string]any, got %T", tc.From)
 
 	err := buildDumpError(t, errorSpec)
-	assert.NotNilf(t, err, "buildDumpError should not return nil")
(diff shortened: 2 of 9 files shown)


internal/libyaml/errors.go
```diff
@@ -88,6 +96,45 @@ func NewLoadError(stage Stage, message string, mark Mark, cause error) *LoadErro
...
+}
+
+// NewDumpError creates a DumpError with an underlying cause.
+// The cause is accessible via Unwrap for use with [errors.Is] and [errors.As].
+func NewDumpError(stage Stage, message string, cause error) *DumpError {
+	return &DumpError{Stage: stage, Message: message, err: cause}
+}
+
+// failDump panics with a YAMLError wrapping a DumpError for the given stage.
+func failDump(stage Stage, err error) {
```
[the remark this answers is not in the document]
ccoVeille: Good remark, the fix is pretty easy indeed

internal/libyaml/errors_test.go
```diff
@@ -58,6 +59,54 @@ func runLoadErrorTest(t *testing.T, tc TestCase) {
...
 }
 
+func runDumpErrorTest(t *testing.T, tc TestCase) {
+	t.Helper()
+
+	errorSpec, ok := tc.From.(map[string]any)
+	assert.Truef(t, ok, "from should be map[string]any, got %T", tc.From)
+
+	err := buildDumpError(t, errorSpec)
+	got := err.Error()
```
ccoVeille, line 72: This would panic if err is nil

internal/libyaml/serializer.go
```diff
@@ -230,14 +231,27 @@ func (s *Serializer) emit(event Event) {
...
+// must panics if the given error is non-nil, routing to the appropriate stage.
 func (s *Serializer) must(err error) {
 	if err != nil {
-		msg := err.Error()
-		if msg == "" {
-			msg = "unknown problem generating YAML content"
+		switch e := err.(type) {
+		case EmitterError:
+			failDumpf(EmitterStage, "%s", e.Message)
+		case WriterError:
```
ccoVeille: This is not a good pattern. We should use errors.As

ccoVeille (approved):
Some minor remarks, otherwise 👍

internal/libyaml/errors.go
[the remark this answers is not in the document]
ingydotnet: 0190836

internal/libyaml/serializer.go
ingydotnet: 0190836

internal/libyaml/errors_test.go
ingydotnet: 0190836

internal/libyaml/errors.go
```diff
@@ -88,6 +96,51 @@ func NewLoadError(stage Stage, message string, mark Mark, cause error) *LoadErro
...
+func NewDumpError(stage Stage, message string, cause error) *DumpError {
+	return &DumpError{Stage: stage, Message: message, err: cause}
+}
+
+// failDump panics with a YAMLError wrapping a DumpError for the given stage.
+// If err is already a *DumpError it is passed through unchanged to avoid
+// double-wrapping (e.g. a user MarshalYAML that returns yaml.NewDumpError).
+func failDump(stage Stage, err error) {
+	var de *DumpError
+	if errors.As(err, &de) {
```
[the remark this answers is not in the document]
ingydotnet: 7f41dde

internal/libyaml/errors_test.go
```diff
@@ -58,6 +59,55 @@ func runLoadErrorTest(t *testing.T, tc TestCase) {
...
+	err := buildDumpError(t, errorSpec)
+	assert.NotNilf(t, err, "buildDumpError should not return nil")
+	got := err.Error()
+	want, ok := tc.Want.(string)
+	assert.Truef(t, ok, "want should be string, got %T", tc.Want)
+
+	assert.Equalf(t, want, got, "error message mismatch")
+
+	// Verify Stage field if specified
+	if stageStr, ok := errorSpec["stage"].(string); ok {
```
ccoVeille: We are not handling the fact the stage is defined and is not a string.

This could let people think something is OK when it's not

internal/libyaml/errors_test.go
```diff
@@ -58,6 +59,55 @@ func runLoadErrorTest(t *testing.T, tc TestCase) {
...
 
+func runDumpErrorTest(t *testing.T, tc TestCase) {
+	t.Helper()
+
+	errorSpec, ok := tc.From.(map[string]any)
+	assert.Truef(t, ok, "from should be map[string]any, got %T", tc.From)
+
+	err := buildDumpError(t, errorSpec)
+	assert.NotNilf(t, err, "buildDumpError should not return nil")
+	got := err.Error()
```
ccoVeille: Calling assert is not enough as it reports the error with t.Error, not t.Fail, so it would panic anyway on line 70

But then the issue is our test suite. Sometimes we need to gave something that call fail like here,or we have to return err being nil and call t.Fail

internal/libyaml/errors_test.go
ingydotnet: a6d771c

internal/libyaml/errors_test.go
ingydotnet: a6d771c

internal/libyaml/serializer.go
```diff
@@ -230,14 +231,28 @@ func (s *Serializer) emit(event Event) {
 	s.must(s.Emitter.Emit(&event))
 }
 
-// must panics if the given error is non-nil.
+// must panics if the given error is non-nil, routing to the appropriate stage.
 func (s *Serializer) must(err error) {
 	if err != nil {
```
ccoVeille: I would use a leave early


```suggestion
 func (s *Serializer) must(err error) {
 	if err == nil {
		return
	}
```

internal/libyaml/serializer.go
```diff
@@ -230,14 +231,28 @@ func (s *Serializer) emit(event Event) {
...
+			// "write error: " prefix that WriterError adds internally.
+			cause := we.Err
+			if unwrapped := errors.Unwrap(we.Err); unwrapped != nil {
+				cause = unwrapped
+			}
+			failDump(WriterStage, cause)
+		}
 		msg := err.Error()
 		if msg == "" {
 			msg = "unknown problem generating YAML content"
```
ccoVeille: Adding the error type could be interesting in such case


```suggestion
			msg = fmt.Sprintf ("unknown problem generating YAML content with %T", err)
```

ccoVeille (approved):

colinjlacy (approved):

internal/libyaml/serializer.go
ingydotnet: 2b5aace

internal/libyaml/serializer.go
ingydotnet: 2b5aace

internal/libyaml/serializer.go
ingydotnet: 2b5aace

colinjlacy (approved):
Approving. I question the idea of `failDump` and `failDumpf` panicking, but I can see that it matches previous behavior

PR #326, opened 2026-04-01
Add perfsprint linter, address findings

scop:
None of the findings here are in hot paths, but they don't hurt readability or otherwise IMHO, so why not.


ccoVeille (approved):
Took me time to review it, but LGTM 👍

colinjlacy (approved):

PR #330, opened 2026-04-12
Fix shared aliases behavior between documents

### LINKED ISSUE #328: Alias dereferincing between documents
>> go-yaml accepts cross-document alias references, which the YAML spec forbids (anchors are scoped to a single document).
>
>
> We should detect at go-yaml level if it's relevant
_Originally posted by @ccoVeille in https://github.com/mikefarah/yq/issues/2647#issuecomment-4190973328_
            

It's confirmed by

> https://yaml.org/spec/1.2.2/#91-documents
>
> A YAML character [stream](https://yaml.org/spec/1.2.2/#streams) may contain several documents. Each document is completely independent from the rest.

I tried with go-yaml

```console
$ cat /tmp/alias-bug.yaml
---
&foo 42
---
*foo
```

```console
$ ./go-yaml -t /tmp/alias-bug.yaml
- {token: STREAM-START}
- {token: DOCUMENT-START}
- {token: ANCHOR, value: foo}
- {token: SCALAR, value: 42}
- {token: DOCUMENT-START}
- {token: ALIAS, value: foo}
- {token: STREAM-END}
```

```console
$ ./go-yaml -y /tmp/alias-bug.yaml
42
---
42
```

the bug is confirmed

We should report to something like this

```console
$ ./go-yaml -y /tmp/alias-bug.yaml
42
Failed to process YAML:failed to decode YAML: go-yaml load error in composer at L4.C1: unknown anchor 'foo' referenced
```

Also, my IDE reported me this when I worked on the file I wanted to test 😄 

Image

ccoVeille:
Considering the YAML specification mentions that all documents are independent, alias should not be shared between documents

```console
$ cat /tmp/alias-bug.yaml
---
&foo 42
---
*foo
```

```console
$ ./go-yaml -t /tmp/alias-bug.yaml
- {token: STREAM-START}
- {token: DOCUMENT-START}
- {token: ANCHOR, value: foo}
- {token: SCALAR, value: 42}
- {token: DOCUMENT-START}
- {token: ALIAS, value: foo}
- {token: STREAM-END}
```

Before:

```console
$ ./go-yaml -y /tmp/alias-bug.yaml
42
---
42
```

After

```console
$ ./go-yaml -y /tmp/alias-bug.yaml
Failed to process YAML:failed to decode YAML: go-yaml load error in composer at L4.C1: unknown anchor 'foo' referenced
```

Fixes #328

### DIFF
--- a/internal/libyaml/composer.go
+++ b/internal/libyaml/composer.go
@@ -137,6 +137,9 @@ func (c *Composer) node(kind Kind, tag, value string) *Node {
 // document composes a document node by parsing its content between
 // DOCUMENT_START and DOCUMENT_END events.
 func (c *Composer) document() *Node {
+	// Anchors are scoped to a single document.
+	c.resetAnchors()
+
 	n := c.node(DocumentNode, "", "")
 	c.doc = n
 	c.expect(DOCUMENT_START_EVENT)
@@ -271,7 +274,7 @@ func (c *Composer) init() {
 	if c.doneInit {
 		return
 	}
-	c.anchors = make(map[string]*Node)
+	c.resetAnchors()
 	// Peek to get the encoding from STREAM_START_EVENT
 	if c.peek() == STREAM_START_EVENT {
 		c.encoding = c.event.GetEncoding()
@@ -283,6 +286,10 @@ func (c *Composer) init() {
 	if c.streamNodes {
 		c.returnStream = true
 	}
+}
+
+func (c *Composer) resetAnchors() {
+	c.anchors = make(map[string]*Node)
 }
 
 // Destroy cleans up the composer by deleting any pending event and the
--- a/internal/libyaml/composer.go
+++ b/internal/libyaml/composer.go
@@ -137,6 +137,9 @@ func (c *Composer) node(kind Kind, tag, value string) *Node {
 // document composes a document node by parsing its content between
 // DOCUMENT_START and DOCUMENT_END events.
 func (c *Composer) document() *Node {
+	// Anchors are scoped to a single document.
+	c.resetAnchors()
+
 	n := c.node(DocumentNode, "", "")
 	c.doc = n
 	c.expect(DOCUMENT_START_EVENT)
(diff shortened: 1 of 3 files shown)


ccoVeille:
Here I don't know if we should control this with an option or not

The v3 had[ this bug](https://go.dev/play/p/nK-6IrFt_74), but some people may use it as "feature"

image

I don't think we should promote having a bug as a feature.

But because of the possible discussion, I'm creating the MR in draft.

colinjlacy:
I agree this is worth fixing. It does raise the risk of breaking people's code, but since it puts us closer to the spec I think it's worth merging. Being able to point to the spec if/when people say it's a breaking change will help our argument, IMO

colinjlacy (approved):

ingydotnet:
If this causes problems we can address it with an option or a plugin. The plugin could be : anchor-alias or bugwards-compat or something.

I'd like to be able to move forward towards spec correctness with confidence.

That said, cross document anchors can be really useful. YAMLScript supports them in code mode and data mode. It also errors on them in bare (plain YAML) mode.

I'd bet this was done on purpose. You could set up vars with anchors in one doc and then use them in a following doc. That's one thing that sucks about something like:
```
stuff:
  <<: *defaults
  foo: 42
```

You need to define the defaults node somewhere in the document, and you almost certainly don't want it as part of the data, so you'd need to post-process it afterwards to remove it.

I know why it was done this way in the spec and I think it was a mistake.
I would like to change that in a YAML 2.0 spec.

ingydotnet (approved):
I approve but I'd like to chat about this before merging.

I'd like to find out why this was done this way.

ingydotnet:
My research shows that this was added inadvertently in v3. 
Also, there are no tests for it.
I think we can merge this now and provide an option later for cross document aliases.

colinjlacy (approved):
I just re-tested this after the latest commits were added. Any reason we can't go ahead and merge? Seems like everyone's on board.

ingydotnet:
I'll resolve conflicts here and merge.

ccoVeille:
Thanks. I didn't notice there was conflicts

ingydotnet:
I resolved conflicts with main and converted the tests here to YAML ones, which read much more clearly (at least to me!).

ingydotnet (approved):
:+1:

colinjlacy (approved):

ccoVeille:
I thought we had merged it 🤔

PR #333, opened 2026-04-15
Fix float-to-string unmarshal regression

### LINKED ISSUE #331: Behaviour change in rc4: cannot construct !!float `55.7351` into string
To reproduce:

```go
func TestYamlFloat(t *testing.T) {
	b := `example: 55.7351`
	var res map[string]string
	require.NoError(t, yaml.Unmarshal([]byte(b), &res))
}
```

I don't know if that's expected- but this was possible before.

andig:
Fix https://github.com/yaml/go-yaml/issues/331

## Summary

- Add `reflect.String` case to `constructFloat`, allowing YAML float values to be unmarshaled into Go string fields
- This follows the same pattern already used by `constructInt` and `constructBool`
- Fixes the error: `cannot construct !!float into string`

## Problem

When unmarshaling YAML like:
```yaml
value: 55.7351
```

Into a Go struct with a string field:
```go
var result map[string]string
yaml.Unmarshal(data, &result)
```

It fails with: `yaml: construct errors: line 1: cannot construct !!float '55.7351' into string`

## Solution

Added the missing `reflect.String` case to `constructFloat` in `internal/libyaml/constructor.go`:

```go
case reflect.String:
    // Allow float to string conversion
    out.SetString(n.Value)
    return true
```

## Test plan

- [x] Added 12 test cases covering float/int/bool to string conversions
- [x] All existing tests pass
- [x] Tested edge cases: negative floats, scientific notation, infinity, NaN

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/internal/libyaml/constructor.go
+++ b/internal/libyaml/constructor.go
@@ -374,6 +374,9 @@ func (c *Constructor) constructFloat(n *Node, resolved any, out reflect.Value) b
 				}
 			}
 		}
+	case reflect.String:
+		out.SetString(n.Value)
+		return true
 	case reflect.Interface:
 		out.Set(reflect.ValueOf(resolved))
 		return true
--- a/yaml_test.go
+++ b/yaml_test.go
@@ -418,6 +418,59 @@ var unmarshalTests = []struct {
 		map[string]string{"a": strings.Repeat("\x00", 52)},
 	},
 
+	// Scalar to string conversions.
+	// Float to string (regression test for constructFloat missing reflect.String case).
+	{
+		"a: 55.7351",
+		map[string]string{"a": "55.7351"},
+	},
+	{
+		"a: -3.14159",
+		map[string]string{"a": "-3.14159"},
+	},
+	{
+		"a: 1.23e10",
+		map[string]string{"a": "1.23e10"},
+	},
+	{
+		"a: .inf",
+		map[string]string{"a": ".inf"},
+	},
+	{
+		"a: -.inf",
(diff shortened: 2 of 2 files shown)


dolmen:
Regression in comparison to what? Have you identified a particular change?

andig:
See related issue!

ccoVeille (approved):
Thanks for fixing this

andig:
Anything missing to get this merged?

ingydotnet:
> Anything missing to get this merged?

I was missing. Back now. Seems good, but I want to review a bit more. Hopefully today.

ingydotnet:
@andig can you amend the commit to remove the co-authored by Claude text, please?

I use Claude Code heavily for this project and others, but I haven't let it show up as an author yet. 
I'd like to have a bigger conversation around this and settle on a formal policy.

Some projects require a full prompt history with commits.
We might get to that, but for now I just want to keep all commits the same.

andig:
@ingydotnet would you mind just adjusting as you like on merge? Much easier than me editing the commits?

ingydotnet (approved):

andig:
@ingydotnet retesting I'm seeing this difference in the output:

```
  - Old (v3): yaml: line 2: mapping values are not allowed in this context                                                                                          
  - New (v4-rc4): go-yaml load error in scanner at L2.C6: mapping values are not allowed in this context
 ```

I guess this is intentional? However, the Lx.Cy notation is not entirely obvious. Does this warrant a separate issue?

PR #349, opened 2026-05-12
Fix !!merge tag regression for https://github.com/mikefarah/yq/issues/2677

W-Floyd:


ingydotnet:
@W-Floyd thanks. Can you add a test where #348 wasn't working and #349 does work?

W-Floyd:
See https://github.com/mikefarah/yq/pull/2705

ccoVeille:
@W-Floyd while I understand you closed because you found out a fix for yq that would work.

I'm still curious, and I think @ingydotnet too, what you tried and if the fix you tried in go-yaml could be considered or not.

What were you trying that is missing in go-yaml

I feel you found out something for yq, but I would like you to explain a bit more thanks

W-Floyd:
Sure, this is my understanding.

Per [this comment](https://github.com/mikefarah/yq/issues/2677#issue-4300162021), we can see that there was a change to this behavior, specifically [here](https://github.com/yaml/go-yaml/commit/fbd745be58df058ec8eb7d7b42468458f8d8ec64#diff-e331d2f8ca45209abed21d6207bc16aaa2e20593a2cf3eaf184a42ae9c6eb153L296-R298), which removed this:
```go
if nodeValue == "<<" {
    defaultTag = mergeTag
}
```
and [changes](https://github.com/yaml/go-yaml/commit/fbd745be58df058ec8eb7d7b42468458f8d8ec64#diff-e12595387295c7db06c272bbbac8e7dbf51789ae58880605e99e3cb5e2ac321eL1225-R1225)
```go
func isMerge(n *Node) bool {
	return n.Kind == ScalarNode && n.Value == "<<" && (n.Tag == "" || n.Tag == "!" || shortTag(n.Tag) == mergeTag)
}
```
to:
```go
func isMerge(n *Node) bool {
	return n.Kind == ScalarNode && shortTag(n.Tag) == mergeTag
}
```

The first point is addressed [here](https://github.com/yaml/go-yaml/pull/348/changes#diff-acdb842818ed4387defe492012e61d62a3e2475479469795b647b443473e01ccR78-R80).

What I tried adding in this PR was effectively the second part, wherein we interpret an untagged string `<<` as implying the `!!merge` tag. Now I wonder if this would actually be correct to keep ([for YAML 1.1 at least](https://yaml.org/type/merge.html))

Issue #289, opened 2026-02-09
[feature] merge two *yaml.Node-s API?


sitano:
What do you think of adding `Merge(dst, src *yaml.Node)` that will be able to merge mapping node keys instead of replacing them? Something like https://github.com/sitano/go-yaml-merge.

```
foo:
    a: 1
    b: 2

+

foo:
    b: 3
    c: 4

===

foo:
    a: 1
    b: 3
    c: 4
```

I think lots of users will benefit from having this function in std lib.

wrouesnel:
+1 to this. Most of the code already exists to support merge nodes, but there's no way to do something like 

```go
newNode := node.Merge(otherNode, nextOtherNode)
```

(where I imagine the precedence ordering would travel left to right (i.e. left most takes priority)

ccoVeille:
This sounds like a good feature to add.

Feel free to open a PR

sitano:
The draft is here https://github.com/yaml/go-yaml/pull/353. I would love you to take a look first so we can discuss if the approach looks appealing to you where to move forward.

ufukty:
Not here to vote for or against having this feature inside the official parser but I feel this one would not be _complete_ without allowing the merge logic to recur down to deeper levels. Such as:

```yml
a: scalar
b: { x: scalar }
c: [ scalar ]
```
\+
```yml
b: { y: scalar }
c: [ scalar ]
```
=
```yml
a: scalar
b: { x: scalar, y: scalar }
c: [ scalar, scalar ]
```

Not sure what behavior users would expect.

sitano:
@ufukty we do exactly that - recursively merge maps. for Seqs we however replace by default, but there is also an option to append. see https://github.com/yaml/go-yaml/pull/353/changes#diff-9798a9c4558bea307d5f10a8d138fd0204317b7805aa3dc7f2a1c688b2284c15R583-R673.

ingydotnet:
I finally got to thinking on this issue this morning. I had a private discussion with @colinjlacy about it.
At first I wanted to make this a plugin because it certainly will need a lot of config options and that's what plugins are all about.
But it doesn't really fit the plugin model.

My experience with merge functions is that everyone has their own specific needs, and that has certainly already surfaced here. But the problem is bigger than that. Merge is just one function in the world of data transformation. If we're going to offer this we need to plan for that. We need to plan bigger.

[...]

PR #353, opened 2026-05-21
feat(node): add ability to merge mapping nodes (node.merge())

sitano:
A merge API for a `yaml.Node` that can merge/replace mapping and sequence nodes depending on the configuration. The implementation currently consumes `src` Node to avoid copy-pasting the whole tree hierarchy for the use cases where in the most cases I expect the `src` value will be dumped.

```
func (dst *Node) Merge(src *Node, opts ...MergeOption) error {}
func (dst *Node) MergeWithOptions(src *Node, opts MergeOptions) error {}

type MergeOptions struct {
	// When MergeMap is true mapping nodes are merged instead of replaced. Default is true.
	MergeMap bool
	// When AppendSeq is true sequence nodes are appended instead of replaced. Default is false.
	AppendSeq bool
}

var (
	ErrYamlUnknown          = errors.New("yaml: unknown error")
	ErrYamlUnmergable       = errors.New("yaml: unmergable error")
	ErrYamlInvalidNodeKinds = errors.New("yaml: invalid node kinds")
	ErrYamlManyDocs         = errors.New("yaml: too many documents")
)
```

Example
===

```
map:
  key1: value1
  key2:
    map2:
      key3: value2
      key4: value3
seq:
- a

+

map:
  key2:
    map2:
      key4: value4
      key5: value5
  key4: value6
seq:
- b

->

map:
  key1: value1
  key2:
    map2:
      key3: value2
      key4: value4  # overwritten
      key5: value5  # added
  key4: value6 # added
seq:
- a
- b  # appended
```

Notes
===

See test material for more details on the specifics of behavior.

The rules for processing nulls, implicit and explicit values are a bit complicated. Take a look at the tests. 

There are a few cut corners but overall I think its more of a good than the opposite. e.g., it probably should not allow multi doc merge.

### DIFF
--- a/internal/libyaml/node_merge.go
+++ b/internal/libyaml/node_merge.go
@@ -1,0 +1,227 @@
+// Copyright 2026 The go-yaml Project Contributors
+// SPDX-License-Identifier: Apache-2.0
+
+// Node merger.
+// Merges mapping and sequence YAML nodes.
+
+package libyaml
+
+import (
+	"errors"
+)
+
+var (
+	ErrYamlUnknown          = errors.New("yaml: unknown error")
+	ErrYamlUnmergeable      = errors.New("yaml: unmergeable error")
+	ErrYamlInvalidNodeKinds = errors.New("yaml: invalid node kinds")
+	ErrYamlManyDocs         = errors.New("yaml: too many documents")
+)
+
+// MergeOptions configures Node.Merge().
+type MergeOptions struct {
+	// When MergeMap is true mapping nodes are merged instead of replaced. Default is true.
+	MergeMap bool
+	// When AppendSeq is true sequence nodes are appended instead of replaced. Default is false.
+	AppendSeq bool
+}
+
+type MergeOption func(*MergeOptions)
+
+// MergeMap configures Node.Merge() to merge mapping nodes.
+func MergeMap(opts *MergeOptions) {
+	opts.MergeMap = true
+}
+
+// ReplaceMap configures Node.Merge() to replace mapping nodes.
+func ReplaceMap(opts *MergeOptions) {
+	opts.MergeMap = false
+}
+
+// AppendSeq configures Node.Merge() to append sequence nodes.
+func AppendSeq(opts *MergeOptions) {
+	opts.AppendSeq = true
+}
+
+// ReplaceSeq configures Node.Merge() to replace sequence nodes.
+func ReplaceSeq(opts *MergeOptions) {
+	opts.AppendSeq = false
+}
+
+// WithAppendSeq configures Node.Merge() to append sequence nodes.
+func WithAppendSeq(appendSeq bool) MergeOption {
+	return func(opts *MergeOptions) {
+		opts.AppendSeq = appendSeq
+	}
+}
+
+// WithMergeMap configures Node.Merge() to merge mapping nodes.
+func WithMergeMap(mergeMap bool) MergeOption {
+	return func(opts *MergeOptions) {
+		opts.MergeMap = mergeMap
+	}
+}
+
+// DefaultMergeOptions returns the default MergeOptions. By default mapping nodes are merged and sequence nodes are replaced.
+func DefaultMergeOptions() MergeOptions {
+	return MergeOptions{
+		MergeMap:  true,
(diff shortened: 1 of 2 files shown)


colinjlacy:
Hey @sitano, thanks for opening this and sorry for the radio silence on our part. This is a great idea, and it's never an easy implementation since it always involves configuration options and design choices that won't always make everyone happy. So really, thanks for taking this on.

There are some placement concerns about where these functions should live - e.g. `MergeOptions` should probably live in `options.go` - but aren't really important just yet as we talk through what has been implemented.

The first thing that jumps out to me is that this needs to be documented. It's totally fine that it hasn't been done yet, what with us still talking through the implementation approach. But that would help alleviate some possible confusion among end users about some of the choices that were made made. 

I notice that this doesn't copy values, but instead copies pointers. That gives me the impression that subsequent changes to `src` will always be mapped into `dst`. That could make for unexpected behavior if, for example, I want to merge a stable state of a `src` doc into a `dst` doc before I make changes to the `src`. Maybe a merge-by-value or merge-by-pointer option is warranted?

I'm personally flip-flopping on the nil value merge behavior. I can see it going both ways - ignore nils when merging, or overwrite with nils when merging. It may be worth adding an option there too.

Last, the implicit gap that I see in sequence merges is that a sequence of maps won't get the recursive merge that a top-level map would. That's fine and I'm personally ok with that continuing to be the behavior once this PR is ready for review, but I think it's something worth documenting. Could be a great opportunity for a further improvement if someone wants that functionality.

sitano:
Thank you @colinjlacy for the response. I will start working towards the points you've mentioned.

> The first thing that jumps out to me is that this needs to be documented. It's totally fine that it hasn't been done yet, what with us still talking through the implementation approach. But that would help alleviate some possible confusion among end users about some of the choices that were made made.

Totally agree. Could you pls share where do you think it makes sense to put the doc? I suppose you mean something beyond the golang func documentation? In the `/doc`, the package level comment in the file?

> I notice that this doesn't copy values, but instead copies pointers. That gives me the impression that subsequent changes to src will always be mapped into dst.

When I was designing this, I thought that people will just dispose the `src` after use. I was not sure making a deep copy first internally is a good thing from the overhead perspective on allocations, especially considering that I expected in the most cases when you merge you would rather dispose `src`. But we certainly can change that.

> That could make for unexpected behavior if, for example, I want to merge a stable state of a src doc into a dst doc before I make changes to the src. Maybe a merge-by-value or merge-by-pointer option is warranted?

Yeah, we can do that. How do you think does it makes sense to keep those 2 options or just shift to always merge-by-value that will imply full copy of the src?

> I'm personally flip-flopping on the nil value merge behavior. I can see it going both ways - ignore nils when merging, or overwrite with nils when merging. It may be worth adding an option there too.

Yeah, that came historically too to somehow distinuish behavior between implicit and explicit nulls and also to give a `src` an option to affect `dst` nodes. But certainly we can have this as an Option. Mostly it is about answering the question what is the expected behavior when a node from src merges with implicit/explicit null value into a dst node. E.g. should src null be ignored when it merges into the map/seq, or it should erase the target. If it the src null is ignored how src could erase the keys in target..

> Last, the implicit gap that I see in sequence merges is that a sequence of maps won't get the recursive merge that a top-level map would.

Yeah. The thing here is that the seq nodes does not have IDs so we can't identify them and match. Seq are basically ordered Sets that support Append, Insert(pos), Delete. At the moment I struggle to imagine a meaningful strategy for maps merge in seqs.

ingydotnet:
Hi @sitano,

Given https://github.com/yaml/go-yaml/issues/289#issuecomment-4747240439 this PR is unlikely to go further, although I look forward to further discussion with you about it if you'd like.

I just wanted to point out in general that if we had a Merge function we wouldn't have a MergeWithOptions. 
We thought about having LoadWithOptions and DumpWithOptions but went with the much better (imho) variadic forms of Load and Dump that took any number of WithFoo() composable options.

sitano:
Thank you @ingydotnet for the response. I don't have any arguments to add. I just needed `node.Merge(node)` when I started this. This work can live externally, it is not necessary to embed. I can probably close that one, or feel free to close it.

PR #362, opened 2026-06-12
Handle empty legacy YAML streams

### LINKED ISSUE #342: BC break: yaml construct error on empty/comment-only stream
After upgrading to master we're seing this error when parsing empty/ comment-only yaml:

image

https://github.com/evcc-io/evcc/issues/29672

What's striking here is:

- error on empty document
- error on comment-only document
- strange error message containing `<unknown position>`

ingydotnet:
fixes #342

Restore v3-compatible behavior for the classic decode APIs when the input contains zero YAML documents. yaml.Unmarshal now treats empty, whitespace-only, and comment-only streams as a no-op and returns nil, leaving the destination value unchanged. Decoder.Decode continues to report io.EOF for the same zero-document streams, matching v3 streaming behavior.

Keep the stricter v4 Load API behavior intact: Load still reports "no documents in stream" for empty and comment-only input unless it is being called through the internal legacy compatibility path.

Preserve comments that have no document node when callers opt into WithStreamNodes. Comments unfolded onto STREAM_END_EVENT are now copied onto the emitted StreamNode, so comment-only streams still return the existing single StreamNode shape but no longer drop the comments.

Tests cover Unmarshal no-op behavior for empty, whitespace-only, and comment-only streams across struct, map, and Node targets; Decoder io.EOF behavior; strict Load errors for comment-only input; and WithStreamNodes comment preservation.

### DIFF
--- a/internal/libyaml/composer.go
+++ b/internal/libyaml/composer.go
@@ -163,6 +163,11 @@ func (c *Composer) createStreamNode() *Node {
 	if !c.Textless && c.event.Type != NO_EVENT {
 		n.Line = c.event.StartMark.Line
 		n.Column = c.event.StartMark.Column
+		if c.event.Type == STREAM_END_EVENT {
+			n.HeadComment = string(c.event.HeadComment)
+			n.LineComment = string(c.event.LineComment)
+			n.FootComment = string(c.event.FootComment)
+		}
 	}
 	return n
 }
--- a/internal/libyaml/loader.go
+++ b/internal/libyaml/loader.go
@@ -221,6 +221,9 @@ func loadSingle(in []byte, out any, opts *Options) error {
 	// Load first document
 	err = l.Load(out)
 	if err == io.EOF {
+		if opts.FromLegacy {
+			return nil
+		}
 		msg := "yaml: no documents in stream"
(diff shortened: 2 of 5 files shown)


ccoVeille (approved):
LGTM 👍

colinjlacy (approved):
👍

andig:
Confirmed working in https://github.com/evcc-io/evcc/pull/30789

ingydotnet:
> Confirmed working in [evcc-io/evcc#30789](https://github.com/evcc-io/evcc/pull/30789)

Thanks for checking, @andig!

ingydotnet:
@andig would you like me to push a new RC tag?

andig:
Thank you, but commit works just fine ;)

thaJeztah:
@ingydotnet a new rc tag would be appreciated; I see dependabot started to open PRs in compose to update to rc.5, but updating would break some downstreams for us (https://github.com/ddev/ddev/pull/8492)

ingydotnet:
@thaJeztah origin/main is now tagged v4.0.0-rc.6

thaJeztah:
Thank you!! 🙏

PR #363, opened 2026-06-13
Adds gitignore exceptions for existing YAML files

### LINKED ISSUE #339: .gitignore: /*.yaml rule hides tracked config files when module is vendored
The repository's `.gitignore` contains `/*.yaml`, but the repo ships tracked `.yaml` files at the root: `.golangci.yaml`, `.ls-lint.yaml`, and `.yamllint.yaml`. Since they are already tracked, the rule is inert in this repo. This is misleading and may appear contradictory though.

The real issue surfaces downstream. When a consumer runs `go mod vendor`, all files including `.gitignore` are copied into their `vendor/` directory. There, the `.yaml` files are untracked, so the `/*.yaml` rule kicks in and hides them.

This breaks vendor consistency checks; `go mod vendor` then `git status --porcelain --ignored` will fail when files are ignored in `vendor/` (`go mod vendor` produces files that git would silently drop)

## Reproducer

```bash
mkdir /tmp/test && cd /tmp/test
git init
go mod init example.com/test
cat > main.go <<'EOF'
package main

import _ "go.yaml.in/yaml/v4"

func main() {}
EOF
go mod tidy
go mod vendor
git status --porcelain --ignored -- vendor/
```

Output includes:

```
!! vendor/go.yaml.in/yaml/v4/.golangci.yaml
!! vendor/go.yaml.in/yaml/v4/.ls-lint.yaml
!! vendor/go.yaml.in/yaml/v4/.yamllint.yaml
```

## Suggested fix

Add negation rules for the shipped `.yaml` files:

```gitignore
/*.yaml
!/.golangci.yaml
!/.ls-lint.yaml
!/.yamllint.yaml
```

The change to .gitignore was introduced in https://github.com/yaml/go-yaml/commit/aeb787767b12f2c4540a73233c92818efeeb3ccf

colinjlacy:
To test:
```sh
mkdir /tmp/test && cd /tmp/test
git init
go mod init example.com/test

go mod edit -replace=go.yaml.in/yaml/v4=github.com/colinjlacy/go-yaml/v4@issue339

cat > main.go <<'EOF'
package main

import _ "go.yaml.in/yaml/v4"

func main() {}
EOF

go mod tidy
go mod vendor
git status --porcelain --ignored -- vendor/
```

Should print out:
```sh
?? vendor/
!! vendor/go.yaml.in/yaml/v4/CONTRIBUTING.md
!! vendor/go.yaml.in/yaml/v4/README.md
```

Fixes #339


ccoVeille (approved):

.gitignore
```diff
@@ -9,3 +9,8 @@
 /go-yaml
 /*.md
 /*.yaml
```
ccoVeille: We should have never added these according to me.

It's a matter of cleaning its own house.

The fact one needs this make the repository strange for everyone.

@ingydotnet could you consider removing this, and use .git/ignore local files for this.

I feel these exclusion could make future contributions problematic

.gitignore
ingydotnet: Let's remove lines `/.claude/`, `/note/`, `/worktree/`. `/*.md`, `/*.yaml` and then the new additions.

These were from my personal dev practices and should be handled per developer.

ingydotnet (changes requested):
@colinjlacy Can you make the adjustments I suggested. 
@ccoVeille You were right to point out that they shouldn't have been added.

.gitignore
ingydotnet: Totally agree with you @ccoVeille.
It was wrong for me to add these (and some others in there).

.gitignore
colinjlacy: @ingydotnet done

ccoVeille (approved):

ccoVeille:
I would like all these to bebsquashed into one commit before merging.

There is no need for the back and forth situation we can this in this PR, once merged in main

ingydotnet (approved):

PR #365, opened 2026-06-14
Fixes cyclical reference in LoadErrors creation

### LINKED ISSUE #345: Stack overflow when custom UnmarshalYAML wraps unmarshal callback error
`LoadErrors.Is()` causes infinite recursion (fatal stack overflow) when a custom `UnmarshalYAML` implementation wraps the error returned by the `unmarshal` callback with `%w`.

## Minimal reproduction

```go
package main

import (
    "errors"
    "fmt"
    "os"
    "time"

    "go.yaml.in/yaml/v4"
)

type Inner struct {
    Value string `yaml:"value"`
}

type Wrapper struct {
    Item Inner `yaml:"item"`
}

func (w *Wrapper) UnmarshalYAML(unmarshal func(any) error) error {
    var single Inner
    if err := unmarshal(&single); err != nil {
        // Standard Go idiom — wrapping with %w
        return fmt.Errorf("wrapper failed: %w", err)
    }
    return nil
}

type Root struct {
    Items []Wrapper `yaml:"items"`
}

func main() {
    done := make(chan struct{})
    go func() {
        data := []byte("items:\n  - not-an-object\n")
        var r Root
        err := yaml.Unmarshal(data, &r)
        if err != nil {
            // This triggers fatal error: stack overflow
            _ = errors.Is(err, errors.New("test"))
        }
        close(done)
    }()

    select {
    case <-done:
        fmt.Println("Completed")
    case <-time.After(3 * time.Second):
        fmt.Println("TIMEOUT: stack overflow")
        os.Exit(1)
    }
}
```

## Root cause

`yaml.Unmarshal` reuses a single `LoadErrors` object. When the custom `UnmarshalYAML` wraps the callback's error with `%w`, and `yaml.Unmarshal` adds that wrapped error back to the same `LoadErrors`, it creates a cycle:

```
LoadErrors → ConstructError → fmt.wrapError → LoadErrors
```

`LoadErrors.Is()` calls `errors.Is(err, target)` on each `ConstructError`, which unwraps back to the same `LoadErrors`, causing infinite recursion.

## Environment

- `go.yaml.in/yaml/v4 v4.0.0-rc.4`
- Go 1.26

## Workaround

Custom `UnmarshalYAML` implementations must use `%v` instead of `%w` when wrapping errors from the `unmarshal` callback, but this is non-obvious and breaks standard Go error wrapping conventions.

colinjlacy:
The issue was a self-reference in the error chain, causing unwrap to iterate over the same item in the `LoadErrors` slice. The slice returned to the custom unmarshaler shared backing storage with `c.TypeErrors`, so when the %w-wrapped error was later appended to `c.TypeErrors`, it overwrote the very slot that wrapped error pointed back to — closing the loop. This fix copies the sub-slice in callLegacyConstructor so the returned LoadErrors has its own slot in memory.

To test:
```sh
# Setup
git clone --depth 1 --branch issue345 https://github.com/colinjlacy/go-yaml.git /tmp/go-yaml
mkdir /tmp/repro && cd /tmp/repro
go mod init repro
go mod edit -replace go.yaml.in/yaml/v4=/tmp/go-yaml

# Test file
cat > main.go <<'EOF'
package main
import (
	"errors"
	"fmt"
	"os"
	"time"
	"go.yaml.in/yaml/v4"
)
type Inner struct {
	Value string `yaml:"value"`
}
type Wrapper struct {
	Item Inner `yaml:"item"`
}
func (w *Wrapper) UnmarshalYAML(unmarshal func(any) error) error {
	var single Inner
	if err := unmarshal(&single); err != nil {
		// Standard Go idiom — wrapping with %w
		return fmt.Errorf("wrapper failed: %w", err)
	}
	return nil
}
type Root struct {
	Items []Wrapper `yaml:"items"`
}
func main() {
	done := make(chan struct{})
	go func() {
		data := []byte("items:\n  - not-an-object\n")
		var r Root
		err := yaml.Unmarshal(data, &r)
		if err != nil {
			// Previously triggered: fatal error: stack overflow
			_ = errors.Is(err, errors.New("test"))
		}
		close(done)
	}()
	select {
	case <-done:
		fmt.Println("Completed")
	case <-time.After(3 * time.Second):
		fmt.Println("TIMEOUT: stack overflow")
		os.Exit(1)
	}
}
EOF

# Run test
go mod tidy
go run .
```

Also note that this branch adds regression tests at the interface level and unit tests at the lib level.

Fixes #345

### DIFF
--- a/internal/libyaml/constructor.go
+++ b/internal/libyaml/constructor.go
@@ -1022,7 +1022,7 @@ func (c *Constructor) callLegacyConstructor(n *Node, u legacyConstructor) (good 
 		defer handleErr(&err)
 		c.Construct(n, reflect.ValueOf(v))
 		if len(c.TypeErrors) > terrlen {
-			issues := c.TypeErrors[terrlen:]
+			issues := append([]*LoadError{}, c.TypeErrors[terrlen:]...)
 			c.TypeErrors = c.TypeErrors[:terrlen]
 			return &LoadErrors{issues}
 		}
--- a/internal/libyaml/constructor_test.go
+++ b/internal/libyaml/constructor_test.go
@@ -7,6 +7,7 @@
 package libyaml
 
 import (
+	"fmt"
 	"reflect"
 	"testing"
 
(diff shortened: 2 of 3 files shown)


ccoVeille (approved):
Thanks for this

colinjlacy:
That last push removed comments from the test files that were breaking godoclint, FYI

ccoVeille (approved):

ingydotnet (approved):

PR #366, opened 2026-06-14
Fixes indexing bug in emitter

### LINKED ISSUE #337: Folded block scalars gain one newline per `yaml.Node` round-trip
## Summary

Parsing a YAML document that contains a folded block scalar (`>`) and then re-emitting it via `yaml.Node` + `yaml.Encoder` inserts one extra blank line inside the scalar body. On the next parse, that blank is captured as a real newline in the decoded string, so every subsequent round-trip adds another newline. Any pipeline that parses + emits the same document more than once (overlay systems, template renderers, `yaml.Node`-based refactors) progressively corrupts every folded-scalar value.

Literal block scalars (`|`) do not exhibit this and are a control.

## Versions

- Go 1.23
- `go.yaml.in/yaml/v4 v4.0.0-rc.4`
- Also reproduces on `gopkg.in/yaml.v3 v3.0.1` (same emitter lineage).

## Reproduction

```go
package main

import (
  "bytes"
  "fmt"

  "go.yaml.in/yaml/v4"
)

const src = `key: >
  Heading:

    * first item
    * second item
`

func main() {
  current := []byte(src)
  for i := 0; i <= 3; i++ {
    var m map[string]string
    if err := yaml.Unmarshal(current, &m); err != nil {
      panic(err)
    }
    fmt.Printf("round %d: decoded (%d bytes): %q\n", i, len(m["key"]), m["key"])

    var n yaml.Node
    if err := yaml.Unmarshal(current, &n); err != nil {
      panic(err)
    }
    var buf bytes.Buffer
    enc := yaml.NewEncoder(&buf)
    enc.SetIndent(2)
    if err := enc.Encode(&n); err != nil {
      panic(err)
    }
    if err := enc.Close(); err != nil {
      panic(err)
    }
    current = buf.Bytes()
  }
}
```

## Actual output

```
round 0: decoded (41 bytes): "Heading:\n\n  * first item\n  * second item\n"
round 1: decoded (42 bytes): "Heading:\n\n\n  * first item\n  * second item\n"
round 2: decoded (43 bytes): "Heading:\n\n\n\n  * first item\n  * second item\n"
round 3: decoded (44 bytes): "Heading:\n\n\n\n\n  * first item\n  * second item\n"
```

Replacing `>` with `|` in `src` keeps the decoded value at 41 bytes for all four rounds.

## Expected output

A parse-then-emit round-trip of a `yaml.Node` should preserve every scalar's decoded value, i.e. `decode(roundTrip(b)) == decode(b)`. This holds for literal scalars; it does not hold for folded scalars.

## Validation

- **Spec.** The input is well-formed per [YAML 1.2.2 §8.1](https://yaml.org/spec/1.2.2/#81-block-scalar-styles): bare `>` header (prod [162]), auto-detected content indent of 2 (§8.1.1.1, prod [163]), default CLIP chomping (§8.1.1.2, prod [164]), and a normal folded line + empty separator + more-indented block — the `l-nb-diff-lines` shape (§8.1.3, prod [181], Examples 8.10–8.12).
- **Cross-implementation.** PyYAML 6.x decodes the same input to the same 41-byte value and is byte-stable across 4 `safe_load` → `dump(default_style='>')` → `safe_load` cycles. Round-trip preservation is achievable for this input; the drift is specific to the Go emitter lineage.

## Hypothesis

The folded emitter (`encode.go` / `emitterc.go`, shared with `yaml.v3`) appears to prepend one empty line to the scalar body each time it emits folded style — likely the fix-up the spec requires when a folded scalar begins with more-indented content. On re-parse that empty line is correctly interpreted as a paragraph break (`\n` in the decoded string); on the next emit the fix-up is applied again; the blank count grows by one per round-trip. Literal style has no such fix-up and is therefore stable.

## Real-world impact

An OpenAPI docs pipeline built on `libopenapi` (which depends on `yaml/v4`) had every `>`-authored description grow one `\n` longer per regeneration. The workaround we ended up shipping is to walk the source `yaml.Node` tree and flip every `FoldedStyle` scalar to `LiteralStyle` before any downstream tooling sees it — silently mutating user-authored `>` to `|` purely to dodge the emit asymmetry. A fix in the emitter would let `yaml.Node`-based pipelines stay honest about folded style.

colinjlacy:
This fixes #337 by addressing an indexing bug in how the emitter was adding newlines to folded block scalars, starting at index 0 instead of at the current index in the byte stream. It also adds a boundary to the lookahead so that we don't end up out of bounds. It then tests the boy-howdy out of the fix with unit, regression, and data-driven tests.

To test:
```sh
# Setup
git clone --depth 1 --branch issue337 https://github.com/colinjlacy/go-yaml.git /tmp/go-yaml
mkdir /tmp/repro && cd /tmp/repro
go mod init repro
go mod edit -replace go.yaml.in/yaml/v4=/tmp/go-yaml

# Test file
cat > main.go <<'EOF'
package main

import (
  "bytes"
  "fmt"

  "go.yaml.in/yaml/v4"
)

const src = `key: >
  Heading:

    * first item
    * second item
`

func main() {
  current := []byte(src)
  for i := 0; i <= 3; i++ {
    var m map[string]string
    if err := yaml.Unmarshal(current, &m); err != nil {
      panic(err)
    }
    fmt.Printf("round %d: decoded (%d bytes): %q\n", i, len(m["key"]), m["key"])

    var n yaml.Node
    if err := yaml.Unmarshal(current, &n); err != nil {
      panic(err)
    }
    var buf bytes.Buffer
    enc := yaml.NewEncoder(&buf)
    enc.SetIndent(2)
    if err := enc.Encode(&n); err != nil {
      panic(err)
    }
    if err := enc.Close(); err != nil {
      panic(err)
    }
    current = buf.Bytes()
  }
}
EOF

# Run test
go mod tidy
go run .
```

### DIFF
--- a/internal/libyaml/emitter_test.go
+++ b/internal/libyaml/emitter_test.go
@@ -28,6 +28,60 @@ func TestEmitter(t *testing.T) {
 	})
 }
 
+func emitFoldedScalar(t *testing.T, value string) string {
+	t.Helper()
+
+	events := []Event{
+		NewStreamStartEvent(UTF8_ENCODING),
+		NewDocumentStartEvent(nil, nil, true),
+		NewScalarEvent(nil, nil, []byte(value), true, true, FOLDED_SCALAR_STYLE),
+		NewDocumentEndEvent(true),
+		NewStreamEndEvent(),
+	}
+
+	emitter := NewEmitter()
+	emitter.SetIndent(2)
+	var output []byte
+	emitter.SetOutputString(&output)
+	for i := range events {
+		err := emitter.Emit(&events[i])
+		assert.NoErrorf(t, err, "Emit() error: %v", err)
+	}
+	return string(output)
+}
+
+func TestEmitFoldedScalarNoExtraNewline(t *testing.T) {
+	cases := []struct {
+		name  string
+		value string
+		want  string
+	}{
+		{
+			name:  "heading then more-indented block",
+			value: "Heading:\n\n  * first item\n  * second item\n",
+			want:  ">\n  Heading:\n\n    * first item\n    * second item\n",
+		},
+		{
+			name:  "single newline between plain lines",
+			value: "one\ntwo\n",
+			want:  ">\n  one\n\n  two\n",
+		},
+		{
+			name:  "trailing more-indented block",
+			value: "intro\n\n  indented tail\n",
+			want:  ">\n  intro\n\n    indented tail\n",
+		},
+	}
+
+	for _, tc := range cases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			got := emitFoldedScalar(t, tc.value)
+			assert.Equal(t, tc.want, got)
+		})
+	}
+}
+
 func runEmitWriterTest(t *testing.T, tc TestCase) {
 	t.Helper()
 
--- a/internal/libyaml/emitter_test.go
+++ b/internal/libyaml/emitter_test.go
@@ -28,28 +28,6 @@ func TestEmitter(t *testing.T) {
 	})
 }
 
-func emitFoldedScalar(t *testing.T, value string) string {
-	t.Helper()
-
-	events := []Event{
-		NewStreamStartEvent(UTF8_ENCODING),
-		NewDocumentStartEvent(nil, nil, true),
-		NewScalarEvent(nil, nil, []byte(value), true, true, FOLDED_SCALAR_STYLE),
-		NewDocumentEndEvent(true),
-		NewStreamEndEvent(),
-	}
-
-	emitter := NewEmitter()
-	emitter.SetIndent(2)
-	var output []byte
-	emitter.SetOutputString(&output)
-	for i := range events {
-		err := emitter.Emit(&events[i])
-		assert.NoErrorf(t, err, "Emit() error: %v", err)
-	}
-	return string(output)
-}
-
 func TestEmitFoldedScalarNoExtraNewline(t *testing.T) {
 	cases := []struct {
 		name  string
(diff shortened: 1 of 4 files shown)


internal/libyaml/emitter_test.go
```diff
@@ -28,6 +28,60 @@ func TestEmitter(t *testing.T) {
...
+	}
+
+	for _, tc := range cases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			got := emitFoldedScalar(t, tc.value)
+			assert.Equal(t, tc.want, got)
+		})
+	}
+}
```
ccoVeille: I would have avoided to add emiyFoldedScalar

The whole code could be moved in the t.Run that is almost empty

node_test.go
```diff
@@ -300,6 +300,72 @@ func TestNodeOmitEmpty(t *testing.T) {
...
+	}
+	var buf bytes.Buffer
+	enc := yaml.NewEncoder(&buf)
+	enc.SetIndent(indent)
+	if err := enc.Encode(&n); err != nil {
+		t.Fatalf("Encode Node: %v", err)
+	}
+	if err := enc.Close(); err != nil {
+		t.Fatalf("Close encoder: %v", err)
+	}
```
ccoVeille: please consider using assert.NoError

node_test.go
```diff
@@ -300,6 +300,72 @@ func TestNodeOmitEmpty(t *testing.T) {
...
+			}
+			want := first["key"]
+
+			current := []byte(tc.src)
+			for round := 0; round < 4; round++ {
+				var m map[string]string
+				if err := yaml.Unmarshal(current, &m); err != nil {
+					t.Fatalf("round %d: Unmarshal: %v", round, err)
+				}
+				if m["key"] != want {
```
ccoVeille: Please use assert.Equal

ccoVeille (approved):
My comments would be only about the way you tests things

Except that LGTM 👍

ccoVeille (approved):

ingydotnet (approved):
lgtm. thanks @colinjlacy

PR #375, opened 2026-06-29
Fix Marshal mutating *Node input #371

### LINKED ISSUE #371: Marshal strips all resolvable tags in place
I encountered an issue in [libopenapi-validator](https://github.com/pb33f/libopenapi-validator).
But based on the digging with AIs, it might be triggered by go-yaml's bug.

Sorry, I tried to understand it as best I could, but there are so many layers...
Please confirm whether it is your library's bug or not.
Below is a summary I've prepared with AI through "our" digging.

## Summary
Marshalling a `*yaml.Node` mutates the caller's node tree instead of treating it as read-only input. The encode pipeline's Desolver stage (which strips inferable tags such as `!!str`, `!!int`, `!!map` to produce clean output) runs directly on the node you passed in — not on a copy. After a single `Marshal`, every resolvable tag in the tree has been set to `""`.
The tree is still structurally intact, but its type information is gone. Any code that holds a `*yaml.Node`, marshals it, and later inspects `.Tag` (e.g. tag-based type detection, schema validation, round-trip overlay systems) is silently broken — the first encode is a destructive side effect.

This is conceptually the same class of bug as #337 (folded scalars gaining a newline per `yaml.Node` round-trip): the encode path mutating data that flows through a `yaml.Node`. This one affects tags specifically.

## Versions
- go.yaml.in/yaml/v4, current main (also reproduces on the v4.0.0-rc.x series).
- Go 1.26.2
- OS: Debian GNU/Linux 12 (bookworm)

## Reproduction
Here is a test that fails:
```go
package yaml_test

import (
	"testing"

	"go.yaml.in/yaml/v4"
)

// Bug: yaml.Marshal MUTATES its *yaml.Node argument instead of treating it as
// read-only input. While encoding, the "desolver" stage strips resolvable tags
// (!!str, !!int, !!map, ...) directly off the caller's nodes — it runs on the
// node you passed in, not on a copy. After a single Marshal, every node's Tag is
// "" even though Marshal is conceptually a read-only serialization.
//
// This silently breaks any code that holds a *yaml.Node and later relies on its
// Tag (e.g. tag-based type detection): the tree is intact structurally but its
// type information is gone.
func TestMarshalMutatesNodeInput(t *testing.T) {
	var doc yaml.Node
	if err := yaml.Unmarshal([]byte("type: array\nlimit: 5\n"), &doc); err != nil {
		t.Fatal(err)
	}

	m := doc.Content[0]      // the mapping node
	mapTag := m.Tag          // "!!map"
	typeVal := m.Content[1]  // scalar "array", tagged "!!str"
	limitVal := m.Content[3] // scalar "5",     tagged "!!int"

	// Sanity: parsing assigned resolved tags.
	if typeVal.Tag != "!!str" || limitVal.Tag != "!!int" || mapTag != "!!map" {
		t.Fatalf("unexpected pre-state: map=%q type=%q limit=%q", mapTag, typeVal.Tag, limitVal.Tag)
	}

	// Marshal is supposed to be read-only. It is not.
	if _, err := yaml.Marshal(&doc); err != nil {
		t.Fatal(err)
	}

	// Correct behavior: the input nodes are untouched. This assertion FAILS
	// because Marshal wiped every tag to "" in place.
	if m.Tag != "!!map" || typeVal.Tag != "!!str" || limitVal.Tag != "!!int" {
		t.Fatalf("yaml.Marshal mutated its input nodes: map=%q type=%q limit=%q (want !!map/!!str/!!int)",
			m.Tag, typeVal.Tag, limitVal.Tag)
	}
}
```

Result: yaml.Marshal mutated its input nodes: map="" type="" limit="" (want !!map/!!str/!!int).

## Root cause
On the encode path, `Represent` returns the caller's node without copying, and the Desolver then mutates it in place:

internal/libyaml/representer.go — `Represent` returns the input node as-is for document nodes, and `nodev` returns the original `*Node` pointer for any other node. No defensive copy is made:
```go
// representer.go
if node != nil && node.Kind == DocumentNode {
    // Already a document node, return as-is
    return node
}
// ...
func (r *Representer) nodev(in reflect.Value) *Node {
    // Return the node as-is - no conversion needed
    return in.Interface().(*Node)
}
```

internal/libyaml/dumper.go — `Dump` then desolves that same node:
```go
node := d.representer.Represent("", reflect.ValueOf(v))
d.desolver.Desolve(node)   // <-- mutates the caller's tree
d.serializer.Serialize(node)
```

internal/libyaml/desolver.go — `Desolve` walks the tree and assigns n.Tag = "" on scalars (`desolveScalar`) and collections (`desolveCollection`), and recurses into `n.Content`.

The same pattern is present in `(*Node).Encode` and `(*Node).Dump` (`internal/libyaml/node.go`), which call `Represent` + `Desolve` directly — so a `*yaml.Node` passed as the value to encode is also mutated.

## Impact
- `yaml.Marshal(&node)` / `Encoder.Encode(&node)` are not idempotent and have a destructive side effect on first call.
- A node tree that is marshalled and then re-inspected loses all resolvable tag information.
- This breaks the natural assumption that serialization is read-only, matching the behavior of `encoding/json` and prior `gopkg.in/yaml.v3`.

## Expected behavior
Marshalling a `*yaml.Node` must not modify the input. The Desolver should operate on a copy of the tree (deep-copy before desolve), or be rewritten to be non-mutating, so the caller's `.Tag`/`.Style` fields are preserved.

## Suggested fix direction
Deep-copy the node tree at the boundary before the Desolver runs (in `Dumper.Dump`, `(*Node).Encode`, and `(*Node).Dump)`, or have the Desolver produce a transformed copy rather than mutating in place. The copy only needs to exist for the duration of encoding.

truffle-dev:
Fixes #371.

`Marshal` is documented to treat its input as read-only, but passing a
`*Node` leaves the caller's tree with every resolvable tag stripped:

```go
var node yaml.Node
yaml.Unmarshal([]byte("type: array\nlimit: 5\n"), &node)
yaml.Marshal(&node)
// node.Content[0].Tag is now "" (was "!!map"), and so on for the
// !!str / !!int children.
```

### Cause

The representer hands a user-supplied `*Node` straight into the rest of
the dump pipeline: `Represent` returns a document node as-is, and
`nodev` returns nested `*Node` values as-is. The desolver then clears
inferable tags (`!!str`, `!!int`, `!!map`, ...) **in place** to keep the
output clean, so the mutation lands on the caller's own node.

### Fix

Deep-copy the user-supplied node in the representer before it enters the
pipeline. A `deepCopyNode` helper copies `Content`, `Alias`, and
`Stream`, using a `seen` map so shared anchors keep their identity and
cyclic alias graphs don't recurse forever. It's wired at both `*Node`
entry points: the document-node branch in `Represent` and `nodev` (which
also covers `*Node` values nested inside a struct, map, or slice).

Only `*Node` inputs pay the copy cost; representer-built trees (from
maps, structs, slices) are already fresh and untouched.

### Verification

- `go test . -run TestMarshalDoesNotMutateNode` passes with the fix and
  fails without it (tags wiped on the input node).
- `go test ./...`: main, internal, and cmd suites green; the
  yaml-test-suite results are unchanged at 1383 pass / 225 known-fail.

### DIFF
--- a/internal/libyaml/representer.go
+++ b/internal/libyaml/representer.go
@@ -55,8 +55,10 @@ func (r *Representer) Represent(tag string, in reflect.Value) *Node {
 		node, _ = in.Interface().(*Node)
 	}
 	if node != nil && node.Kind == DocumentNode {
-		// Already a document node, return as-is
-		return node
+		// Deep-copy the user-supplied document so downstream stages (the
+		// desolver in particular) never mutate the caller's node tree.
+		// Marshal must treat its input as read-only.
+		return deepCopyNode(node, map[*Node]*Node{})
 	} else {
 		// Wrap the represented value in a document node
 		contentNode := r.represent(tag, in)
@@ -65,6 +67,52 @@ func (r *Representer) Represent(tag string, in reflect.Value) *Node {
 			Content: []*Node{contentNode},
 		}
 	}
+}
+
+// deepCopyNode returns a deep copy of n so that a caller passing a *Node to
+// Marshal is not mutated by later pipeline stages. The seen map preserves
+// shared node identity and prevents infinite recursion on cyclic anchor and
+// alias graphs.
+func deepCopyNode(n *Node, seen map[*Node]*Node) *Node {
+	if n == nil {
+		return nil
+	}
+	if c, ok := seen[n]; ok {
+		return c
+	}
+	c := &Node{
+		Kind:        n.Kind,
+		Style:       n.Style,
+		Tag:         n.Tag,
+		Value:       n.Value,
+		Anchor:      n.Anchor,
+		HeadComment: n.HeadComment,
+		LineComment: n.LineComment,
+		FootComment: n.FootComment,
+		Line:        n.Line,
+		Column:      n.Column,
+	}
+	seen[n] = c
+	c.Alias = deepCopyNode(n.Alias, seen)
+	if n.Content != nil {
+		c.Content = make([]*Node, len(n.Content))
+		for i, child := range n.Content {
+			c.Content[i] = deepCopyNode(child, seen)
+		}
+	}
+	if n.Stream != nil {
+		s := &Stream{Encoding: n.Stream.Encoding}
+		if n.Stream.Version != nil {
+			v := *n.Stream.Version
+			s.Version = &v
+		}
+		if n.Stream.TagDirectives != nil {
+			s.TagDirectives = make([]StreamTagDirective, len(n.Stream.TagDirectives))
+			copy(s.TagDirectives, n.Stream.TagDirectives)
+		}
+		c.Stream = s
+	}
+	return c
 }
 
 // From http://yaml.org/type/float.html, except the regular expression there
@@ -397,10 +445,11 @@ func (r *Representer) nilv() *Node {
 	}
 }
 
-// nodev returns a node value as-is without conversion.
+// nodev returns a deep copy of a node value. Copying ensures the caller's
+// node tree is never mutated by later pipeline stages, whether the *Node is
+// passed to Marshal directly or nested inside a struct, map, or slice.
 func (r *Representer) nodev(in reflect.Value) *Node {
-	// Return the node as-is - no conversion needed
-	return in.Interface().(*Node)
+	return deepCopyNode(in.Interface().(*Node), map[*Node]*Node{})
 }
 
 // Len returns the number of keys in the list.
--- a/node_test.go
+++ b/node_test.go
@@ -784,3 +784,21 @@ func TestNodeDumpInvalidOptions(t *testing.T) {
 	assert.NotNil(t, err)
 	assert.ErrorMatches(t, ".*indent must be.*", err)
 }
+
+func TestMarshalDoesNotMutateNode(t *testing.T) {
+	const data = "type: array\nlimit: 5\n"
+
+	var node yaml.Node
+	assert.NoError(t, yaml.Unmarshal([]byte(data), &node))
+
+	// An independent decode of the same input is the read-only baseline.
+	var want yaml.Node
+	assert.NoError(t, yaml.Unmarshal([]byte(data), &want))
+
+	_, err := yaml.Marshal(&node)
+	assert.NoError(t, err)
+
+	// Marshal must treat its input as read-only: the resolved tags
+	// (!!map, !!str, !!int) decoded into node must survive unchanged.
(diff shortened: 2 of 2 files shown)


ccoVeille:
I have asked Copilot to review in a separate agent discussion here is what was reported

```markdown
I reviewed PR **#375**’s file changes and here are the likely remaining edge cases / risks:

### What looks solid
- Deep copy is applied at both `*Node` ingress points:
  - `Represent` when input is already a `DocumentNode`
  - `nodev` for nested `*Node` inside structs/maps/slices
- `seen map[*Node]*Node` correctly preserves shared identity and prevents infinite recursion on cycles/aliases.
- Test `TestMarshalDoesNotMutateNode` covers the core regression.

### Potential edge cases / risks

1) **`Stream.Encoding` copy style**
- In `deepCopyNode`, `s := &Stream{Encoding: n.Stream.Encoding}`.
- If `Encoding` is a pointer/slice/map/interface-with-pointer-semantics (not sure without `Stream` type definition), this may still be shallow for that field.
- If it’s scalar/enum, no issue.

2) **Line/Column copied even though representer-built trees might ignore them**
- Copy includes `Line`/`Column`. Usually harmless.
- But if any downstream behavior unexpectedly depends on these positions for represented nodes, copying preserves original positions (probably desired, but worth noting).

3) **Performance on very large node graphs**
- This is O(N) memory/time per marshal for `*Node` inputs.
- Acceptable for correctness, but workloads marshaling huge prebuilt node graphs repeatedly may notice overhead.
- A benchmark would be useful (`Marshal(*Node)` before/after with large anchors/aliases).

4) **Test coverage breadth**
- Current test validates tag preservation on a simple map/int/string shape.
- Missing targeted tests for:
  - alias cycles
  - shared subtrees (same node referenced twice)
  - comments/style preservation after marshal
  - nested `*Node` inside struct/map/slice path (the `nodev` hook)

5) **Semantic expectation: identity no longer preserved through marshal pipeline**
- Previously (buggy) behavior mutated original nodes.
- Now marshal operates on clone, so any internal side-effect expectations (if anyone relied on them) disappear.
- That is the correct API behavior, but could surprise anyone depending on old side effects.

### Overall assessment
The fix direction is correct and likely robust for the reported bug (#371).  
The main things I’d still add are **2–4 focused tests** (aliases/shared nodes/nested `*Node`) plus an optional micro-benchmark.
```

PR #379, opened 2026-07-10
Allow timestamp-tagged scalars to unmarshal into strings

### LINKED ISSUE #376: Cannot construct !!timestamp `2023-01-01` into string
When updating from rc4 to rc5/rc6 I noticed that timestamp parsing into string stopped working.

### Example

```go
package main

import (
	"log"
	"strings"

	yaml "go.yaml.in/yaml/v4"
)

var raw = `
day: 2026-01-01
`

func main() {
	var untyped map[string]string
	err := yaml.NewDecoder(strings.NewReader(raw)).Decode(&untyped)
	if err != nil {
		log.Fatal(err.Error())
	}
	log.Println("success")
}
```

```
2026/07/02 11:03:04 yaml: construct errors: line 2: cannot construct !!timestamp `2026-01-01` into string
exit status 1
```

Related to #331

oguzkilcan:
constructTimestamp only handled time.Time and interface{} targets. A plain scalar resolving to the timestamp tag left a string field at its zero value instead of the raw text. This adds a string case that writes the scalar directly, the same fallback other scalar tags already have.

Fixes #376

### DIFF
--- a/yaml_test.go
+++ b/yaml_test.go
@@ -556,6 +556,46 @@ var unmarshalTests = []struct {
 		"a: 2015-01-01",
 		map[string]any{"a": time.Date(2015, 1, 1, 0, 0, 0, 0, time.UTC)},
 	},
+	{
+		// implicit timestamp tag into string.
+		"a: 2015-01-01",
+		map[string]string{"a": "2015-01-01"},
+	},
+	{
+		// implicit timestamp tag on quoted string into string.
+		"a: \"2015-01-01\"",
+		map[string]string{"a": "2015-01-01"},
+	},
+	{
+		// implicit timestamp tag with time and fraction into string.
+		"a: 2015-02-24T18:19:39.12Z",
+		map[string]string{"a": "2015-02-24T18:19:39.12Z"},
+	},
+	{
+		// implicit timestamp tag with time and fraction on quoted string into string.
+		"a: \"2015-02-24T18:19:39.12Z\"",
+		map[string]string{"a": "2015-02-24T18:19:39.12Z"},
+	},
+	{
+		// explicit timestamp tag on unquoted string into string.
+		"a: !!timestamp 2015-01-01",
+		map[string]string{"a": "2015-01-01"},
+	},
+	{
+		// explicit timestamp tag into string.
+		"a: !!timestamp \"2015-01-01\"",
+		map[string]string{"a": "2015-01-01"},
+	},
+	{
+		// explicit timestamp tag with time and fraction on unquoted string into string.
+		"a: !!timestamp 2015-02-24T18:19:39.12Z",
+		map[string]string{"a": "2015-02-24T18:19:39.12Z"},
+	},
+	{
+		// explicit timestamp tag with time and fraction into string.
+		"a: !!timestamp \"2015-02-24T18:19:39.12Z\"",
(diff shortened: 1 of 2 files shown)


yaml_test.go
```diff
@@ -556,6 +556,11 @@ var unmarshalTests = []struct {
 		"a: 2015-01-01",
 		map[string]any{"a": time.Date(2015, 1, 1, 0, 0, 0, 0, time.UTC)},
 	},
+	{
+		// implicit timestamp tag into a plain string field.
+		"a: 2015-02-24T18:19:39.12Z\n",
```
ccoVeille:
```suggestion
		"a: 2015-02-24T18:19:39.12Z",
```

Others tested values have no \n

ccoVeille: Are you sure about your test here?

It sounds like it would parse the string as string, no timestamp involved, so you might not be testing the code you changed

My suggestion

https://github.com/yaml/go-yaml/pull/379#discussion_r3563594306

yaml_test.go
ccoVeille, line 553: I feel like it's what should be tested

```
	{
		// explicit timestamp tag on unquoted string into string
		"a: !!timestamp 2015-01-01",
		map[string]string{"a": "2015-01-01"},
	},
	{
		// explicit timestamp tag into string.
		"a: !!timestamp \"2015-01-01\"",
		map[string]string{"a": "2015-01-01"},
	},
```

ccoVeille:
Thanks @oguzkilcan 

I feel like the fix is good.

But I'm doubtful about the tests.

Note: I edited your PR description to mention the issue you opened

oguzkilcan:
Thank you for the review. I incorporated your suggestions into the test suite. It run successfully.
```
❯ go test -run TestUnmarshal
PASS
ok      go.yaml.in/yaml/v4      0.194s
```

ccoVeille (approved):

ccoVeille (approved):

ccoVeille (approved):

ingydotnet:
`- !!timestamp foo` - I'm wondering if this should error.

I probably need to consult with the core team.

ingydotnet:
The 1.1 and 1.2 specs _allow_ it to load as a string. It's really up to us what to do.

`!!timestamp` isn't part of the YAML 1.2 spec's Core yaml-schema, so I'm a bit loathe to make this a default.

Honestly I'd like all of tag resolution and construction to be handled at the RuleSet plugin level. Then a RuleSet would be made for the current behavior inherited from go-yaml v3 and that would be the default, and there would become myriad other RuleSets to use or extend.

That said I just noticed this is a fix for a v4 regression, so I'll take a closer look now.

ingydotnet:
Looks like the regression was here: https://github.com/yaml/go-yaml/blob/082f4f4bc52bb12aaebcc7c1740360b8bd1ff8e7/internal/libyaml/constructor.go#L653-L666

and also regressed `!!float` but that got fixed here: https://github.com/yaml/go-yaml/commit/fcab72bcf4bb134980b6c1407af155587a6bd1f9

ingydotnet (changes requested):
@oguzkilcan Can you add a test for an implicit timestamp, as that is what was reported by #376.

ingydotnet (approved):

PR #380, opened 2026-07-11
Honor quote preference for legacy-quoted scalars

oguzkilcan:
`stringv()` method of `Representer` hardcoded single-quote style whenever a string needed quoting to avoid being misread as an old-style bool, base60 float, or merge key, bypassing the configured QuotePreference entirely. `Marshal()` always requests `WithV3Defaults()` (`QuoteLegacy`) for backward compatibility with the legacy encoder, so every caller of plain `Marshal()` silently got single-quoted output no matter what preference it asked for.

### DIFF
--- a/internal/libyaml/representer.go
+++ b/internal/libyaml/representer.go
@@ -289,6 +289,10 @@ func (r *Representer) stringv(tag string, in reflect.Value) *Node {
 	case needsQuoting:
 		// Force quoting for YAML 1.1 compatibility values
 		style = SingleQuotedStyle
+		// and honor the configured quote preference.
+		if r.quotePreference.ScalarStyle() == DOUBLE_QUOTED_SCALAR_STYLE {
+			style = DoubleQuotedStyle
+		}
 	default:
 		// Plain style by default - Desolver will add quotes if type mismatch
 		style = 0
--- a/testdata/encode.yaml
+++ b/testdata/encode.yaml
@@ -315,7 +315,7 @@
       a: '1:1'
     type: map[string]string
     want: |
-      a: '1:1'
+      a: "1:1"
 
 - encode:
     name: null byte in string
@@ -396,7 +396,7 @@
       '<<': []
     type: map[string]any
     want: |
-      '<<': []
+      "<<": []
 
 - encode:
     name: merge key as value
@@ -404,7 +404,7 @@
       foo: '<<'
     type: map[string]any
     want: |
-      foo: '<<'
+      foo: "<<"
 
 # Long strings
 - encode:
@@ -509,7 +509,7 @@
       a: off
     type: testStructA_String
     want: |
-      a: 'off'
+      a: "off"
 
 # Struct tests with yaml tags
 - encode:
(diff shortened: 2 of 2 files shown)


internal/libyaml/representer.go
```diff
@@ -287,8 +287,14 @@ func (r *Representer) stringv(tag string, in reflect.Value) *Node {
...
-		// Force quoting for YAML 1.1 compatibility values
-		style = SingleQuotedStyle
+		// Force quoting for YAML 1.1 compatibility values, honoring the
+		// configured quote preference.
+		switch r.quotePreference.ScalarStyle() {
+		case DOUBLE_QUOTED_SCALAR_STYLE:
+			style = DoubleQuotedStyle
+		default:
+			style = SingleQuotedStyle
+		}
```
ccoVeille: Something simpler would be better

```suggestion
		// Force quoting for YAML 1.1 compatibility values
		style = SingleQuotedStyle
		// and honor the configured quote preference.
		if r.quotePreference.ScalarStyle() == DOUBLE_QUOTED_SCALAR_STYLE {
			style = DoubleQuotedStyle
		}
```

But I can see the pattern you used in existing code, so why not.

Also, maybe there is a need to handle legacy quoted style here.

Your test shows the code you updated uses double quote for legacy, but I'm unsure how it was in the v3. We should check

testdata/encode.yaml
```diff
@@ -396,15 +396,15 @@
       '<<': []
     type: map[string]any
     want: |
-      '<<': []
+      "<<": []
```
ccoVeille, line 399: There is something strange to me here.

There is no reason to use the double quote unless we are in v3 compatibility, right?

Or is it our test suite that forces this setting.

What do you think @colinjlacy ?

ccoVeille: @oguzkilcan ?

oguzkilcan: Sorry @ccoVeille I missed the question mark there, I thought you were expecting @colinjlacy to answer.

Yes, it is the test suite. `runEncodeTest` uses `yaml.Marshal`, which always passes `WithV3Defaults()`, so the preference is `QuoteLegacy`, and `QuoteLegacy` means double quotes in the representer.

About v3 compatibility: v3 does not quote `<<` at all, so it cannot tell us single or double here. I kept double because `QuoteLegacy` already gives it without a special case.

I also added two `encode-opts` tests for `<<`, one with single and one with double, so both preferences are now covered.

```shell
❯ make test-internal
go test ./internal/...    
ok      go.yaml.in/yaml/v4/internal/libyaml     (cached)
ok      go.yaml.in/yaml/v4/internal/testutil/assert     (cached)
?       go.yaml.in/yaml/v4/internal/testutil/datatest   [no test files]
ALL INTERNAL FILES PASS
```

ccoVeille: OK thank you. I feel like we should move the test suites to v4 default then.

I'm not expecting this to be done in your PR.

We need to discuss it with other maintainers

ccoVeille (approved):

PR #388, opened 2026-08-03
Quote multiline scalars that start with tabs

### LINKED ISSUE #383: Emitted yaml not accepted back when block scalar starts with a tab character
Here is a reproduction

```go
package main

import (
	"fmt"
	"log"

	"go.yaml.in/yaml/v4"
)

type Data struct {
	Text string `yaml:"text"`
}

func main() {
	// Struct -> YAML (serialization)
	data := Data{
		Text: "\tthis\nis\nmultiline",
	}

	output, err := yaml.Marshal(data)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Println("Serialized YAML:")
	fmt.Println(string(output))

	var decoded Data

	err = yaml.Unmarshal([]byte(output), &decoded)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Println("Deserialized struct:")
	fmt.Printf("%+v\n", decoded)
}
```

https://go.dev/play/p/lG7KJfXJKpp

This seems to be caused by 2 things: 

This line in the emitter, that only adds the indentation indicator if the first character of the line is a space character

https://github.com/yaml/go-yaml/blob/b8d7755848342c16a63e5f628651aa7efb4dd55b/internal/libyaml/emitter.go#L1988-L1996

And also here, in the scanner, it aborts when finding a tab character while detecting indentation of the block scalar.

https://github.com/yaml/go-yaml/blob/b8d7755848342c16a63e5f628651aa7efb4dd55b/internal/libyaml/scanner.go#L1774-L1777

However, based on the spec here (If it is the correct one) : https://yaml.org/spec/1.2.2/#block-scalar-headers

> If no indentation indicator is given, then the content indentation level is equal to the number of leading [spaces](https://yaml.org/spec/1.2.2/#white-space-characters) on the first non-[empty line](https://yaml.org/spec/1.2.2/#empty-lines) of the contents. If there is no non-[empty line](https://yaml.org/spec/1.2.2/#empty-lines) then the content indentation level is equal to the number of spaces on the longest line.

It says nothing about lines containing tabs, only about lines that **doesn't start with spaces**.

So to me the issue is more in the scanner, where encountering a tab character should just stop the detection because it is the first non-space character in the line if the user decided to start their string with a tab character.

Guflly:
Fixes #383.

Multiline values whose first content character is a tab are now emitted as double-quoted scalars, so output from Marshal can be read back by Unmarshal. The same selection is used for unstyled Nodes created with SetString.

Tabs on later lines continue to use literal style. Tests cover direct leading tabs, initial line breaks, map values, and Nodes.

### DIFF
--- a/internal/libyaml/node.go
+++ b/internal/libyaml/node.go
@@ -250,6 +250,10 @@ func shouldUseLiteralStyle(s string) bool {
 	if !strings.Contains(s, "\n") || len(s) < 2 {
 		return false
 	}
+	withoutLeadingBreaks := strings.TrimLeft(s, "\r\n\u0085\u2028\u2029")
+	if strings.HasPrefix(withoutLeadingBreaks, "\t") {
+		return false
+	}
 	// Must contain at least one non-whitespace character
 	for _, r := range s {
 		if !unicode.IsSpace(r) {
--- a/yaml_test.go
+++ b/yaml_test.go
@@ -2984,28 +2984,38 @@ func TestScalarStyleWithTabs(t *testing.T) {
 		},
 		{
 			"\tThis starts with tab\nand is long enough\nfor literal style",
-			"|-\n    \tThis starts with tab\n    and is long enough\n    for literal style\n",
+			"\"\\tThis starts with tab\\nand is long enough\\nfor literal style\"\n",
 			"Multiline starting with tab",
 		},
 		{
 			"\tB\n\tC\n",
-			"|\n    \tB\n    \tC\n",
+			"\"\\tB\\n\\tC\\n\"\n",
 			"Tab B newline tab C newline",
 		},
 		{
 			"\ta\n",
-			"|\n    \ta\n",
+			"\"\\ta\\n\"\n",
 			"Tab + char + newline",
 		},
 		{
 			"\thello\n",
-			"|\n    \thello\n",
+			"\"\\thello\\n\"\n",
 			"Tab + text + newline",
 		},
 		{
 			"\t\nhello",
-			"|-\n    \t\n    hello\n",
+			"\"\\t\\nhello\"\n",
 			"Tab + newline + text",
+		},
+		{
+			"\n\tthis\nnext",
+			"\"\\n\\tthis\\nnext\"\n",
+			"Tab after initial newline",
+		},
+		{
+			"first\n\tsecond",
+			"|-\n    first\n    \tsecond\n",
+			"Tab after content line",
 		},
 	}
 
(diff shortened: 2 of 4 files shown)


yaml_test.go
```diff
@@ -2984,29 +2984,39 @@ func TestScalarStyleWithTabs(t *testing.T) {
 		},
 		{
 			"\tThis starts with tab\nand is long enough\nfor literal style",
-			"|-\n    \tThis starts with tab\n    and is long enough\n    for literal style\n",
+			"\"\\tThis starts with tab\\nand is long enough\\nfor literal style\"\n",
```
ccoVeille, line 2987: Maybe you could use string literal (here and everywhere) to improve readability

```suggestion
			`"\tThis starts with tab\nand is long enough\nfor literal style"`+"\n",
```

internal/libyaml/node.go
```diff
@@ -250,6 +250,10 @@ func shouldUseLiteralStyle(s string) bool {
 	if !strings.Contains(s, "\n") || len(s) < 2 {
 		return false
 	}
+	withoutLeadingBreaks := strings.TrimLeft(s, "\r\n\u0085\u2028\u2029")
+	if strings.HasPrefix(withoutLeadingBreaks, "\t") {
+		return false
+	}
```
ccoVeille, line 256: I'm surprised to see the Unicode separators in addition of `\r\n`

Could you cover this with tests?

Also I would like to see these Unicode character to constants with clear name and comment.

I'm a bit surprised by the fact it trims everything. It could have consequences for multi line maybe.

Something like \n\n\n\t

Maybe I'm wrong.

I feel like the issue you try to fix is about having some characters in front of \t

I feel like using strings.HasPrefix with iterative checks such as `\r\n\t`, `\n\t`, `\uxxxx\t`

yaml_test.go
```diff
@@ -3018,6 +3028,53 @@ func TestScalarStyleWithTabs(t *testing.T) {
...
+		data, err = yaml.Marshal(&node)
+		assert.NoError(t, err)
+		assert.Equal(t, testCase.scalarYAML, string(data))
+
+		var gotScalar string
+		err = yaml.Unmarshal(data, &gotScalar)
+		assert.NoError(t, err)
+		assert.Equal(t, testCase.value, gotScalar)
+	}
+}
```
ccoVeille, line 3076: I think we have something in the test suite that only uses .yaml files to handle this round trip check. I think it's in emitter.yaml

Could you please check?

Thanks

ccoVeille (commented):
Thanks for PR.

I reviewed it.

PR #389, opened 2026-08-03
Add UnmarshalerWithContext on the v3 line

### LINKED ISSUE #187: Provide context to Unmarshaler
For my project, I need a way to inject dependencies into an unmarshaled type or modify its behavior based on the unmarshaling context.

I'm thinking of something like this:
```go
type Foo struct {
  intValue int
}

func (f *Foo) UnmarshalYAML(ctx context.Context, value *Node) error {
  var strValue string
  value.Decode(&strValue)

  transformer := ctx.Value("transformer").(Transformer)
  f.intValue = transformer.Transform(strValue)
}

func main() {
  ctx := context.WithValue(context.Background(), "transformer", NewTransformer())
 
  var foo Foo
  yaml.UnmarshalWithContext(ctx, []byte{"foo"}, &foo)
}
```

I understand that my specific needs might go against some principles, and perhaps a different approach, such as transforming the decoded configuration into another type after unmarshaling, would be more suitable.

However, I believe that providing a `context.Context` to an `Unmarshaler` could be beneficial for other reasons as well.

My idea is to introduce a new `UnmarshalerWithContext` interface and an `UnmarshalWithContext` method. This method would forward the `context.Context` to the `UnmarshalerWithContext` interface if the target type implements it.

```
type UnmarshalerWithContext interface {
  UnmarshalYAML(ctx context.Context, value *Node) error
}

func UnmarshalWithContext(ctx context.Context, in []byte, out any) (err error) {
  ...
}
```

I'm aware that there's an ongoing discussion about a new `Unmarshaler` interface in issue #56, so it might be worth considering adding context there.

Additionally, I like the idea of adding options to Unmarshal as proposed in issue #155. In that scenario, the context could be provided via an option rather than by introducing new methods.

I've already implemented this functionality for my own needs in my fork (Draft PR #186). While I'm quite happy with it as is, it would be perfect if something similar could get a chance to be introduced in a future v4 release.

reuvenharrison:
Closes #187 on the v3 line. Carries forward the unmarshal half of #186.

A type implementing `UnmarshalerWithContext` receives the context its decode was started with:

```go
type UnmarshalerWithContext interface {
	UnmarshalYAML(ctx context.Context, value *Node) error
}
```

with `UnmarshalWithContext`, `Decoder.DecodeWithContext` and `Node.DecodeWithContext` to supply it. A type implementing both interfaces gets this one. A type implementing neither is untouched, and `Unmarshal` on a context-only type passes `context.Background()`.

## On the open design question in #187

The discussion there stalled on whether this should be an option rather than new interfaces, so this PR takes a position and I am happy to be overruled on it.

An option would mean holding a `context.Context` in the options struct, which is the pattern the Go standard library and vet checks steer away from, and which @ccoVeille raised in the thread. What this PR does instead keeps the context as a first argument at every public entry point, and the decoder holds it only for the length of a single call:

```go
func (dec *Decoder) DecodeWithContext(ctx context.Context, v interface{}) error {
	prev := dec.ctx
	dec.ctx = ctx
	defer func() { dec.ctx = prev }()
	return dec.Decode(v)
}
```

So it is not stored across calls, and nothing about one decode reaches another. If you would still prefer the option shape, say so and I will rework it; the dispatch and the plumbing are the same either way, and only the entry points change.

## Why v3 and not main

#186 targets `main`, which is now `v4` and still at `rc`. The released line is `v3.0.5`, so a consumer that cannot take a release candidate cannot use this until v4 ships.

I have the same change ported to the current `main` layout as well, where the restructure into `internal/libyaml` is what left #186 conflicting. Happy to open that as a separate PR, or to fold this into a refresh of #186, whichever you prefer. I did not want to push at #186 uninvited.

## Tests

`decode_context_test.go` covers the context reaching an unmarshaler, `Unmarshal` on a context-only type getting a background context, and twenty concurrent decodes each asserting it observed its **own** context. The existing suite passes under `-race`.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -16,6 +16,7 @@
 package yaml
 
 import (
+	"context"
 	"encoding"
 	"encoding/base64"
 	"fmt"
@@ -325,6 +326,12 @@ type decoder struct {
 	aliasDepth  int
 
 	mergedFields map[interface{}]bool
+
+	// ctx is handed to types implementing UnmarshalerWithContext.
+	// UnmarshalYAML is given a node and nothing else, so a type needing state
+	// belonging to the decode rather than to its own node has no other way to
+	// reach it. Nil means no context was supplied.
+	ctx context.Context
 }
 
 var (
@@ -364,6 +371,24 @@ func (d *decoder) terror(n *Node, tag string, out reflect.Value) {
 
 func (d *decoder) callUnmarshaler(n *Node, u Unmarshaler) (good bool) {
 	err := u.UnmarshalYAML(n)
+	if e, ok := err.(*TypeError); ok {
+		d.terrors = append(d.terrors, e.Errors...)
+		return false
+	}
+	if err != nil {
+		fail(err)
+	}
+	return true
+}
+
+// callUnmarshalerWithContext invokes the context-aware UnmarshalYAML method,
+// handling errors exactly as callUnmarshaler does.
+func (d *decoder) callUnmarshalerWithContext(n *Node, u UnmarshalerWithContext) (good bool) {
+	ctx := d.ctx
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	err := u.UnmarshalYAML(ctx, n)
 	if e, ok := err.(*TypeError); ok {
 		d.terrors = append(d.terrors, e.Errors...)
 		return false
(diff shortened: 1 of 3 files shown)


ccoVeille:
Thanks for working on this.

I’m a bit hesitant about this suggestion.

Our v3 branch is intended to remain stale, except for security fixes. We recently accepted a PR to backport the removal of a testing-only dependency, but that did not change the v3 runtime code itself.

I understand and agree with your point that users should not be expected to move to v4 while it is still in RC. At the same time, I don’t want to introduce additional changes in v3 that would effectively require us to maintain two active branches in parallel.

So at this stage, I’m likely to decline this PR for the v3 branch.

That said, I’d be interested in hearing feedback from other maintainers

reuvenharrison:
Understood, and thank you for the quick answer. The v3 policy is clear and I would rather not be the reason two branches need maintaining.

Reopened against `main` as #390. It is the same change on the `internal/libyaml` layout, which is also what left #186 conflicting, so #390 carries that forward rather than competing with it.

Happy for this one to be closed.

One note for whoever picks up #390, since it is the open question rather than the code: the thread on #187 stalled on whether this should be an option instead of an interface, and #390 takes the interface position with the reasoning written out. I am happy to be overruled and rework it as an option; only the entry points change.

ccoVeille:
Let's close this one.

We will review the other one

PR #390, opened 2026-08-04
Add UnmarshalerWithContext

### LINKED ISSUE #187: Provide context to Unmarshaler

reuvenharrison:
Closes #187. Opened against `main` per @ccoVeille's note on #389 that v3 takes no runtime changes; #389 is the same change on the v3 layout and can be closed in favour of this.

This also carries forward the unmarshal half of #186, which has conflicted since the move into `internal/libyaml`.

A type implementing `UnmarshalerWithContext` receives the context its decode was started with:

```go
type UnmarshalerWithContext interface {
	UnmarshalYAML(ctx context.Context, n *Node) error
}
```

supplied by `UnmarshalWithContext`, `Decoder.DecodeWithContext` and `Node.DecodeWithContext`. A type implementing both interfaces gets this one. A type implementing neither is untouched, and decoding a context-only type without one passes `context.Background()`.

## On the design question in #187

The thread stalled on whether this should be an option rather than an interface, so this takes a position and I am happy to be overruled.

An option means holding a `context.Context` in the options struct, which @ccoVeille raised as an anti-pattern. Here the context is a first argument at every public entry point, and the constructor holds it only for the length of a call:

```go
func (l *Loader) LoadWithContext(ctx context.Context, v any) error {
	prev := l.constructor.Ctx
	l.constructor.Ctx = ctx
	defer func() { l.constructor.Ctx = prev }()
	return l.Load(v)
}
```

Nothing is stored across calls and no decode observes another's context. If the option shape is still preferred, the dispatch and plumbing are identical and only the entry points change; say so and I will rework it.

## Tests

The context reaching an unmarshaler, a context-only type decoded without one getting `context.Background()`, and twenty concurrent decodes each asserting it observed its **own** context. The existing suite passes, including under `-race`.

### DIFF
--- a/decode_context_test.go
+++ b/decode_context_test.go
@@ -1,0 +1,73 @@
+package yaml_test
+
+import (
+	"context"
+	"sync"
+	"testing"
+
+	yaml "go.yaml.in/yaml/v4"
+)
+
+type ctxKey struct{}
+
+// withCtx records the context it was decoded with, and the value it found.
+type withCtx struct {
+	Name string
+	seen string
+}
+
+func (w *withCtx) UnmarshalYAML(ctx context.Context, n *yaml.Node) error {
+	type bis withCtx
+	if err := n.Decode((*bis)(w)); err != nil {
+		return err
+	}
+	if v, ok := ctx.Value(ctxKey{}).(string); ok {
+		w.seen = v
+	}
+	return nil
+}
+
+func TestUnmarshalWithContext(t *testing.T) {
+	ctx := context.WithValue(context.Background(), ctxKey{}, "from-caller")
+	var w withCtx
+	if err := yaml.UnmarshalWithContext(ctx, []byte("name: n\n"), &w); err != nil {
+		t.Fatal(err)
+	}
+	if w.Name != "n" || w.seen != "from-caller" {
+		t.Fatalf("got name=%q seen=%q", w.Name, w.seen)
+	}
+}
+
+// Unmarshal still works on a type implementing only the context form; it just
+// receives a background context.
+func TestUnmarshalWithoutContext(t *testing.T) {
+	var w withCtx
+	if err := yaml.Unmarshal([]byte("name: n\n"), &w); err != nil {
+		t.Fatal(err)
+	}
+	if w.Name != "n" || w.seen != "" {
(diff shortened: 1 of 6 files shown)


ccoVeille:
Thanks for working on this! I have a slightly different approach in mind for this feature.

I've opened another PR with an alternative implementation:

- https://github.com/yaml/go-yaml/pull/392

Would you mind taking a look and sharing your thoughts?

reuvenharrison:
Closing in favour of #392, as above.

PR #395, opened 2026-08-15
Recognize supplementary-plane runes as printable

sergiobuilds:
## Summary

`isPrintable` (`internal/libyaml/scanner.go`) only recognizes 1- to 3-byte UTF-8 lead bytes as potentially printable. Every character in the Unicode supplementary planes (U+10000-U+10FFFF, encoded as 4-byte UTF-8 — this includes most emoji) is therefore classified as a "special character" by `Emitter.analyzeScalar`.

That disables `block_allowed` and `single_quoted_allowed` for any scalar containing such a character, so the emitter falls back to double-quoted style with an escaped rune (`\U0001F6A7`) even when the scalar would otherwise qualify for literal/folded block style — e.g. an ordinary multiline string that happens to contain an emoji.

Reported against yq: [mikefarah/yq#2797](https://github.com/mikefarah/yq/issues/2797). yq uses this library (`go.yaml.in/yaml/v4`) for its output encoding, and the actual decision lives here in `isPrintable`, not in yq's own code — `Style` is left unset when yq builds the node, so the library picks the style.

## Fix

Add the missing 4-byte UTF-8 lead-byte range (`0xF0`-`0xF4`, matching `U+10000`-`U+10FFFF`) to `isPrintable`.

## Reproduction

```go
data := map[string]string{
    "longDescription": "Multiline file containing\n\n🚧\n\nA \"weird\" utf8 char?\n",
}
out, _ := yaml.Marshal(data)
fmt.Println(string(out))
```

Before:
```yaml
longDescription: "Multiline file containing\n\n\U0001F6A7\n\nA \"weird\" utf8 char?\n"
```

After:
```yaml
longDescription: |
    Multiline file containing

    🚧

    A "weird" utf8 char?
```

## Test Plan

- Added three `char-predicate` cases to `internal/libyaml/testdata/scanner.yaml`:
  - U+1F6A7 (4-byte lead byte `0xF0`) — printable
  - U+10FFFF, the highest valid code point (lead byte `0xF4`) — printable
  - `0xF5`, the first invalid lead byte past the valid 4-byte range — **not** printable (guards against widening the range too far)
- Confirmed all three fail on unpatched `main` (`isPrintable("🚧", 0) = false, want true`, etc.) and pass with the fix.
- `make check` (fmt + lint + test, including the full yaml-test-suite): all pass, no regressions. `go test ./...` also run directly: `ok` across all packages.
- Verified the end-to-end repro above by hand against both the unpatched and patched library.

### DIFF
--- a/internal/libyaml/scanner.go
+++ b/internal/libyaml/scanner.go
@@ -627,7 +627,8 @@ func isPrintable(b []byte, i int) bool {
 		(b[i] == 0xEE) ||
 		(b[i] == 0xEF && // #xE000 <= . <= #xFFFD
 			!(b[i+1] == 0xBB && b[i+2] == 0xBF) && // && . != #xFEFF
-			!(b[i+1] == 0xBF && (b[i+2] == 0xBE || b[i+2] == 0xBF))))
+			!(b[i+1] == 0xBF && (b[i+2] == 0xBE || b[i+2] == 0xBF))) ||
+		(b[i] >= 0xF0 && b[i] <= 0xF4)) // #x10000 <= . <= #x10FFFF
 }
 
 // Check if the character at the specified position is NUL.
--- a/internal/libyaml/testdata/scanner.yaml
+++ b/internal/libyaml/testdata/scanner.yaml
@@ -686,6 +686,33 @@
     data: [0x19]
     want: false
 
+- char-predicate:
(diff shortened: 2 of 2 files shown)


ccoVeille:
There was a pending fix for this

- https://github.com/yaml/go-yaml/pull/357

Could you please check it, and make feedbacks on it?

Issue #354, opened 2026-05-22
Emoji/unicode is incorrectly escaped


TomWright:
I'm the maintainer of [dasel](https://github.com/TomWright/dasel).

I've had an [issue raised](https://github.com/TomWright/dasel/issues/437) highlighting that emoji's are converted to escaped unicode whereas it is not for JSON, TOML etc.

I have a PR for a local fix here: https://github.com/TomWright/dasel/pull/549.
I can introduce a hacky fix for this (PR above) and I see a related PR on [go-yaml/yaml](https://github.com/go-yaml/yaml/pull/738/changes) but that PR is no longer maintained.

Is this something we could get fixed at the source?

Thanks

---

Maintainer edits:

Fixes https://github.com/mikefarah/yq/issues/2797

- https://github.com/mikefarah/yq/issues/2797

ccoVeille:
I feel like the fix could be easy by using the "classic range of emoji" in [isPrintable](https://github.com/yaml/go-yaml/blob/db7ba150748579b6ba5f8b894665fceacef42de9/internal/libyaml/scanner.go#L621)

But we have to keep in mind that Unicode is not fun, and emojis are worst

For example
 🏳️‍🌈 = 🏳️ +\x{200D}	Zero Width Joiner (ZWJ)+ 🌈

but, 👩🏼‍❤️‍💋‍👨🏾 =
 
1 👩	\x{1F469}	Woman	
2	🏼	\x{1F3FC}	Type-3 Skin Tone	
3	‍	\x{200D}	Zero Width Joiner (ZWJ)
4	❤️	\x{2764}\x{FE0F}	Heavy Black Heart + Variation Selector-16
5	‍	\x{200D}	Zero Width Joiner (ZWJ)
6	💋	\x{1F48B}	Kiss Mark
7	‍	\x{200D}	Zero Width Joiner (ZWJ)	
8	👨🏾

We can think about a regex or something like that, but it might be incomplete, and new emojis and emojis variations are often added

that's why there are lib out there that are maintained

https://github.com/forPelevin/gomoji
https://github.com/rivo/uniseg (that uses https://github.com/mattn/go-runewidth )

But these would bring dependencies

I feel like we can:
- provide something basic that can detect most of "simple unicode" with a regexp
- we can think about having a way to extend the go-yaml capabilities with a plugin. It will expect a function to validate what is an emoji, for people who want "full emoji support" by using an external lib to catch them all

ccoVeille:
[...]

PR #356, opened 2026-05-24
Fix Unicode emoji sequences support

ccoVeille:
This PR fixes the way Unicode and especially emojis are supported

The following YAML

```yaml
- 😀
- 👍🏽
- 🇺🇸
- 🏳️‍🌈
- 👩🏼‍❤️‍💋‍👨🏾
```

was computed as

```yaml
- {token: STREAM-START}
- {token: BLOCK-SEQUENCE-START}
- {token: BLOCK-ENTRY}
- {token: SCALAR, value: "\U0001F44D"}
- {token: BLOCK-ENTRY}
- {token: SCALAR, value: "\U0001F44D\U0001F3FD"}
- {token: BLOCK-ENTRY}
- {token: SCALAR, value: "\U0001F1FA\U0001F1F8"}
- {token: BLOCK-ENTRY}
- {token: SCALAR, value: "\U0001F3F3️‍\U0001F308"}
- {token: BLOCK-ENTRY}
- {token: SCALAR, value: "\U0001F469\U0001F3FC‍❤️‍\U0001F48B‍\U0001F468\U0001F3FE"}
- {token: BLOCK-END}
- {token: STREAM-END}
```

and rendered as

```yaml
- "\U0001F44D"
- "\U0001F44D\U0001F3FD"
- "\U0001F1FA\U0001F1F8"
- "\U0001F3F3️‍\U0001F308"
- "\U0001F469\U0001F3FC‍❤️‍\U0001F48B‍\U0001F468\U0001F3FE"
```

With this PR, they are computed as

```yaml
- {token: STREAM-START}
- {token: BLOCK-SEQUENCE-START}
- {token: BLOCK-ENTRY}
- {token: SCALAR, value: 👍}
- {token: BLOCK-ENTRY}
- {token: SCALAR, value: 👍🏽}
- {token: BLOCK-ENTRY}
- {token: SCALAR, value: 🇺🇸}
- {token: BLOCK-ENTRY}
- {token: SCALAR, value: 🏳️‍🌈}
- {token: BLOCK-ENTRY}
- {token: SCALAR, value: 👩🏼‍❤️‍💋‍👨🏾}
- {token: BLOCK-END}
- {token: STREAM-END}
```

and rendered as

```yaml
- 😀
- 👍🏽
- 🇺🇸
- 🏳️‍🌈
- 👩🏼‍❤️‍💋‍👨🏾
```

Fixes #354

PR #397, opened 2026-08-16
Treat code points above the BMP as printable

vjymisal0:
Relates to #354.

### The problem

`isPrintable` walks the UTF-8 lead byte ranges up to `0xEF` but has no case for `0xF0`-`0xF4`, so every code point above U+FFFF is reported as unprintable. Two things follow from that in `yaml_emitter_analyze_scalar`: the character is escaped in double quoted style, and `special_characters` clears `block_allowed`, so a multiline string containing one can no longer use literal style.

```go
data, _ := yaml.Marshal(map[string]string{"desc": "one\n\U0001F6A7\ntwo\n"})
```

before:

```yaml
desc: "one\n\U0001F6A7\ntwo\n"
```

after:

```yaml
desc: |
    one
    🚧
    two
```

A BMP symbol such as ✅ (U+2705) already produced the literal block, which is what makes it look as though only "some" characters are affected - the dividing line is exactly whether the character needs four bytes in UTF-8.

### The fix

`c-printable` in the spec is:

```
[1] c-printable ::= #x9 | #xA | #xD | [#x20-#x7E]        /* 8 bit */
                  | #x85 | [#xA0-#xD7FF] | [#xE000-#xFFFD] /* 16 bit */
                  | [#x10000-#x10FFFF]                     /* 32 bit */
```

The last line is the missing one, and `[#x10000-#x10FFFF]` is exactly the set encoded by lead bytes `0xF0`-`0xF4`, so it is one more case in the existing chain. This mirrors how the other ranges are matched on the lead byte, and needs no lookahead, so it doesn't add any new indexing beyond the end of the buffer.

Worth noting for the discussion in #354: no emoji-specific handling is needed for this. ZWJ (U+200D), variation selectors (U+FE0F) and the skin tone modifiers are already printable via the existing ranges - the sequences in that thread were rendering badly only because their base characters live above the BMP. Anything that is a valid code point above U+FFFF is printable, so there is nothing to keep in step with new Unicode releases.

### Tests

- `TestScalarStyleAboveBMP` in `yaml_test.go` covers U+10000, U+10FFFF and an astral plane emoji inline, plus multiline scalars with an astral and a BMP character, and round trips each one.
- Three `isPrintable` cases in `internal/libyaml/testdata/scanner.yaml` for the boundaries.

All of the new tests fail without the change. `go test ./...` passes, including the `yaml-test-suite` conformance run (`go test ./yts/...`) after `make get-test-data`.

The only failure I see locally is `cmd/go-yaml`'s `TestCLI`, which fails the same way on an unmodified checkout here - it shells out with a path that loses its separators on Windows - so it looks unrelated to this change.

### DIFF
--- a/internal/libyaml/scanner.go
+++ b/internal/libyaml/scanner.go
@@ -627,7 +627,8 @@ func isPrintable(b []byte, i int) bool {
 		(b[i] == 0xEE) ||
 		(b[i] == 0xEF && // #xE000 <= . <= #xFFFD
 			!(b[i+1] == 0xBB && b[i+2] == 0xBF) && // && . != #xFEFF
-			!(b[i+1] == 0xBF && (b[i+2] == 0xBE || b[i+2] == 0xBF))))
+			!(b[i+1] == 0xBF && (b[i+2] == 0xBE || b[i+2] == 0xBF))) ||
+		(b[i] >= 0xF0 && b[i] <= 0xF4)) // #x10000 <= . <= #x10FFFF
 }
 
 // Check if the character at the specified position is NUL.
--- a/internal/libyaml/testdata/scanner.yaml
+++ b/internal/libyaml/testdata/scanner.yaml
@@ -675,6 +675,21 @@
     data: [0xC2, 0xA0]
 
 - char-predicate:
+    name: Is printable - U+10000 (lowest 4 byte sequence)
+    func: isPrintable
+    data: [0xF0, 0x90, 0x80, 0x80]
+
+- char-predicate:
+    name: Is printable - U+1F6A7 (astral plane emoji)
+    func: isPrintable
+    data: [0xF0, 0x9F, 0x9A, 0xA7]
+
+- char-predicate:
+    name: Is printable - U+10FFFF (highest code point)
+    func: isPrintable
+    data: [0xF4, 0x8F, 0xBF, 0xBF]
+
+- char-predicate:
     name: Is printable - null
     func: isPrintable
     data: [0x00]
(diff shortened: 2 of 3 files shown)


ccoVeille:
There was a pending fix for this

- https://github.com/yaml/go-yaml/pull/357

Could you please check it, and make feedbacks on it?

vjymisal0:
Took a look at #357, thanks for the pointer @ccoVeille.

Technical comparison of the two `isPrintable` fixes for the 4-byte range:

- This PR (#397): `b[i] >= 0xF0 && b[i] <= 0xF4` — lead-byte only, matching how every other range in the existing function is checked (none of them validate continuation bytes either).
- #357: `(b[i]==0xF0 && b[i+1] in [0x90,0xBF]) || (0xF0<b[i]<0xF4 && b[i+1] in [0x80,0xBF]) || (b[i]==0xF4 && b[i+1] in [0x80,0x8F])` — also bounds the second byte, so it correctly rejects overlong encodings under 0xF0 (already representable in 3 bytes) and sequences under 0xF4 that decode above U+10FFFF (not a valid Unicode scalar).

#357 is the more correct fix — it actually validates against the U+10000..U+10FFFF boundary instead of just gating on the lead byte, so it won't misclassify a handful of byte sequences that are structurally 4-byte UTF-8 but out of the valid Unicode range. My version is simpler and stylistically consistent with the rest of the function (which also doesn't validate trailing bytes elsewhere), but that consistency isn't a correctness argument in this case.

#357 also has the broader scope (fixes the linked yq issue, supersedes #356) and was opened first, so I'd rather not duplicate effort — happy to close this one in favor of #357 once its second-byte-bounded version lands. Let me know if you'd like me to close it now or keep it open until #357 merges.

ccoVeille:
I would prefer if you could close it.

Feel free to comment #357 with any feedback you may have

PR #398, opened 2026-08-20
Restores indentation of closing single-quote

sidsri14:
## Summary

A single-quoted scalar whose value ends with one or more line breaks loses the indentation of its closing `'`: the closing quote is emitted at column 0 instead of at the content column.

```yaml
' Get quacked.

  - Duck

'
```

should be emitted as:

```yaml
' Get quacked.

  - Duck

  '
```

## Root cause

`yaml_emitter_write_single_quoted_scalar` in libyaml writes an indentation before the closing quote whenever the scalar ends with line breaks:

```c
    if (breaks)
        if (!yaml_emitter_write_indent(emitter)) return 0;
    if (!yaml_emitter_write_indicator(emitter, "'", 0, 0, 0)) return 0;
```

That block was dropped in the Go port of the emitter. This restores it, matching libyaml behaviour.

## Test plan

Added `TestEmitSingleQuotedScalarClosingQuoteIndent`, which fails before this change and passes after. `go test ./internal/libyaml/` passes; the remaining failures in `go test ./...` (`cmd/go-yaml` binary execution on Windows paths, and `yts` requiring `make get-test-data`) are pre-existing on this checkout and unrelated to this change.

Motivated by mikefarah/yq#2819.

## AI assistance

This patch and its test were prepared with assistance from an AI coding assistant.

### DIFF
--- a/internal/libyaml/emitter_test.go
+++ b/internal/libyaml/emitter_test.go
@@ -28,6 +28,48 @@ func TestEmitter(t *testing.T) {
 	})
 }
 
+func TestEmitSingleQuotedScalarClosingQuoteIndent(t *testing.T) {
+	cases := []struct {
+		name  string
+		value string
+		want  string
+	}{
+		{
+			name:  "multi-line scalar ending with a line break",
+			value: " Get quacked.\n- Duck\n",
+			want:  "' Get quacked.\n\n  - Duck\n\n  '\n",
+		},
+		{
+			name:  "two short lines ending with a line break",
+			value: "a\nb\n",
+			want:  "'a\n\n  b\n\n  '\n",
+		},
+	}
+
+	for _, tc := range cases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			events := []Event{
+				NewStreamStartEvent(UTF8_ENCODING),
+				NewDocumentStartEvent(nil, nil, true),
+				NewScalarEvent(nil, nil, []byte(tc.value), true, true, SINGLE_QUOTED_SCALAR_STYLE),
+				NewDocumentEndEvent(true),
+				NewStreamEndEvent(),
+			}
+
+			emitter := NewEmitter()
+			emitter.SetIndent(2)
+			var output []byte
+			emitter.SetOutputString(&output)
+			for i := range events {
+				err := emitter.Emit(&events[i])
+				assert.NoErrorf(t, err, "Emit() error: %v", err)
+			}
+			assert.Equal(t, tc.want, string(output))
+		})
+	}
+}
+
 func TestEmitFoldedScalarNoExtraNewline(t *testing.T) {
 	cases := []struct {
 		name  string
--- a/internal/libyaml/emitter_test.go
+++ b/internal/libyaml/emitter_test.go
@@ -28,48 +28,6 @@ func TestEmitter(t *testing.T) {
 	})
 }
 
-func TestEmitSingleQuotedScalarClosingQuoteIndent(t *testing.T) {
-	cases := []struct {
-		name  string
-		value string
-		want  string
-	}{
-		{
-			name:  "multi-line scalar ending with a line break",
-			value: " Get quacked.\n- Duck\n",
-			want:  "' Get quacked.\n\n  - Duck\n\n  '\n",
-		},
-		{
-			name:  "two short lines ending with a line break",
-			value: "a\nb\n",
-			want:  "'a\n\n  b\n\n  '\n",
-		},
-	}
-
-	for _, tc := range cases {
-		tc := tc
-		t.Run(tc.name, func(t *testing.T) {
-			events := []Event{
-				NewStreamStartEvent(UTF8_ENCODING),
-				NewDocumentStartEvent(nil, nil, true),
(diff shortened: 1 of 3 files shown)


internal/libyaml/emitter_test.go
```diff
@@ -28,6 +28,48 @@ func TestEmitter(t *testing.T) {
...
+			var output []byte
+			emitter.SetOutputString(&output)
+			for i := range events {
+				err := emitter.Emit(&events[i])
+				assert.NoErrorf(t, err, "Emit() error: %v", err)
+			}
+			assert.Equal(t, tc.want, string(output))
+		})
+	}
+}
```
ccoVeille, line 71: Could you please move this test out of this file and use the emitter.yaml file to test this?

As it's a YAML project under the GitHub YAML organization we try to use YAML everywhere

ccoVeille:
The error in the CI is caused by golangci-lint that needs to be bumped to 2.13.X first version that supports Go 1.27

So it's OK

Follow-up of this will be done with

- #400

ccoVeille:
Thanks you for catching this

```
yaml_emitter_write_single_quoted_scalar in libyaml writes an indentation before the closing quote whenever the scalar ends with line breaks:

    if (breaks)
        if (!yaml_emitter_write_indent(emitter)) return 0;
    if (!yaml_emitter_write_indicator(emitter, "'", 0, 0, 0)) return 0;
That block was dropped in the Go port of the emitter. This restores it, matching libyaml behaviour.
```

sidsri14:
Rebased this branch on top of #401 (golangci-lint v2.13.1 bump), so the `test (1.27.0)` job — which previously failed with the known golangci-lint export-data error — should now pass. Verified locally: `go test ./internal/libyaml/` green, `git diff --check` clean.

internal/libyaml/emitter_test.go
sidsri14: Done in 4a83a28. The regression cases now live in `internal/libyaml/testdata/emitter.yaml` as two `emit` entries (values ` Get quacked.
- Duck
` and `a
b
`, each asserting the closing quote is indented via `want: "  '"`), and the Go test was removed from `emitter_test.go`. Verified the cases fail without the fix and pass with it (`go test ./internal/libyaml/`).

ccoVeille:
@sidsri14 could you rebase as #401 was merged

sidsri14:
Rebased onto current `main` (which now includes the merged #401 golangci-lint v2.13.2 bump and #400 reflect.Pointer change). The branch is now two commits:

- `Restores indentation of closing single-quote` — the `emitter.go` fix
- `Move closing-quote indent test into emitter.yaml` — the YAML-driven regression cases

The two previously-carried commits (`Bump golangci-lint v2.13.1`, `Use reflect.Pointer`) were dropped since they're now in upstream. Verified locally: `go test ./internal/libyaml/` green and `go build ./...` clean. The only failing suite is `yts`, which needs the external `make get-test-data` fixtures (pre-existing requirement).

Conflict is cleared — CI is re-running on the new head.

ccoVeille (approved):

ingydotnet:
https://play.yaml.com/?show=gdhflLnJPpxHUrsS#JyBHZXQgcXVhY2tlZC4KCiAgLSBEdWNrCgon

```
' Get quacked.

  - Duck

'
```

is valid yaml. 
From my PoV there isn't an exact proper way to format this as long as it is valid YAML.

What problem are you hitting or trying to resolve?

Also, use the local CLI for reporting issues. Just run `make cli` then
```
~/src/go-yaml main $ ./go-yaml -Y
' Get quacked.

  - Duck

'
' Get quacked.

  - Duck

'
~/src/go-yaml main $ 

```

ingydotnet:
This however is invalid YAML output:
```
~/src/go-yaml main $ ./go-yaml -Y
- ' Get quacked.

  - Duck

'
- ' Get quacked.

  - Duck

'
```

https://play.yaml.com/?show=gdhflLnJPpxHUrsS#LSAnIEdldCBxdWFja2VkLgoKICAtIER1Y2sKCicK

As is this:

```
~/src/go-yaml main $ ./go-yaml -Y
foo: {x:
y}
foo: {x: y}
```

https://play.yaml.com/?show=gdhflLnJPpxHUrsS#Zm9vOiB7eDoKeX0K

neither of those inputs should load, but they both do. The first one also outputs bad YAML, but the second one does not.

ingydotnet:
I will try these with your patches here.

ingydotnet:
```
~/src/go-yaml/patched fix-single-quote-closing-indent $ ./go-yaml -Y
- ' Get quacked.

  - Duck

'
- ' Get quacked.

  - Duck

  '
~/src/go-yaml/patched fix-single-quote-closing-indent $ ./go-yaml -Y
' Get quacked.

  - Duck

'
' Get quacked.

  - Duck

  '
```
The first example is invalid YAML, which is not something we're worried about here. We're just worried about the output and you fix it nicely. 
The second one loads valid YAML and to be honest I don't think it should be indented the way you do it.
So for top level scalar nodes, I don't think we have a change to make here. 
But you've inadvertently fixed a critical bug. 
I would add more testing on single quoted scalars in various collection contexts. 
I can come up with some examples for you if you want.

PR #402, opened 2026-08-23
Draft: Implement support for loader/dumper custom marshallers

wrouesnel:
Foreword: I've been using this successfully for an internal project where it accomplishes the goal - namely that it is possible, with some support code, to define a full system to directly decode Go struct's which have interface members which get specialized by discriminant tags.

The patch here is a rebase of the code I have been using (which is here: https://github.com/wrouesnel/yaml.go-yaml) which I am now trying to orient towards being merged upstream since it has proven quite useful.

Basically you can have something like:

```yaml
object:
  members:
  - type: a
  - type: b
  - type: c
```

be seamlessly decoded into a Go struct defined to look like:

```go
type MemberInterface interface {

}

type Object struct {
  Members []MemberInterface
}
```

provided you can do the discriminant (which is an exercise left to the implementor - I stick a Type() field on all my objects and register decoders use that as a discriminant, there's no reason at all you couldn't just use reflection - but nothing is prescribed in this patch, just the support structures to make it happen).

_Does the direction of changes here look good?_ (this draft is a reminder to myself to set aside some time to rebase and catch up to the main branch).

# Patch Desc
Add support for configuring external-to-type custom marshallers, specified as Options, for loading/dumping.

This feature specifically enables configuring YAML marshalling for types such as interfaces which lack a concrete implementation.

The primary use case is to allow type discrimination of struct's which contain interface members during unmarshalling: an interface-specific decoder and encoder can be specified, which then selects and builds the correct concrete type which can then be unmarshalled using it's own YAML unmarshaller (or built in types).

To support this use, the `SubstituteError` type is added which can be returned from a custom unmarshaler to replace the `libyaml.Node` being used for unmarshalling. This allows a custom unmarshaller to insert, modify or otherwise correct the YAML subtree which is necessary to unmarshal into the concrete type.


ccoVeille:
Thanks for the PR. I know it's still a draft, but the new code doesn’t include any tests yet. That makes it hard to evaluate and leaves the behavior effectively unverified. Could you please add tests for the changes?

I'm also concerned about the added complexity, especially since I’m not sure there’s a real need for it yet. Do you have any real-world examples of code that would use this branch's functionality, or evidence that this is solving a current problem?

wrouesnel:
So for an example of the real-world use, there's this package here: https://github.com/wrouesnel/interface_convention (go test exercises the code)

Which is why I developed this - this functionality is the minimum needed so you can implement discriminant-based deserialization to _interface_ types - i.e.

```go
type AbstractSomething interface {
   Type() string
}

type Something struct {
  Contents []AbstractSomethings
}
```

Being able to register custom unmarshaller's let's you detect the type, parse the discriminant and then switch the unmarshalling type to an actual concrete implementation - i.e.

```go
type Thing1 struct {}

func (t *Thing1) Type() string {
  return "type-1"
}

type Thing2 struct {}

func (t *Thing2) Type() string {
  return "type-1"
}
```

so you can have this serialize and deserialize to something like:

```yaml
something:
  contents:
  - type: type-2
  - type: type-1
```

which is the pattern `interface_convention` implements (for my specific use case). It _also_ lets you have a nice defaults system - i.e. in that library the defaults can just be stored as YAML files in the program via embed.FS and get seamlessly added as merge keys at deserialization time, which makes it very ergonomic to use.

All of this makes it _very nice_ to write a configuration system based on YAML, or parse a lot of existing YAML types which use exactly this sort of a pattern.

ccoVeille:
Thanks for your detailed reply.

For now, I'm still doubtful about the need to add so much complexity in the code for something that might not be used.

There is nothing like this in encoding/json either. So it sounds something a library based on yaml/go-yaml could do if there is a need. So something out of the lib itself.

This is non-definitive point of view. Also, some other maintainers could provide feedback that would change my mind.

cc @colinjlacy @ingydotnet 

Also other people could come here and share their thoughts.

Maybe @ingydotnet might be interested to write this as a plugin. Or it could be a way to showcase what plugin could do.

ingydotnet:
This does seem like plugin territory.

wrouesnel:
So I will say I'm a little unclear what defines a plugin, looking at the code - implementation wise they look very similar to options.

PR #372, opened 2026-06-26
Add ErrorFmt plugin API

ingydotnet:
Add a plugin API for YAML error formatting and ship official v3/v4 error formatting plugins.

This makes error string formatting configurable while preserving structured error values such as `*yaml.LoadError`, `*yaml.DumpError`, and `*yaml.LoadErrors`.

The branch adds:

- `yaml.ErrorFmtPlugin`
- `plugin/errfmt/v3`
- `plugin/errfmt/v4`
- YAML config support through `yaml.OptsYAML`
- plugin user guides under `docs/plugin/`

`WithV3Defaults()` now installs the v3 formatter. Bare/default loading and dumping, `WithV2Defaults()`, and `WithV4Defaults()` use the v4 formatter.

Add `yaml.ErrorFmtPlugin`:

```go
type ErrorFmtPlugin interface {
    FormatLoadError(err *yaml.LoadError) string
    FormatDumpError(err *yaml.DumpError) string
}
```

Plugins are registered with the existing `yaml.WithPlugin(...)` option.

Add `go.yaml.in/yaml/v4/plugin/errfmt/v3`.

This plugin renders v3-compatible strings:

```text
yaml: line 1: message
yaml: message
```

Add `go.yaml.in/yaml/v4/plugin/errfmt/v4`.

This plugin renders the current structured v4 style:

```text
go-yaml load error in scanner at L1.C8: message
go-yaml dump error in representer: message
```

The v4 formatter supports:

- `WithPositionStyle(PositionShort)`
- `WithPositionStyle(PositionLong)`
- `WithPositionStyle(PositionLine)`
- `WithLoadTemplate(...)`
- `WithDumpTemplate(...)`
- `WithTemplate(...)` as a load-template alias

Thread configured load-error formatting through:

- reader
- scanner
- parser
- composer
- resolver
- constructor

This keeps load errors structured while allowing `Error()` output to be selected by defaults or plugins.

Thread configured dump-error formatting through:

- representer
- serializer
- emitter error conversion
- writer error conversion

Exact user-created `yaml.NewDumpError(...)` values are not double-wrapped, but they still receive the active formatter before being returned.

Extend `yaml.OptsYAML` plugin config:

```yaml
plugin:
  errfmt:
    v3:
```

```yaml
plugin:
  errfmt:
    v4:
      position: long
      load-template: '{{.Stage}} at {{pos .Mark}}: {{.Message}}'
      dump-template: '{{.Stage}}: {{.Message}}'
```

[...]

PR #403, opened 2026-08-23
Report unhashable map keys with a clear error

### LINKED ISSUE #378: panic: runtime error: hash of unhashable type []interface {}
// Minimal reproduction (found by fuzzing)

```go

package main

import (
	"fmt"

	"gopkg.in/yaml.v3"
)

func main() {
	body := "<<:\n? -"
	fmt.Printf("YAML body:\n%s\n---\n", body)
	var out map[string]any
	yaml.Unmarshal([]byte(body), &out)
}
```

hdimer:
Fixes #378.

## The problem

A mapping that carries both a merge key (`<<`) and an unhashable key (a sequence or a mapping used as a key) reaches the merge-dedup bookkeeping map, and `m[key]` on a `map[any]bool` panics. `setPossiblyUnhashableKey` / `getPossiblyUnhashableKey` already recover from that panic, but they re-wrapped the raw Go runtime text with `fmt.Errorf("%v", err)`, so the reported repro from #378:

```go
var out map[string]any
yaml.Unmarshal([]byte("<<:\n? -"), &out)
```

produced

```
go-yaml load error in constructor at L2.C3: runtime error: hash of unhashable type []interface {}
```

`mapping()` already had a message for this exact error class a few hundred lines up, so the same key shape reported two different things depending on which path it took.

## The change

`mapping()`'s message moves into `failUnhashableKey`, next to `failWantMap`, and all three sites call it. The input above now reports:

```
go-yaml load error in constructor at L2.C3: cannot use '[]interface {}{interface {}(nil)}' as a map key; try decoding into yaml.Node
```

No behavior change for the pre-existing path: the helper builds the same `Mark` from the node it is given, and the existing `Slice as map key` / `Map as map key` cases pass untouched.

The two recover paths now discard the panic value. Both are guarded against a nil map by their call sites (`constructor.go:664` and the `make(map[any]bool)` at `:791`), and a nil-map read is legal in Go, so the only recoverable panic reachable there is the unhashable-key hash.

## Tests

Two cases in `testdata/unmarshal_errors.yaml`, one per helper: the first is the verbatim repro from #378 and lands in `setPossiblyUnhashableKey` via `merge()`; the second lands in `getPossiblyUnhashableKey` via `mapping()` with `mergedFields` set. Both fail on unpatched code with the raw runtime string. The `type:` line on the first case is load-bearing: with the default `any` target the key routes through the pre-existing check in `mapping()` instead.

`make check` passes (`fmt`, `tidy`, `lint` at the pinned golangci-lint v2.8.0, `test` including the yaml-test-suite at the known 1383/225 baseline).

## Note

#372 touches the same three sites, converting `formatConstructorError` and `failWantMap` into methods and rewriting every `want:` string in `unmarshal_errors.yaml`. It keeps `fmt.Errorf("%v", err)` in the recover blocks, so it does not overlap with this fix, but it will conflict textually. Happy to rebase on top of it if you would rather land that first.

---
Used AI assistance on this; I reviewed and tested the change myself.

### DIFF
--- a/internal/libyaml/constructor.go
+++ b/internal/libyaml/constructor.go
@@ -673,10 +673,7 @@ func (c *Constructor) mapping(n *Node, out reflect.Value) (good bool) {
 				kkind = k.Elem().Kind()
 			}
 			if kkind == reflect.Map || kkind == reflect.Slice {
-				Fail(formatConstructorError(
-					fmt.Errorf("cannot use '%#v' as a map key; try decoding into yaml.Node", k.Interface()),
-					Mark{Line: n.Content[i].Line, Column: n.Content[i].Column},
-				))
+				failUnhashableKey(k.Interface(), n.Content[i])
 			}
 			e := reflect.New(et).Elem()
 			if c.Construct(n.Content[i+1], e) || n.Content[i+1].ShortTag() == nullTag && (mapIsNew || !out.MapIndex(k).IsValid()) {
@@ -853,6 +850,15 @@ func isMerge(n *Node) bool {
 func failWantMap(n *Node) {
 	Fail(formatConstructorError(
 		fmt.Errorf("map merge requires map or sequence of maps as the value"),
+		Mark{Line: n.Line, Column: n.Column},
+	))
+}
+
+// failUnhashableKey panics with an error message for keys that cannot be used
+// as Go map keys.
+func failUnhashableKey(key any, n *Node) {
+	Fail(formatConstructorError(
+		fmt.Errorf("cannot use '%#v' as a map key; try decoding into yaml.Node", key),
 		Mark{Line: n.Line, Column: n.Column},
 	))
 }
@@ -1108,11 +1114,8 @@ func settableValueOf(i any) reflect.Value {
 // type is unhashable.
 func (c *Constructor) setPossiblyUnhashableKey(m map[any]bool, key any, value bool, n *Node) {
 	defer func() {
-		if err := recover(); err != nil {
-			Fail(formatConstructorError(
-				fmt.Errorf("%v", err),
-				Mark{Line: n.Line, Column: n.Column},
-			))
+		if recover() != nil {
+			failUnhashableKey(key, n)
 		}
 	}()
 	m[key] = value
@@ -1122,11 +1125,8 @@ func (c *Constructor) setPossiblyUnhashableKey(m map[any]bool, key any, value bo
 // the key type is unhashable.
 func (c *Constructor) getPossiblyUnhashableKey(m map[any]bool, key any, n *Node) bool {
 	defer func() {
-		if err := recover(); err != nil {
-			Fail(formatConstructorError(
-				fmt.Errorf("%v", err),
-				Mark{Line: n.Line, Column: n.Column},
-			))
+		if recover() != nil {
+			failUnhashableKey(key, n)
 		}
 	}()
 	return m[key]
--- a/testdata/unmarshal_errors.yaml
+++ b/testdata/unmarshal_errors.yaml
@@ -82,20 +82,18 @@
 - unmarshal-error:
     name: Slice as map key beside a merge key (Issue 378)
     yaml: |
-      <<:
-      ? -
+      <<: {a: 1}
+      [b]: 2
     type: map[string]any
-    want: "go-yaml load error in constructor at L2.C3: cannot use '[]interface {}{interface {}(nil)}' as a map key; try decoding into yaml.Node"
+    want: "go-yaml load error in constructor at L2.C1: cannot use '[]interface {}{\"b\"}' as a map key; try decoding into yaml.Node"
 
 - unmarshal-error:
     name: Slice as map key in a merged mapping
     yaml: |
-      a: 1
-      <<:
-        ? -
-        : 2
+      <<: {[a]: 2}
+      b: 2
     type: map[any]any
-    want: "go-yaml load error in constructor at L3.C5: cannot use '[]interface {}{interface {}(nil)}' as a map key; try decoding into yaml.Node"
+    want: "go-yaml load error in constructor at L1.C6: cannot use '[]interface {}{\"a\"}' as a map key; try decoding into yaml.Node"
 
 - unmarshal-error:
     name: Forward reference to anchor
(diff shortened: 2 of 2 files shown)


ccoVeille (approved):

ccoVeille:
Thank you. It sounds the right fix indeed

hdimer:
> Thank you. It sounds the right fix indeed


Merci Christophe. Ravi de pouvoir apporter ma pierre à l'édifice.

internal/libyaml/constructor.go
```diff
@@ -1108,11 +1114,8 @@ func settableValueOf(i any) reflect.Value {
 // type is unhashable.
 func (c *Constructor) setPossiblyUnhashableKey(m map[any]bool, key any, value bool, n *Node) {
 	defer func() {
-		if err := recover(); err != nil {
-			Fail(formatConstructorError(
-				fmt.Errorf("%v", err),
```
ccoVeille, line 1113: There is one thing that annoys me.

The recovered error can longer be caught. I'm unsure what could have been the error. And if was possible to catch it out of the Unmarshal.

I don't have a computer with me, I cannot check right now.

Maybe @colinjlacy could check

hdimer: Checked it. Nothing was catchable before this PR either.

The old code used `fmt.Errorf("%v", err)`, not `%w`, so the recovered value was already flattened to a plain string before it reached the `LoadError`. On `main`, decoding `"<<:\n? -\n"`:

```
err.Error():                 go-yaml load error in constructor at L2.C3: runtime error: hash of unhashable type []interface {}
errors.As(err, &runtimeErr): false
unwrap chain:                *libyaml.LoadError -> *errors.errorString
```

Same chain on the branch, only `Message` differs. So `errors.Is` / `errors.As` against a `runtime.Error` returned false before too. The text was the only thing a caller ever had, and that text is what #378 is about.

On what the error could have been: each deferred block guards exactly one map operation on a `map[any]bool`, and `Construct` runs outside it. A map read can only panic on an unhashable key. A write can also panic on a nil map, but both callers pass a non-nil map (`mapping` guards on `mergedFields != nil`, `merge` allocates it with `make`). So it is always `hash of unhashable type <T>`. A user `UnmarshalYAML` that panics is unaffected, it never reaches these recovers.

Side note: neither recover block had test coverage before. The two new cases hit one each (set via `merge`, get via `mapping`), and both go red if I revert only the recover blocks.

If you do want it catchable, I would rather add a library sentinel (`errors.Is(err, ErrUnhashableKey)`) than leak the runtime type into the public error chain. Happy to do that as a follow-up, but it is new API, not something this PR took away.

ccoVeille: Thanks for checking.

I assumed that `fmt.Errorf("%#v", err)` that breaking the flow (maybe on purpose 🤷), so it's OK.

hdimer: Good. Leaving the error text as the contract then, and not adding the sentinel here. If anyone ever needs to branch on it in code, that is a small separate PR.

ingydotnet:
You can test this with the go-yaml CLI.

```
$ <<<'{<<: {a: 1}, b: 2}' ./go-yaml -y --unmarshal
a: 1
b: 2
$ <<<'{<<: {a: 1}, []: 2}' ./go-yaml -y --unmarshal
2026/09/01 10:27:41 Failed to process YAML:failed to load YAML: go-yaml load error in constructor at L1.C14: cannot use '[]interface {}{}' as a map key; try decoding into yaml.Node
$ 
```

testdata/unmarshal_errors.yaml
```diff
@@ -79,6 +79,24 @@
...
+      <<:
+      ? -
+    type: map[string]any
+    want: "go-yaml load error in constructor at L2.C3: cannot use '[]interface {}{interface {}(nil)}' as a map key; try decoding into yaml.Node"
+
+- unmarshal-error:
+    name: Slice as map key in a merged mapping
+    yaml: |
+      a: 1
+      <<:
```
ingydotnet: Can we change this to `<<: {[]: 2}`. I feel like it's clearer than using explicit block key syntax.

testdata/unmarshal_errors.yaml
```diff
@@ -79,6 +79,24 @@
     yaml: '{{.}}'
     want: "go-yaml load error in constructor at L1.C2: cannot use 'map[string]interface {}{\".\":interface {}(nil)}' as a map key; try decoding into yaml.Node"
 
+- unmarshal-error:
+    name: Slice as map key beside a merge key (Issue 378)
+    yaml: |
+      <<:
```
ingydotnet: I'd change this to:
```
<<: {a: 1}
[b]: 2
```

ingydotnet (changes requested):
I feel like using explicit key syntax in test examples is not necessary and harder to grok easily.

ingydotnet:
I can make the test changes if you want.

ingydotnet (approved):

hdimer:
Thanks, your version reads better than mine.

I checked the two cases still land on the two different recover paths after the rewrite: `[b]: 2` goes through `setPossiblyUnhashableKey`, `{[a]: 2}` through `getPossiblyUnhashableKey`. And reverting `constructor.go` to pre-fix still turns both red with the raw `runtime error: hash of unhashable type []interface {}` text, so the regression coverage is intact.

Full suite green here. Nothing further from my side.

testdata/unmarshal_errors.yaml
ingydotnet: [78073e9](https://github.com/yaml/go-yaml/pull/403/commits/78073e95eab5fe5a651c92e0d43e63ada51d4fac)

testdata/unmarshal_errors.yaml
ingydotnet: [78073e9](https://github.com/yaml/go-yaml/pull/403/commits/78073e95eab5fe5a651c92e0d43e63ada51d4fac)

aaronbronow (approved):
LGTM

ccoVeille (approved):

ingydotnet:
Merged. Thanks, @hdimer!

Issue #235, opened 2026-01-04
Support omitzero as std lib supports for JSON


ccoVeille:
The idea is to look for a method `IsZero() bool` on a field when it uses `omitzero` in the `yaml` Go tags option.

This could help to avoid rendering some nodes when needed.

I don't think there is an need for an option for this. But if we disagree on this, it should be enabled by default and there could be an option to disable it.

ccoVeille:
I would like to amend my idea with what I wrote here


> This would require further checks.
> encoding/json omitzero is not omitempty
> yaml/go-yaml omitempty might behave the same way for struct (I have to check) that encoding/json handles omitzero
> But then I'm unsure yaml/go-yaml should handle omitzero the same way for non-struct
> So it might not be about aliasing. Maybe it's more about handing omitzero the same way that omitempty for struct only.

_Originally posted by @ccoVeille in https://github.com/yaml/go-yaml/pull/404#issuecomment-5415325690_

PR #404, opened 2026-08-25
Accept omitzero as an alias of omitempty

AshSgDe29071999:
Go 1.24 added `omitzero` for `encoding/json`. `getStructInfo` only knows `omitempty`, `flow`, and `inline`, so a struct that shares tags with json panics in `mappingStruct`:

```
panic: unsupported flag "omitzero" in tag "...,omitzero"
```

This library's `omitempty` already drops zero values (including structs), so treat `omitzero` as the same flag.

This is the same failure as the archived v3 tree ([go-yaml/yaml#1068](https://github.com/go-yaml/yaml/issues/1068)).

---

Maintainers' edits to the PR description 👇

Somehow related to #235

### DIFF
--- a/internal/libyaml/structmeta.go
+++ b/internal/libyaml/structmeta.go
@@ -157,7 +157,11 @@ func getStructInfo(st reflect.Type) (*structInfo, error) {
 		if len(fields) > 1 {
 			for _, flag := range fields[1:] {
 				switch flag {
-				case "omitempty":
+				case "omitempty", "omitzero":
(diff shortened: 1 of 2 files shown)


ccoVeille:
This would require further checks.

encoding/json omitzero is not omitempty

yaml/go-yaml omitempty might behave the same way for struct (I have to check) that encoding/json handles omitzero

But then I'm unsure yaml/go-yaml should handle omitzero the same way for non-struct

So it might not be about aliasing. Maybe it's more about handing omitzero the same way that omitempty for struct only.

Issue #360, opened 2026-05-29
Enhance our CLI to use alternate inputs.


ingydotnet:
Currently, the go-yaml CLI takes YAML input and displays four main different outputs which you select:

* Tokens (load stack)
* Events (load stack)
* Nodes (load stack)
* YAML (dump stack)

This is wonderful and really useful for debugging situations to see where in the stack things go wrong. 
However, it would be even better if we could use some of these outputs as inputs to another CLI process. 
Something like this:

```
./go-yaml --token <<<'a: b   # c' |
./go-yaml --from=tokens --YAML
```

The first process outputs tokens as YAML and the second process reads that YAML into the part of the loader that needs tokens, and turns the YAML into token structs and continues. 

Of course, we don't have to use pipes. We could write stuff to files and mutate it and then feed it as input to different processes:

```
./go-yaml --token <<<'a: b   # c' > tokens.yaml
vim tokens.yaml   # make changes
./go-yaml --from=tokens --YAML < tokens.yaml
```

Or even pipe it into a program that makes changes automatically (`yq` comes to mind). 

```
./go-yaml --token <<<'a: b   # c' |
yq '...' |
./go-yaml --from=tokens --YAML
```

We are really going to want to have this going forward. It lets people test almost every situation for go-yaml situationally and without needing to write go code to do it. It allows people to show reproducible behavior and possible solutions with a few lines in a issue or pull request.

ccoVeille:
I understand the need

Maybe we will have to think about the CLI args

The --from=tokens is interesting, but the having --to=events could be interesting.

Right now the -t -y -j -e is not obvious.

having a --from and --to that share same arguments logic would be an improvement, I think.

Not that we need to remove existing arguments, but I feel like it could be more explicit.

Also the --help could display the tokens, events, yaml flow.

PR #409, opened 2026-09-01
Enable composable CLI pipelines

ingydotnet:
The CLI now treats token, event, node, and YAML representations as interchangeable pipeline stages.
Each stage can be serialized, transformed by ordinary text tools, and read back by a later go-yaml invocation.
Input stages are detected from their serialized contracts, while the -f option resolves ambiguous input explicitly.

For example, token output can be edited before conversion back to YAML:

    $ <<<'X: Y' ./go-yaml -t | perl -pe 's/X/XXX/' | ./go-yaml -Y
    XXX: Y

This makes short shell pipelines useful as executable bug reports and allows yq or other filters to operate between go-yaml stages.

The change also preserves alias node values and restores the public Node field name from Text to Value so serialized node contracts round trip correctly.

Refs #360.


cmd/go-yaml/main.go
```diff
@@ -481,6 +483,39 @@ func main() {
...
+			target, profuse = stageNode, *nodeProfuseMode
+		case *yamlMode, *yamlPreserveMode:
+			target = stageYAML
+		case *jsonMode, *jsonPrettyMode:
+			log.Fatal("JSON output is only supported for YAML text input")
+		default:
+			log.Fatal("No output stage specified")
+		}
+		if err := processStructuredInput(structured, target, profuse,
+			!*longMode, *yamlPreserveMode, opts); err != nil {
```
ccoVeille: Having a `!*longMode` looks very strange to me, as a Gopher.

You then define

                 compact := !*longMode // compact is default, long mode negates it

I would expect compact to be defined at the top and being used

cmd/go-yaml/stage.go
```diff
@@ -0,0 +1,907 @@
...
+		return nil, errors.New("invalid event input: stream boundaries are required")
+	}
+	return events, nil
+}
+
+func validEventType(value string) bool {
+	switch EventType(value) {
+	case EventStreamStart, EventStreamEnd, EventDocumentStart, EventDocumentEnd,
+		EventScalar, EventSequenceStart, EventSequenceEnd, EventMappingStart,
+		EventMappingEnd, "ALIAS", EventTailComment:
```
ccoVeille: Why no constant for ALIAS ?

cmd/go-yaml/main.go
```diff
@@ -481,6 +483,39 @@ func main() {
 		os.Exit(1)
 	}
 
+	inputData, err := io.ReadAll(input)
+	if err != nil {
+		log.Fatal("Failed to read input:", err)
+	}
```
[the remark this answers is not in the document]
ingydotnet: fine for CLI

cmd/go-yaml/main.go
```diff
@@ -481,6 +483,39 @@ func main() {
...
+	if structured.stage != stageYAML {
+		var target stageKind
+		var profuse bool
+		switch {
+		case *tokenMode, *tokenProfuseMode:
+			target, profuse = stageToken, *tokenProfuseMode
+		case *eventMode, *eventProfuseMode:
+			target, profuse = stageEvent, *eventProfuseMode
+		case *nodeMode, *nodeProfuseMode, *longMode:
+			target, profuse = stageNode, *nodeProfuseMode
```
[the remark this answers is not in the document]
ingydotnet: 56b2207

cmd/go-yaml/main.go
```diff
@@ -481,6 +483,41 @@ func main() {
...
+			target = stageYAML
+		case *jsonMode, *jsonPrettyMode:
+			log.Fatal("JSON output is only supported for YAML text input")
+		case *longMode:
+			target = stageNode
+		default:
+			log.Fatal("No output stage specified")
+		}
+		if err := processStructuredInput(structured, target, profuse,
+			!*longMode, *yamlPreserveMode, opts); err != nil {
```
ccoVeille: Having a `!*longMode` looks very strange to me, as a Gopher.

You then define

                 compact := !*longMode // compact is default, long mode negates it

I would expect compact to be defined at the top and being used

cmd/go-yaml/stage.go
```diff
@@ -0,0 +1,913 @@
...
```
ccoVeille: Why no constant for ALIAS ?

ccoVeille (approved):
The is way too much to review.

So I will simply approve based on the fact it's for the CLI, even if they are changes in the YAML lib to make work.

ingydotnet:
> The is way too much to review.
> 
> So I will simply approve based on the fact it's for the CLI, even if they are changes in the YAML lib to make work.

I agree. I really think it would be good if we could move this CLI to its own repo. I just want things to stay in sync, and start reinforcing the habits that people use the CLI to report and discuss go-yaml issues.

cmd/go-yaml/main.go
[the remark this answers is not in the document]
ingydotnet: 1599fd1

cmd/go-yaml/main.go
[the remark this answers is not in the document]
ingydotnet: 1599fd1

cmd/go-yaml/main.go
ingydotnet: 1599fd1

cmd/go-yaml/stage.go
ingydotnet: 1599fd1

PR #411, opened 2026-09-01
Validate implicit document endings

### LINKED ISSUE #410: Leading whitespaces cause go-yaml to incorrectly ignore everything afterwards
Hi all! I found the following bug/edge case, where the contents of `data` is not valid yaml syntax:

```go
package main

import (
	"fmt"
	"log"
	"go.yaml.in/yaml/v4"
)

// notice the extra whitespace before the `name` key
var data = `
 name: test
value: 2
`

type Example struct {
	Name  string
	Value int
}

func main() {
	item := Example{}

	err := yaml.Unmarshal([]byte(data), &item) // parses as {test 0} instead of {test 2} or returning an error
	if err != nil { log.Fatalf("error: %v", err) }

	fmt.Printf("%v\n", item) // prints {test 0}
}
```

This should result in an error, but it parses in an unexpected way where only the initial indented lines are parsed and everything after the indented lines is ignored

### Reproducibility

This behaviour occurs both on `go1.27.0 darwin/arm64` as well as `go1.27.0 windows/amd64` with `go.yaml.in/yaml/v4 v4.0.0-rc.6`

ingydotnet:
Reject an implicit document end when the following token cannot start a properly separated document or end the stream.

Add data-driven coverage for issue 410 and update YAML Test Suite expectations for newly passing load cases.

Fixes #410

### DIFF
--- a/internal/libyaml/parser.go
+++ b/internal/libyaml/parser.go
@@ -595,10 +595,18 @@ func (parser *Parser) parseDocumentEnd(event *Event) error {
 	end_mark := token.StartMark
 
 	implicit := true
-	if token.Type == DOCUMENT_END_TOKEN {
+	switch token.Type {
+	case DOCUMENT_END_TOKEN:
 		end_mark = token.EndMark
 		parser.skipToken()
 		implicit = false
+	case VERSION_DIRECTIVE_TOKEN, TAG_DIRECTIVE_TOKEN,
+		DOCUMENT_START_TOKEN, STREAM_END_TOKEN:
+		// Document ended properly.
+		// https://github.com/yaml/go-yaml/pull/410
+	default:
+		return formatParserError(
+			"did not find expected <document start>", token.StartMark)
 	}
 
 	parser.tag_directives = parser.tag_directives[:0]
--- a/testdata/unmarshal_errors.yaml
+++ b/testdata/unmarshal_errors.yaml
@@ -129,6 +129,13 @@
     want: "go-yaml load error in scanner (while scanning a simple key) at L3.C1-L4.C1: could not find expected ':'"
 
 - unmarshal-error:
+    name: Unseparated content after implicit document (Issue 410)
+    yaml: |2
+       name: test
+      value: 2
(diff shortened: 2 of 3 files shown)


ingydotnet:
I love that this fixed 3 failing suite tests!

internal/libyaml/parser.go
```diff
@@ -599,6 +599,12 @@ func (parser *Parser) parseDocumentEnd(event *Event) error {
 		end_mark = token.EndMark
 		parser.skipToken()
 		implicit = false
+	} else if token.Type != VERSION_DIRECTIVE_TOKEN &&
+		token.Type != TAG_DIRECTIVE_TOKEN &&
+		token.Type != DOCUMENT_START_TOKEN &&
+		token.Type != STREAM_END_TOKEN {
+		return formatParserError(
+			"did not find expected <document start>", token.StartMark)
 	}
```
ccoVeille: Can we use a switch ?


```suggestion
	switch token.Type {
	case DOCUMENT_END_TOKEN:
 		end_mark = token.EndMark
 		parser.skipToken()
 		implicit = false
	case VERSION_DIRECTIVE_TOKEN, TAG_DIRECTIVE_TOKEN, DOCUMENT_START_TOKEN, STREAM_END_TOKEN:
		// the expected one
	default:
		return formatParserError(
			"did not find expected <document start>", token.StartMark)
 	}
```

ingydotnet: d24c76e

ccoVeille (approved):

ccoVeille:
I approved it earlier @ingydotnet

I don't see changes since I approved it

ingydotnet:
> I approved it earlier @ingydotnet
> 
> I don't see changes since I approved it

@ccoVeille Sorry. Can you approve it again?

ccoVeille (approved):
I still see no changes, but I can approve it again, sure 🤷‍♂️

aaronbronow (approved):
all tests pass on my machine
Linux aarch64 (arm64) with Go 1.27.0

PR #412, opened 2026-09-02
Keep interior comments in empty flow collections

### LINKED ISSUE #406: comment in empty flow mapping are leading to invalid YAML
By taking a lot at #405 

- #405

I have identified some bugs in the code that can be tested with this in emitter.yaml

```yaml
# Known bug: a comment inside an otherwise-empty flow mapping/sequence is
# attached as a FootComment on the *START* event (there is no key/value event
# to carry it), but the caller that dispatches that event flushes the foot
# comment right away, before the paired END event ever writes the opening
# bracket. The comment ends up hoisted out before the collection entirely,
# instead of staying inside the brackets.
- roundtrip:
    name: Comment in empty flow mapping as sequence item is misplaced
    yaml: |
      ---
      - {
      # comment
      }
    want: |
      ---
      - {
        # comment
        }

- roundtrip:
    name: Comment in empty flow sequence as sequence item is misplaced
    yaml: |
      ---
      - [
      # comment
      ]
    want: |
      ---
      - [
        # comment
        ]

- roundtrip:
    name: Comment in empty top-level flow mapping is misplaced
    yaml: |
      ---
      {
      # comment
      }
    want: |
      ---
      {
      # comment
      }

# This case is the most severe: the comment is hoisted above the block
# mapping value entirely, producing output that fails to re-parse.
- roundtrip:
    name: Comment in empty flow mapping as block value is misplaced
    yaml: |
      ---
      key:
        {
      # comment
        }
    want: |
      ---
      key:
        {
        # comment
        }

```

BetterAndBetterII:
Empty flow collections were dropping interior comments (`FootComment` with empty content). Hold that comment as `HeadComment` so it survives emit.

Fixes #406

### DIFF
--- a/internal/libyaml/emitter.go
+++ b/internal/libyaml/emitter.go
@@ -602,6 +602,11 @@ func (emitter *Emitter) emitFlowSequenceItem(event *Event, first, trail bool) er
 				return err
 			}
 		}
+		// Interior comments of an empty flow sequence are held as HeadComment
+		// on START; write them after '[' and before ']'.
+		if err := emitter.processHeadComment(); err != nil {
+			return err
+		}
 		emitter.flow_level--
 		emitter.indent = emitter.indents[len(emitter.indents)-1]
 		emitter.indents = emitter.indents[:len(emitter.indents)-1]
@@ -1005,6 +1010,7 @@ func (emitter *Emitter) emitSequenceStart(event *Event) error {
 	} else {
 		emitter.state = EMIT_BLOCK_SEQUENCE_FIRST_ITEM_STATE
 	}
+	emitter.holdEmptyFlowCollectionComment()
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


internal/libyaml/testdata/emitter.yaml
```diff
@@ -493,6 +493,47 @@
...
+# on the START event. The emitter must keep it inside the brackets instead of
+# flushing it before the opening indicator is written.
+- roundtrip:
+    name: Comment in empty flow mapping as sequence item stays inside
+    yaml: |
+      ---
+      - {
+      # comment
+      }
+    want: "{\n  # comment\n}"
```
ingydotnet: I would prefer the `want` field be a literal scalar here:
```
want: |-
  {
    # comment
  }
```

internal/libyaml/testdata/emitter.yaml
```diff
@@ -493,6 +493,47 @@
...
+
+- roundtrip:
+    name: Comment in empty flow sequence as sequence item stays inside
+    yaml: |
+      ---
+      - [
+      # comment
+      ]
+    want: "[\n  # comment\n]"
```
ingydotnet: same (literal) for all of these

ingydotnet (changes requested):

internal/libyaml/emitter.go
```diff
@@ -1023,9 +1029,36 @@ func (emitter *Emitter) emitMappingStart(event *Event) error {
...
+//
+// The parser attaches that comment as FootComment on the START event because
+// there is no key/value event to carry it. emitNode only switches state, so
+// the caller would write the foot comment before the matching END event
+// writes '{' or '['. Reclassify it as HeadComment so it is emitted inside
+// the collection after the opening indicator.
+func (emitter *Emitter) holdEmptyFlowCollectionComment() {
+	if len(emitter.FootComment) == 0 || len(emitter.HeadComment) != 0 {
+		return
+	}
```
ccoVeille, line 1047: This makes me think we could still have a bug with a code that would have an head and foot comment in an empty map or array

I feel like we should things like this to the emitter.yaml file with roundtrip to check the behavior

```yaml
# head comment
[]
# foot comment
```

```
# head comment
[
  # inner comment
]
# foot comment
```

I'm unsure about the way to handle such yaml with our lib.

Note: Comment made on a phone, without checking further. I can be simply wrong

internal/libyaml/emitter_test.go
```diff
@@ -75,6 +75,88 @@ func TestEmitFoldedScalarNoExtraNewline(t *testing.T) {
...
+				var event Event
+				err := parser.Parse(&event)
+				assert.NoErrorf(t, err, "re-parse of emitted YAML %q: %v", got, err)
+				if event.Type == STREAM_END_EVENT {
+					break
+				}
+			}
+		})
+	}
+}
```
ccoVeille: I don't think we need this

The YAML test suite (the things added to emitter.yaml) is enough, I think

internal/libyaml/emitter.go
ccoVeille: Thanks for adding these to tests and confirming I was wrong

ccoVeille (approved):
Sounds good. Thanks for working on this

PR #414, opened 2026-09-07
Fix multiline scalar serialization round-trips

### LINKED ISSUE #76: Marshal-ing a multi-line scalar starting with an empty line produces invalid YAML
Unmarshal() on the following (note the empty line following the literal scalar):

```yaml
jobs:
      - run: |

             uname -a
```

Will be Marshal() as:

```yaml
jobs:
    - run: |4

        uname -a
```

Which fails Unmarshal() because there is a two space indentation although the [block indentation indicator](https://yaml.org/spec/1.2.2/#8111-block-indentation-indicator) is set to four

`decode_test.go:...: Unmarshal() returned error: yaml: line 1: did not find expected key`

A reproducer is provided at https://github.com/yaml/go-yaml/pull/75/files

### LINKED ISSUE #383: Emitted yaml not accepted back when block scalar starts with a tab character

### LINKED ISSUE #399: Marshal produces spec-invalid YAML for multi-line strings inside block sequences

### LINKED ISSUE #1071: leading spaces in multiline string: marshal/unmarshal fails
In specific combination of list item and leading spaces in string value, this results in an unparseable result:

```go
func TestLoadYaml(t *testing.T) {
	r := require.New(t)

	type Params struct {
		Description string `yaml:"description"`
	}
	type Spec struct {
		Parameters []Params `yaml:"parameters"`
	}
	spec := Spec{
		Parameters: []Params{
			{
				Description: "  a\nb",
			},
		},
	}

	marshalledContents, err := yaml.Marshal(&spec)
	r.NoError(err)
	r.NoError(yaml.Unmarshal(marshalledContents, &spec))
}
```

If Marshalling with encoder indent forced to 1, then test passes:

```go
func TestLoadYaml(t *testing.T) {
	r := require.New(t)

	type Params struct {
		Description string `yaml:"description"`
	}
	type Spec struct {
		Parameters []Params `yaml:"parameters"`
	}
	spec := Spec{
		Parameters: []Params{
			{
				Description: "     a\nb",
			},
			{
				Description: " a\n      b",
			},
		},
	}

	buf := new(bytes.Buffer)
	enc := yaml.NewEncoder(buf)
	enc.SetIndent(1)
	r.NoError(enc.Encode(&spec))

	var unmarshalledSpec Spec
	r.NoError(yaml.Unmarshal(buf.Bytes(), &unmarshalledSpec))
	r.Equal(spec, unmarshalledSpec)
}
```

injeniero:
## Problem

`Marshal` can emit multiline scalars that `Unmarshal` then rejects with `did not find expected key`. Three defects in the same class are addressed here; all reproduce on a plain `Marshal` → `Unmarshal` round-trip, with no explicit style set.

```go
// 1) leading empty line + indented first content line  (issue #76)
yaml.Marshal(map[string]string{"k": "\n  a\n b\n"})

// 2) multiline string inside a block sequence  (issue #399)
yaml.Marshal([]any{"  a\nb\n"})

// 3) multiline scalar whose first content line starts with a tab  (issue #383)
yaml.Marshal(map[string]string{"k": "\tthis\nis\nmultiline"})
```

## Cause and fix

All three live in the block-scalar path:

1. **Missing indentation indicator.** `writeBlockScalarHints` emitted the indicator only when the *first byte* was a space, so a scalar starting with empty lines then an indented line got none, and the parser inferred the wrong indentation. Fix: skip leading line breaks, then test for a space.

2. **Wrong indicator value.** The indicator was always `BestIndent`, but per the spec it is relative to the parent node; a block sequence item only adds two columns, so `BestIndent` overstated it. Fix: derive the indicator from the content's indentation relative to the parent, falling back to `BestIndent` when outside the 1..9 range.

3. **Leading-tab scalars.** A multiline value whose first content line begins with a tab was emitted as a literal block, but block indentation is spaces only, so it never parsed back. Fix: such values are emitted double-quoted (`shouldUseLiteralStyle` returns false; the serializer falls back to double-quoted for multiline values that are not literal-eligible).

The #65 behavior is preserved: a leading blank line followed by *unindented* content still gets no indicator.

## Tests

- `TestBlockScalarIndentIndicatorRoundTrip` and `TestTabLeadingScalarRoundTrip` cover the round-trips across `SetIndent` values and nesting.
- Added `roundtrip`/`encode`/`should-literal` fixtures in the testdata files.
- Existing `TestScalarStyleWithTabs` expectations updated to the corrected double-quoted output.
- Full suite green including the yaml-test-suite (e.g. `F6MC`).

Closes #76
Closes #399
Closes #383
Also fixes go-yaml/yaml#1071. Supersedes #393, #396, #388.

### DIFF
--- a/internal/libyaml/emitter.go
+++ b/internal/libyaml/emitter.go
@@ -1985,11 +1985,39 @@ func (emitter *Emitter) writeDoubleQuotedScalar(value []byte, allow_breaks bool)
 // writeBlockScalarHints writes the indentation and chomping indicators for
 // block scalars.
 func (emitter *Emitter) writeBlockScalarHints(value []byte) error {
-	if isSpace(value, 0) {
-		// https://github.com/yaml/go-yaml/issues/65
-		// isLineBreak(value, 0) removed as the linebreak will only
-		// write the indentation value.
-		indent_hint := []byte{'0' + byte(emitter.BestIndent)}
+	// A parser infers a block scalar's indentation from its first non-empty
+	// line, so an explicit indentation indicator is needed whenever that line
+	// begins with a space. Leading line breaks only produce empty lines, which
+	// carry no indentation of their own, so skip past them first: this keeps
+	// the #65 behavior (no indicator when the first content line is not
+	// indented) while covering leading-blank-then-indented scalars.
+	// https://github.com/yaml/go-yaml/issues/65
+	// https://github.com/yaml/go-yaml/issues/76
+	i := 0
+	for i < len(value) && isLineBreak(value, i) {
+		i += width(value[i])
+	}
+	if i < len(value) && isSpace(value, i) {
+		// The indicator is the content's indentation relative to the parent
+		// node, not BestIndent: a block sequence item only adds 2 columns, so
+		// emitting BestIndent produces an unparseable document. It must also be
+		// a single digit 1..9, so clamp it to that range and align the actual
+		// content indentation to the clamped value; that keeps the document
+		// valid and round-trippable even when a caller requests a wider indent
+		// (e.g. SetIndent(10)).
+		// https://github.com/go-yaml/yaml/issues/1071
+		parent_indent := 0
+		if len(emitter.indents) > 0 && emitter.indents[len(emitter.indents)-1] > 0 {
+			parent_indent = emitter.indents[len(emitter.indents)-1]
+		}
+		indent := emitter.indent - parent_indent
+		if indent < 1 {
+			indent = 1
+		} else if indent > 9 {
+			indent = 9
+		}
+		emitter.indent = parent_indent + indent
+		indent_hint := []byte{'0' + byte(indent)}
 		if err := emitter.writeIndicator(indent_hint, false, false, false); err != nil {
 			return err
 		}
--- a/internal/libyaml/serializer.go
+++ b/internal/libyaml/serializer.go
@@ -215,8 +215,13 @@ func (s *Serializer) node(node *Node, tail string) {
 			style = LITERAL_SCALAR_STYLE
 		case node.Style&FoldedStyle != 0:
 			style = FOLDED_SCALAR_STYLE
-		case strings.Contains(value, "\n"):
+		case shouldUseLiteralStyle(value):
 			style = LITERAL_SCALAR_STYLE
+		case strings.Contains(value, "\n"):
+			// Multiline but not literal-eligible (e.g. first content line
+			// starts with a tab): a block scalar would not round-trip, so
+			// emit a double-quoted scalar instead.
+			style = DOUBLE_QUOTED_SCALAR_STYLE
 		case forceQuoting:
 			style = s.quotePreference.ScalarStyle()
 		}
--- a/yaml_test.go
+++ b/yaml_test.go
@@ -2739,6 +2739,78 @@ func TestSetIndent(t *testing.T) {
 	assert.Equal(t, "a:\n        b:\n                c: d\n", buf.String())
 }
 
+// A block scalar whose first non-empty line is indented needs an explicit
+// indentation indicator, and that indicator is relative to the parent node's
+// indentation, not BestIndent.
+// This covers both the missing-indicator case (a leading empty line followed
+// by indented content) and the wrong-value case (a scalar nested in a block
+// sequence item, which only adds two columns of indentation).
+// See https://github.com/yaml/go-yaml/issues/76
+// and https://github.com/go-yaml/yaml/issues/1071
+func TestBlockScalarIndentIndicatorRoundTrip(t *testing.T) {
+	type params struct {
+		Description string `yaml:"description"`
+	}
+	type spec struct {
+		Parameters []params `yaml:"parameters"`
+	}
+
+	// Includes indents at and beyond the 1..9 indicator range (9, 10) to
+	// cover clamping of the indentation indicator.
+	for _, indent := range []int{0, 1, 2, 3, 4, 8, 9, 10} {
+		for _, description := range []string{
+			"  a\nb",
+			"     a\nb",
+			" a\n      b",
+			"\n  indented\nregular",
+			"\n\n  more indented\nregular",
+		} {
+			in := spec{Parameters: []params{{Description: description}}}
+
+			var buf bytes.Buffer
+			enc := yaml.NewEncoder(&buf)
+			if indent > 0 {
+				enc.SetIndent(indent)
+			}
+			assert.NoError(t, enc.Encode(&in))
+			assert.NoError(t, enc.Close())
+
+			var out spec
+			err := yaml.Unmarshal(buf.Bytes(), &out)
+			assert.NoErrorf(
+				t, err, "indent %d, encoded as:\n%s", indent, buf.String())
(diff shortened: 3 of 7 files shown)


injeniero:
@ingydotnet Pinging you as the last committer on the repo. This PR fixes several critical bugs with the lib we found in real use, like not being able to serialize properly text with new lines.

I tried to address all present bugs and PRs that partially fix related issues in a single one. Please let me know how I can help to get this merged.

Thanks!!

yaml_test.go
```diff
@@ -2739,6 +2739,78 @@ func TestSetIndent(t *testing.T) {
...
+		in := doc{Text: text}
+		out, err := yaml.Marshal(&in)
+		assert.NoError(t, err)
+
+		var back doc
+		assert.NoErrorf(t, yaml.Unmarshal(out, &back),
+			"encoded as:\n%s", out)
+		assert.DeepEqualf(t, in, back, "encoded as:\n%s", out)
+	}
+}
```
ccoVeille, line 2812: Can't we express this in one of the .yaml test file ?

yaml_test.go
```diff
@@ -2739,6 +2739,78 @@ func TestSetIndent(t *testing.T) {
...
+
+			var out spec
+			err := yaml.Unmarshal(buf.Bytes(), &out)
+			assert.NoErrorf(
+				t, err, "indent %d, encoded as:\n%s", indent, buf.String())
+			assert.DeepEqualf(
+				t, in, out, "indent %d, encoded as:\n%s", indent, buf.String())
+		}
+	}
+}
```
ccoVeille, line 2786: I can see the loop on indentation, but could you try using the .yanl test file to validatw them.

The yaml test suite is intended to be reused on other projects using YAML.

We would appreciate

internal/libyaml/serializer.go
```diff
@@ -215,8 +215,13 @@ func (s *Serializer) node(node *Node, tail string) {
...
+		case shouldUseLiteralStyle(value):
 			style = LITERAL_SCALAR_STYLE
+		case strings.Contains(value, "\n"):
+			// Multiline but not literal-eligible (e.g. first content line
+			// starts with a tab): a block scalar would not round-trip, so
+			// emit a double-quoted scalar instead.
+			style = DOUBLE_QUOTED_SCALAR_STYLE
 		case forceQuoting:
 			style = s.quotePreference.ScalarStyle()
 		}
```
ccoVeille, line 227: I'm unsure I saw test for the priority of shoulduseLiteral or new line against forceQuoting ?

Do you think you could add tests for this?

yaml_test.go
```diff
@@ -2984,27 +3056,27 @@ func TestScalarStyleWithTabs(t *testing.T) {
 		},
 		{
 			"\tThis starts with tab\nand is long enough\nfor literal style",
-			"|-\n    \tThis starts with tab\n    and is long enough\n    for literal style\n",
+			"\"\\tThis starts with tab\\nand is long enough\\nfor literal style\"\n",
```
ccoVeille, line 3059: Here and everwhere. Please use raw string literal for readability


```suggestion
			`"\tThis starts with tab\nand is long enough\nfor literal style"` + "\n",
```

ccoVeille:
@silverwind

You had review another PR

https://github.com/yaml/go-yaml/pull/396#issuecomment-5392842527

Could you review this one, and let us know what you think about this one compared to the others?

Thanks

internal/libyaml/emitter.go
```diff
@@ -1987,11 +1987,39 @@ func (emitter *Emitter) writeDoubleQuotedScalar(value []byte, allow_breaks bool)
...
+		// emitting BestIndent produces an unparseable document. It must also be
+		// a single digit 1..9, so clamp it to that range and align the actual
+		// content indentation to the clamped value; that keeps the document
+		// valid and round-trippable even when a caller requests a wider indent
+		// (e.g. SetIndent(10)).
+		// https://github.com/go-yaml/yaml/issues/1071
+		parent_indent := 0
+		if len(emitter.indents) > 0 && emitter.indents[len(emitter.indents)-1] > 0 {
+			parent_indent = emitter.indents[len(emitter.indents)-1]
+		}
```
ccoVeille, line 2014: Please use what I had suggested in another PR

https://github.com/yaml/go-yaml/pull/393#discussion_r3768416901

```suggestion
		parent_indent := 0
		if n := len(emitter.indents); n > 0 && emitter.indents[n-1] > 0 {
			parent_indent = emitter.indents[n-1]
		}
```

ccoVeille:
Also, may I know your motivation behind opening a new PR for things that were already in review? I would have appreciated if you dropped your comments on the existing PRs instead.

I'm concerned about the "supersedes #X, #Y, and #Z" framing, particularly since I can see code borrowed from those PRs. If you're building on existing work, could you please use `Co-Authored-By` in your commit messages to give proper credit?

Additionally, while your code may indeed supersede the others, it would help to have an explicit explanation of *why* and *how* your approach is superior. Right now, the reasoning behind the supersession isn't clear from the PR description.

PR #417, opened 2026-09-17
scanner: keep comments after document separators

### LINKED ISSUE #416: Comment above a document separator is silently dropped when the next document has two comment blocks
### What happens

A comment block written immediately above a document separator is silently
dropped, if the document *after* that separator is introduced by two comment
blocks separated by a blank line. A second block also migrates across the
separator, so the surviving comments come out attached to different documents
than they were written on.

Nothing is reported — no error, and the comment is not reachable on any node
after `Decode`, so a caller cannot detect the loss and work around it.

### Reproducer

```go
package main

import (
	"bytes"
	"fmt"
	"strings"

	yaml "go.yaml.in/yaml/v3"
)

const src = `a: 1

# before-separator
---
# block-one

# block-two
b: 2
`

func main() {
	var docs []*yaml.Node
	dec := yaml.NewDecoder(strings.NewReader(src))
	for {
		var n yaml.Node
		if err := dec.Decode(&n); err != nil {
			break
		}
		docs = append(docs, &n)
	}

	var buf bytes.Buffer
	enc := yaml.NewEncoder(&buf)
	enc.SetIndent(2)
	for _, d := range docs {
		if err := enc.Encode(d); err != nil {
			panic(err)
		}
	}
	enc.Close()

	fmt.Print(buf.String())
	fmt.Printf("\n%q survived: %v\n", "# before-separator",
		strings.Contains(buf.String(), "before-separator"))
}
```

### Expected

All three comments preserved, each on the document it was written on.

### Actual

```yaml
a: 1

# block-one
---
# block-two
b: 2
```

```
"# before-separator" survived: false
```

`# before-separator` is gone. `# block-one`, written *below* the separator,
comes back *above* it — so on a reparse it belongs to the first document
rather than the second.

Walking the whole node tree after `Decode` confirms the comment is not
attached anywhere: the input has three comment lines, and the sum of all
`HeadComment`, `LineComment` and `FootComment` fields across both documents
accounts for only two.

### Versions

Reproduces on:

- `go.yaml.in/yaml/v3` v3.0.5
- `go.yaml.in/yaml/v4` v4.0.0-rc.6
- `gopkg.in/yaml.v3` v3.0.1

Go 1.27.1, darwin/arm64.

### Narrowing it down

The loss needs both halves of the shape. With only one comment block after the
separator, or with no comment before it, the same input round-trips cleanly:

```yaml
# round-trips fine — one block after the separator
a: 1

# before-separator
---
# block-one
b: 2
```

```yaml
# round-trips fine — nothing before the separator
a: 1
---
# block-one

# block-two
b: 2
```

So it is specifically the combination: a block above the separator, and two
blank-line-separated blocks below it. The parser appears to file the first
post-separator block as the preceding document's `FootComment`, overwriting
the block that was actually there and discarding it.

### Why it matters

This makes decode→encode lossy in a way that is invisible to the caller, and
it breaks the fixed-point property that matters for a formatter: the emitter
*produces* this exact shape when it writes a document `FootComment` ahead of a
document whose head comment contains a blank line. So a formatter's own output
does not reparse to itself, and running it twice deletes a comment that
survived the first run.

I hit this in a linter/formatter for Kubernetes manifests, where the house
style is a `# label` line above each `---`. Formatting such a file dropped
comments and needed three passes to reach a stable result. The workaround is
the one @mikefarah describes in #34 — take the document-boundary comments off
the tree and emit them as text yourself — which works, but means callers who
care about comment fidelity cannot use the node comment fields at document
boundaries at all.

Related to the general comment-handling discussion in #34; filing separately
since this is a specific, minimal case with a concrete round-trip consequence.

team-humaki:
Comments after `---` were treated as a foot of the document-start token when a blank line followed. That attached them to the previous document, and a later comment block in the same scan could overwrite a real pre-separator foot comment.

`---` / `...` are delimiters, not content nodes, so comments after them stay with the following document.

Fixes #416


ccoVeille:
I feel like this lead to have an empty YAML document that contains only a comment, right?

I'm a bit afraid about the breaking change it could lead on some project.

Maybe it's a matter of option that should be be enabled with the yaml.V3 setting used by Unmarshal, but be fixed with Load

What do you think @ingydotnet @colinjlacy ?

Maybe I'm over thinking and we should simply fix it like it's done in this PR.

@team-umaki please do not rush in doing changes while we discuss about it. I'm curious about youe feedback also

saitanallensantiago26-beep:
> Comments after `---` were treated as a foot of the document-start token when a blank line followed. That attached them to the previous document, and a later comment block in the same scan could overwrite a real pre-separator foot comment.
> 
> `---` / `...` are delimiters, not content nodes, so comments after them stay with the following document.
> 
> Fixes #416

PR #418, opened 2026-09-17
Encoder: return error from SetIndent

### LINKED ISSUE #323: Encoder.SetIndent panics on negative input — undocumented, opportunity to return error before v4 stable
## Summary

`Encoder.SetIndent` panics when called with a negative number of spaces. The doc comment does not mention this panic, so callers have no way to discover this behavior without reading the source. Since v4 is still a release candidate, this is a good time to consider returning an error instead.

## Reproduction

Confirmed on `go.yaml.in/yaml/v4 v4.0.0-rc.4`, Go 1.25.1, darwin/arm64.

```go
package main

import (
    "bytes"
    "go.yaml.in/yaml/v4"
)

func main() {
    var buf bytes.Buffer
    enc := yaml.NewEncoder(&buf)
    enc.SetIndent(-1)
    // panic: yaml: cannot indent to a negative number of spaces
}
```

## Current Source (yaml.go ~line 547)

```go
// SetIndent changes the used indentation used when encoding.
func (e *Encoder) SetIndent(spaces int) {
    if spaces < 0 {
        panic("yaml: cannot indent to a negative number of spaces")
    }
    e.encoder.Indent = spaces
}
```

The doc comment says only *"SetIndent changes the used indentation used when encoding."* — no mention of panic behavior.

## Why This Matters

- **Downstream impact already observed.** The `yq` tool has [mikefarah/yq#2329](https://github.com/mikefarah/yq/issues/2329) where `yq -I=-1` produces a raw panic stacktrace because the value is passed through to `SetIndent`. End users see a crash instead of a meaningful error.
- If `spaces` comes from user input, configuration, or computation, a negative value crashes the entire program with no opportunity for the caller to handle it gracefully.
- The Go standard library is inconsistent here — `text/tabwriter.Init` panics on negative values (also undocumented), but `encoding/json.Encoder.SetIndent` does no validation at all. Neither is ideal; returning an error is the most composable option for a library.

## Suggested Fix

Since v4 has not yet reached stable, the signature can still change without breaking semver:

**Option A — Return an error (preferred):**
```go
func (e *Encoder) SetIndent(spaces int) error {
    if spaces < 0 {
        return fmt.Errorf("yaml: cannot indent to a negative number of spaces")
    }
    e.encoder.Indent = spaces
    return nil
}
```

**Option B — Clamp to zero:**
```go
func (e *Encoder) SetIndent(spaces int) {
    if spaces < 0 {
        spaces = 0
    }
    e.encoder.Indent = spaces
}
```

**Option C — At minimum, document the panic:**
```go
// SetIndent changes the used indentation used when encoding.
// It panics if spaces is negative.
func (e *Encoder) SetIndent(spaces int) {
```

## Related

- #167 (umbrella: get rid of panic and return errors where possible)

## Environment

- yaml version: go.yaml.in/yaml/v4 v4.0.0-rc.4
- Go version: go1.25.1 darwin/arm64

team-humaki:
`Encoder.SetIndent` (and `Dumper.SetIndent`) paniced on a negative indent because `failDumpf` is not recovered on this path. Callers that pass user or computed values (for example `yq -I=-1`) crash instead of getting an error.

Return a `DumpError` instead, matching the existing dump-error types and the `WithIndent` option which already rejects invalid values with an error.

Fixes #323


ccoVeille:
The change of the signature is a no-go for me

Also, sounds like AI contribution noise to me

ccoVeille (changes requested):
No breaking changes should be introduced.

team-humaki:
Restored the original SetIndent signature. Negative indent is now 0 instead of a panic, so yq -I=-1 no longer crashes the process.

`go test . ./internal/libyaml` green.

saitanallensantiago26-beep:
> `Encoder.SetIndent` (and `Dumper.SetIndent`) paniced on a negative indent because `failDumpf` is not recovered on this path. Callers that pass user or computed values (for example `yq -I=-1`) crash instead of getting an error.
> 
> Return a `DumpError` instead, matching the existing dump-error types and the `WithIndent` option which already rejects invalid values with an error.
> 
> Fixes #323

PR #419, opened 2026-09-18
yaml: accept omitzero as an omitempty alias

### LINKED ISSUE #235: Support omitzero as std lib supports for JSON

team-humaki:
encoding/json added \`omitzero\` in Go 1.24. go-yaml treated it as an unknown flag and paniced from \`getStructInfo\` during both Marshal and Unmarshal.

Treat \`omitzero\` like \`omitempty\`. Existing \`isZero\` already omits zero structs and honors \`IsZeroer\`, matching omitzero's intent.

Fixes #235


ccoVeille:
Duplicate of

- https://github.com/yaml/go-yaml/pull/404

Both PRs need to be reviewed

Also, please check my comment here

https://github.com/yaml/go-yaml/pull/404#issuecomment-5415325690

PR #421, opened 2026-09-20
Indent closing quote after a trailing break in single-quoted scalars

ritk7:
## Problem

A single-quoted scalar whose content ends right after a line break gets its
closing quote emitted with no indentation, landing at column 0:

```yaml
# actual
description: ' Get quacked.

  - Duck

'
```

```yaml
# expected
description: ' Get quacked.

  - Duck

  '
```

Reported downstream at [mikefarah/yq#2819](https://github.com/mikefarah/yq/issues/2819)
(yq depends on this library for its YAML encoding). I reproduced it directly
against this library with no yq code involved - a plain `Marshal` of a
`yaml.Node` with `Style: SingleQuotedStyle` and a value ending in `\n`
reproduces it standalone.

I checked the decoded value itself against PyYAML first, since the input
YAML here is a somewhat unusual case (a quoted scalar with embedded blank
lines): PyYAML decodes it to the exact same string this library does, so
the parse side is correct. Re-dumping that same string with PyYAML as a
single-quoted scalar produces indentation matching what this fix produces,
confirming the encoder is what's wrong, and what the correct output should
look like.

## Cause

`writeSingleQuotedScalar` (`internal/libyaml/emitter.go`) re-indents before
writing the next character whenever the previous one was a line break:

```go
} else {
    if breaks {
        if err := emitter.writeIndent(); err != nil {
            return err
        }
    }
    ...
}
```

but that only runs from inside the loop, triggered by a following
character. When the value's last byte is itself a line break, the loop
exits with `breaks` still `true` and the closing quote indicator gets
written immediately, with nothing having called `writeIndent()` since the
last break.

## Fix

Do the same `breaks` check once more after the loop, before writing the
closing indicator.

## Testing

- Added a case to `marshalTests` in `yaml_test.go` covering this exact
  shape (a single-quoted scalar ending in a line break).
- Confirmed the new test fails without the fix (`- Duck` line unindented at
  column 0 for the closing quote) and passes with it.
- `go test ./...` passes, including the YAML test suite (`go test
  ./yts/...` after `make get-test-data`).


ccoVeille:
Duplicate of

- https://github.com/yaml/go-yaml/pull/398

Discussion is in progress there

